PR #7, opened 2016-08-17
misc fixes for issues encountered when building CUDA apps

jackmitch:


madisongh:
These all look good.  Thanks!

PR #11, opened 2016-09-21
linux: enable SWAP in defconfig

jackmitch:


madisongh:
Any particular reason for having this turned on by default?

jackmitch:
I would say the ability to have a swap file/partition is a pretty standard assumption on Linux. I also imagine that the tegra is quite often used for processing neural nets which can very easily exceed the 4G on the board. I'm not too fussed either way and can carry it as a override in my own layer.

madisongh:
Thanks, I'll think about it.  Swap isn't typically used on embedded systems, but the TX1 isn't typical.

madisongh:
I've gone ahead and merged this.  Thanks!

PR #17, opened 2016-10-21
RFC: add support for tegra tk1 (tegra124) machine

jackmitch:
Just an RFC for now to see if this is something you would be willing take in the main layer. There is still room for improvement but this is the majority of the work done. Let me know your thoughts on it.

Boot and CUDA application tested on a TK1 Development board.

Notable missing support:
- Flashtool type image
- cudnn/visonworks support
- X11 integration (binaries installed, but our image doesn't use X11 so untested)

A major quirk is that its built using the Linaro GCC 4.9 compiler support form meta-linaro-toolchain, this is due to CUDA 6.5 only supporting the GCC 4.x series compilers. It's very hard to work around as the C++ ABI changed in GCC 5.x which means the pre-compiled binary CUDA libs won't link if a newer compiler with additional workarounds is used (days were wasted on fighting this!).


madisongh:
Overall, this looks great.

I'm not sure what to do about the gcc 4.9 dependency - we have a similar problem with CUDA 8.0, which isn't compatible with gcc 6, but at least gcc 5.x is still available in oe-core morty/master.

I'm traveling at the moment, but I'll give these changes a try when I return. Many thanks!

jackmitch:
Yeah, the GCC requirements are a pain. In our layer I sed out the unsupported error and with a few choice GCC flags you can get CUDA apps to compile with GCC 6. However with the ABI breakage between 4 and 5 you can't do that, so I can't even hack it to get it to build with a currently supported oe-core GCC. I'll keep ticking it over and ask around to see if anybody has come across something similar.

madisongh:
I've been through some test builds and, modulo a few tweaks, things look good.

For kernel builds, I've added a patches-l4t-r21.5 branch to my linux-tegra repo and added a few patches so the the kernel builds with gcc 5 or 6 as the host compiler and gcc 5 as the cross compiler.  With that branch in place, I think we can collapse the separate kernel recipes back into one.

I tweaked the mesa-related bbappends to avoid collisions for the tk1 builds, mirroring what I did for tx1.

For CUDA builds, I had success using gcc 5 with two approaches:
* adding `-D_GLIBCXX_USE_CXX11_ABI=0` to CXXFLAGS for any recipes compiling C++ code that might be used by a CUDA recipe
* configuring the compiler and runtime with `--with-default-libstdcxx-abi=gcc4-compatible`
I had to delete the compiler version check in CUDA's `host_config.h` header file, too, of course.

How does that sound?  I can probably get to merging this in with these modifications over the weekend.

jackmitch:
Sounds great to me, good spot on the gcc4-compatible, that's what I was missing before and hence having trouble linking with nvlibs old ABI, everything else new ABI. If you want to push a testing branch I can also give it a whirl here.

madisongh:
I pushed a tegra-k1-wip branch.  It's missing the patch of host_config.h to remove the compiler version check, and still needs some additional cleanup.

I'm not sure whether the compiler config modification or the CXXFLAGS setting is the better approach, both are in there at the moment.

jackmitch:
I may be wrong here but I expect you would only need the compiler config modification, using the CXXFLAGS is redundant I think. However, I don't think you could use _only_ the CXXFLAGS as then all the other binaries in the system would have the new ABI and then the CUDA apps wouldn't be able to link with system libs.

I'll give the branch a go and report back if it builds before I get off today ;)

madisongh:
Yes, you're right, and I mentioned it because of the redundancy.  The extra CXXFLAGS setting shouldn't break anything, though, and I'll get it cleaned up before merging.

jackmitch:
OK, I can get CUDA apps to compile with GCC5 but GCC6 explodes, so we need a versioning in there to default to 5 as current master is default 6. What happens when 5 drops out of master I'm not sure... cross that bridge when we come to it.

madisongh:
Yeah, when I get some time I'll see if I can make gcc6 work as well.  Thanks!

madisongh:
Merged manually.  Many thanks!

PR #34, opened 2017-06-21
Kernel boot param tweaks for tegra186

sashko:


madisongh:
I'm not sure about these changes.   In my testing on my Jetson-TX2, U-Boot sets the `${devnum}` variable appropriately when it scans for the `extlinux.conf` file on the SDcard, and the kernel is able to mount the SDcard just fine.

Can you explain more about the problem you're trying to solve here?

sashko:
I use U-Boot 2016.07-g5971907 and in my testing on my Jetson-TX2, it doesn't, so the system doesn't boot from sd card.

JackKiefer:
I had to implement this fix as well.

madisongh:
Have you modified your kernel configuration to disable `CONFIG_DEVTMPFS_MOUNT`?   Because that is set in the `defconfig` provided in the layer, so the extra parameter should be redundant.

And if you're booting via the `extlinux.conf` file, U-Boot should be setting `${devnum}`.  My U-Boot is also 2016.07-g5971907, the stock NVIDIA L4T R27.1 version.

A log of the serial console showing the problem might help.

Issue #42, opened 2017-08-16
u-boot: eMMC transfer speed significantly slower than stock u-boot on Tegra186


derekstraka:
When loading a kernel from the eMMC device, the stock u-boot is able to copy data at a rate of ~36.2 MiB/s while the yocto u-boot only transfers at ~314.5 KiB/s.

Stock u-boot:
```
Tegra186 (P2771-0000-500) # version
U-Boot 2016.07-g0ce7ca2 (Jul 20 2017 - 00:45:25 -0700)
aarch64-unknown-linux-gnu-gcc (GCC) 4.8.5
GNU ld (GNU Binutils) 2.24

Tegra186 (P2771-0000-500) # mmc info
Device: Tegra SD/MMC
Manufacturer ID: 11
OEM: 100
Name: 032G3 
Tran Speed: 52000000
Rd Block Len: 512
MMC version 5.1
High Capacity: Yes
Capacity: 29.1 GiB
Bus Width: 8-bit
Erase Group Size: 512 KiB
HC WP Group Size: 4 MiB
User Capacity: 29.1 GiB WRREL
Boot Capacity: 4 MiB ENH
RPMB Capacity: 4 MiB ENH

Retrieving file: /boot/Image
20280368 bytes read in 534 ms (36.2 MiB/s)
```

Built u-boot:
```
Tegra186 (P2771-0000-500) # version
U-Boot 2016.07 (Aug 16 2017 - 13:55:00 +0000)
aarch64-poky-linux-gcc (GCC) 6.3.0
GNU ld (GNU Binutils) 2.28.0.20170307

Retrieving file: /boot/Image-4.4.38-l4t-r28.1+gebda89f
10563776 bytes read in 32753 ms (314.5 KiB/s)

Tegra186 (P2771-0000-500) # mmc info
Device: Tegra SD/MMC
Manufacturer ID: 11
OEM: 100
Name: 032G3 
Tran Speed: 52000000
Rd Block Len: 512
MMC version 5.1
High Capacity: Yes
Capacity: 29.1 GiB
Bus Width: 8-bit
Erase Group Size: 512 KiB
HC WP Group Size: 4 MiB
User Capacity: 29.1 GiB WRREL
Boot Capacity: 4 MiB ENH
RPMB Capacity: 4 MiB ENH
```

derekstraka:
I'm guessing the clock rate isn't being set correctly on the yocto version of the bootloader.

derekstraka:
[...]

PR #44, opened 2017-08-18
image_types_tegra: allow the underlying filesystem type to be overridden

derekstraka:
The usage of ext4 with a yocto image can cause slow loading of the kernel image from
u-boot (See #42).  As a workaround, an ext3 filesystem can be used, so allow
integrators to override the default ext4 filesystem.


madisongh:
Thanks

PR #48, opened 2017-09-22
Pyro l4t r28.1

iyadq:
The proposed patches make use of machine definitions consistent and distinct from SOC.  Patches were build validated with core-image-minimal and tegra{186|210}-minimal-initramfs.


madisongh:
The `tegra<nnn>`s are the SOC family designators (from `soc-family.inc` in OE-Core), rather than MACHINEs.  This makes it easy to create other MACHINEs that use the same SOC/SOM without having to update the compatibility settings for all the recipes.

The `jetson-<xxx>`  MACHINEs are the development boards, and while that's not an ideal naming, it's consistent with common usage.  There's no issue with having hyphens in MACHINE names.

I originally created a separate `-minimal-init` and `-minimal-initramfs` recipes for the TX2 (tegra186) because I wasn't sure I could share the same recipes between TX1 and TX2.  Now that support for the TX2 is in place, it looks like we can, so if you'd like to rework the patches to include just those changes, that would be great, thanks!

PR #50, opened 2017-09-28
linux-tegra_4.4.bb: Change AUTOREV to specific commit hash

BassemMohsen:
Having AUTOREV prevents yocto reproducible offline builds
based on previous downloaded sources.

This is important when releasing the sources to customer and guarantee
that
it's the same source code used in their released binaries.

Having a floating AUTOREV means less control over the source code used.


madisongh:
Thanks, I meant to go back to a fixed srcrev once I had got a base set of patches in place.

PR #52, opened 2017-10-04
Use KB for  IMAGE_ROOTFS_ALIGNMENT value.

stammw:
As stated in meta/classes/image_types.bbclass:7, IMAGE_ROOTFS_ALIGNMENT is in KB.


stammw:
Sure, all three files are now corrected.

madisongh:
Thanks!

PR #55, opened 2017-11-10
image_types_tegra: include flashtools to tegraflash package

sapiippo:
Add all the necessary tools from ${SOC_FAMILY}-flashtools-native
to the tegraflash zip, so it can be flashed without needing the
tools available separately.


madisongh:
Thanks for this, it's something I've been meaning to add but hadn't got round to.

PR #66, opened 2018-04-01
Various small fixes

shr-project:
I don't use tegra for anything, these are just fixes for build issues detected by my world builds when includes meta-tegra layer while building for other MACHINEs.


madisongh:
Thanks, Martin, much appreciated.  I've manually merged most of these, with a couple of modifications.  The kernel gcc-7 build issue should already be fixed in master due to some patches in the kernel repository.

Thanks again!

shr-project:
Hi Matt, thanks for quick merge.

The first commit "tegra-libraries: add runtime dependency on pango and cairo" also doesn't seem to be merged, was it intentionally skipped?

I'll send similar pull requests for morty, pyro and rocko now.

shr-project:
Ah I see that it's not needed in master anymore after:

commit 4eec72b41f20a69ff7e9a3486e6b5df1c497aed6
Author: Matt Madison <matt@madison.systems>
Date:   Tue Apr 3 15:46:37 2018 -0700

    tegra-libraries: split libnvosd.so out into separate package
    
    and add pango and cairo to its runtime dependencies.

will try to rebuild world with my change dropped.

PR #85, opened 2018-04-26
alsa shipp into /etc/

BarbareBzhdev:
Need to set ship package to fix error `alsa-state: 1 installed and not shipped files. [installed-vs-shipped]`


recipes-bsp/alsa-state/alsa-state.bbappend
```diff
@@ -9,3 +9,5 @@ do_install_append_tegra() {
 RCONFDEPS = ""
 RCONFDEPS_tegra = "tegra-configs-alsa"
 RDEPENDS_${PN} += "${RCONFDEPS}"
+
+FILES_${PN} += "/etc"
```
madisongh: Could you change the `/etc` there to `"${sysconfdir}"`?

madisongh:
Thanks!

PR #93, opened 2018-07-24
linux-tegra: add GCC 8 support

martingkelly:
There are a number of GCC 8-related compile issues. With these patches,
linux-tegra compiles properly with GCC 8. These patches are mostly
workarounds, as the final fixes are either not yet merged upstream or
many patches, which will likely cause merge conflicts. When we move to
another Linux kernel version, we should drop these patches.

Signed-off-by: Martin Kelly <mkelly@xevo.com>


madisongh:
Thanks, @martingkelly, for the patches.  I have pulled them directly into the `patches-l4t-r28.2` branch in [linux-tegra](https://github.com/madisongh/linux-tegra) and will just update the SRCREV here.

madisongh:
I have pushed the SRCREV update that includes these patches.  Thanks again!

martingkelly:
Thanks!

PR #97, opened 2018-08-13
alsa-state: add the missing tegra-configs-alsa-21.6 package

AlbanBedel:
Since commit f3c95e88 (alsa-state: update bbappend to use common tegra
override) building alsa-state on TK1 is failling because there is no
tegra-configs-alsa-21.6 package.

Update the TK1 alsa config to also use a tegra-configs-alsa package
like the other tegra platforms.


madisongh:
Thanks, this is something I'd been meaning to do but never got around to it.

I'm curious, though, about the build errors you were seeing, since I haven't been seeing any errors in my test builds.

AlbanBedel:
I don't have any log left, but the gist of it is that alsa-state has `RCONFDEPS_tegra = "tegra-configs-alsa"` but there was no tegra-configs-alsa_*.bb that supported tegra124. So building alsa-state on tegra124 was not possible any more.

PR #99, opened 2018-08-20
Add gl header and pkgconfig files to target SDK

BassemMohsen:
This will install gl header files and pkgconfig files to target sdk. This is required during cross compilation of gl rendering applications.

Signed-off-by: Bassem Mohsen <bmohsen@luxoft.com>

### DIFF
--- a/recipes-graphics/mesa/mesa_%.bbappend
+++ b/recipes-graphics/mesa/mesa_%.bbappend
@@ -48,3 +48,11 @@ PRIVATE_LIBS_${PN}-stubs-dev = "\
     libGLESv2.so.2 \
     libGL.so.1 \
 "
+
+TOOLCHAIN_TARGET_TASK += " \
+    libgles1-mesa-dev \
+    libgles2-mesa-dev \
+    libgles3-mesa-dev \
+    libegl-mesa-dev \
+    libgl-mesa-dev \
+"


madisongh:
Hmm, I'm not sure this will do what you want it to do, since the TOOLCHAIN_TARGET_TASK variable is normally used in an image recipe to specify the packages to include, rather than within a package recipe.

The right way to do this, I think, is to refine the tegra-libraries recipe a bit... split the GL libraries out into separate packages, including -dev packages, and have their -dev packages RDEPENDS on the the mesa -dev packages you mention here.  Could be a bit tricky to do without introducing circular dependencies somewhere, though.

BassemMohsen:
You are right, It works only when I add it to my image. just tested it, and files aren't populated to the SDK.
another way is to include those files in the meta-tegra/recipes-bsp/tegra-binaries/files.
and deploy them as part of the dev package of the tegra-libraries.

Specially that the gl package config files .pc description says `Description: Mesa OpenGL ES 2.0 library`
What do you think?

madisongh:
NVIDIA doesn't appear to ship any of the pkgconfig or header files of their own, at least with the L4T kit, so getting the mesa ones pulled in is probably the right approach.  Could you try [this change](https://github.com/madisongh/meta-tegra/commit/80777b036592f40d276a0aec66de7c625f9593a8) and see if works for you?

BassemMohsen:
I think the headers exist in Tegra_Multimedia_API_R28.2.1_aarch64. but there's not pkgconfig though. :/
I think we would have to write them ourselves.
I will try the change and get back to you.

BassemMohsen:
I tried the change, It does populate the sdk with headers and pkgconfig files. I wonder why didn't you add libgles3-mesa-dev as well to the RDEPENDS_${PN}-dev?

BassemMohsen:
I see the answer to this is 
`poky/meta/recipes-graphics/mesa/mesa.inc:RDEPENDS_libgles2-mesa-dev += "libgles3-mesa-dev"`

BassemMohsen:
I think you can merge your commit to  rocko-l4t-r28.2 and just close this PR.

BassemMohsen:
Thanks :)

PR #101, opened 2018-08-29
Kernel config fragments

mardy:
Submitting this for rocko: hopefully it's the right branch :-)

Anyway, I'm not a Yocto expert, so please review this carefully. The idea is taken from meta-raspberrypi, where a similar solution was adopted before.

### DIFF
--- a/recipes-kernel/linux/linux-tegra_4.4.bb
+++ b/recipes-kernel/linux/linux-tegra_4.4.bb
@@ -4,7 +4,10 @@ DESCRIPTION = "Linux kernel from sources provided by Nvidia for Tegra processors
 LICENSE = "GPLv2"
 LIC_FILES_CHKSUM = "file://COPYING;md5=d7810fab7487fb0aad327b76f1be7cd7"
 
+KERNEL_VERSION_SANITY_SKIP="1"
+
 inherit kernel
+require recipes-kernel/linux/linux-yocto.inc
 
 PV .= "+git${SRCPV}"
 FILESEXTRAPATHS_prepend := "${THISDIR}/${BPN}-${@bb.parse.BBHandler.vars_from_file(d.getVar('FILE', False),d)[1]}:"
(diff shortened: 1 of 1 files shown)


madisongh:
Pulling in the entire `linux-yocto.inc` file is too much - that file is really intended for kernels managed by the Yocto kernel team, or at least following the same basic pattern, which the NVIDIA kernel is most certainly not.

I'm OK with providing a way to make it easier to make kernel configuration changes in a `bbappend`, if that's the main goal here.  Perhaps something a bit simpler, though? I've personally always found the linux-yocto setup difficult to use.

mardy:
Hi! So, what I need is a way to enable AppArmor in the kernel, from our `bbappend` file; do you know how other people have solved this issue before?

philipc:
I'm not familiar with meta-raspberrypi, but it seems to be using [`kernel_configure_variable`](http://git.yoctoproject.org/cgit.cgi/meta-raspberrypi/tree/recipes-kernel/linux/linux-raspberrypi.inc#n46) to add the config changes to `CONF_SED_SCRIPT`, instead of using config fragments, whereas your PR is trying to use `CONF_SED_SCRIPT` without actually defining it?

The meta-raspberrypi approach can probably be made to work, but if you want to use config fragments, then I think all you need is the `merge_config.sh` command from [here](https://git.yoctoproject.org/cgit/cgit.cgi/poky/tree/meta/classes/kernel-yocto.bbclass#n312), but use a list of config fragments defined in the `bbappend` instead of what `kernel-yocto.bbclass` does. Either way, `linux-yocto.inc` shouldn't be needed.

mardy:
I found a way to do it without touching the meta-tegra layer and contain all the changes in my bbappend:

```# Find config path
FILESEXTRAPATHS_append := ":${THISDIR}/files"

# Kernel configs
SRC_URI += " file://apparmor.cfg"

kernel_do_configure_prepend() {
    config_flags="-m"
    configs="${B}/.config ${WORKDIR}/apparmor.cfg"
    PATH=${S}/scripts/kconfig:${PATH}
    CFLAGS="${CFLAGS} ${TOOLCHAIN_OPTIONS}" HOSTCC="${BUILD_CC} ${BUILD_CFLAGS} ${BUILD_LDFLAGS}" HOSTCPP="${BUILD_CPP}" CC="${KERNEL_CC}" ARCH=${ARCH} merge_config.sh -O ${B} ${config_flags} ${configs} > ${B}/merge_config_build.log 2>&1
    if [ $? -ne 0 ]; then
        bbfatal_log "Could not configure ${KMACHINE}-${LINUX_KERNEL_TYPE}"
    fi
}
```
If you think that this is the right way to do it, then feel free to close this issue. :-)

PR #109, opened 2018-09-20
machine/include/tegra124.conf: Fix the DEFAULTTUNE assignment.

AlbanBedel:
The DEFAULTTUNE assignment is unconditional and misplaced so it always
overrides the settings from the distro or local.conf. Instead it
should use the default value assignment (?=) and be placed before the
include of tune-cortexa15.inc.

Furthermore the value set is suboptimal as the tegra124 SoC support
thumb mode and vfpv4, using "cortexa15thf-neon-vfpv4" enable these two
features in the default settings.

### DIFF
--- a/conf/machine/include/tegra124.inc
+++ b/conf/machine/include/tegra124.inc
@@ -2,10 +2,10 @@
 
 SOC_FAMILY = "tegra124"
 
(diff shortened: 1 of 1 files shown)


madisongh:
Thanks!

PR #112, opened 2018-09-24
xserver-xorg: bump bbappend to 1.20.1

martingkelly:
The bbappend references 1.19.6, but the version moved to 1.20.1, so
update it. The patch applied without issues, so keep it.

Signed-off-by: Martin Kelly <mkelly@xevo.com>


madisongh:
Thanks!

PR #114, opened 2018-11-03
Add Mender OTA Update Support

dwalkes:
Thanks for creating and sharing the meta-tegra project!  Amazing work here!

These commits add capabilities related to support for [mender](https://mender.io/), a project which adds OTA firmware update support.

Other required changes for mender support are found in the layer at [meta-mender-tegra](https://github.com/Trellis-Logic/meta-mender-tegra), however I can move any changes into this layer if requested.

I've only tested on the sumo branch thus far, and I only have a Jetson TX2 development board for testing.

I wanted to send the pull request to understand if/how you'd like these merged before I test on other branches and I could use volunteers who could test on other hardware.  I'm especially interested in feedback about:

* The way the data partition change support is implemented.  I'd like to make this less platform specific and not require a specific .xml configuration file as is currently done with the [flash_l4t_t186.custom.xml](https://github.com/Trellis-Logic/meta-mender-tegra/blob/master/recipes-bsp/tegra-binaries/files/flash_l4t_t186.custom.xml) file.  Perhaps a patch to the xml file which could apply across platform.  I'm curious how much of this customization you want to support in meta-tegra and how much I should put in a different layer like the one I've created.  I've put minimal support in meta-tegra as-is.
* Likely related to the questions above, any ideas to streamline the configuration detailed [here](https://github.com/Trellis-Logic/meta-mender-tegra/blob/master/README.md).  I'm wondering if the KERNEL_ROOTSPEC default could be changed to reference ${distro_bootpart} [here](https://github.com/madisongh/meta-tegra/blob/542e7ff5eb93720b72d4b2d4c3a755505a60fb18/recipes-kernel/linux/linux-tegra_3.10.bb#L32) which would remove this step.  The u-boot environment offset calculation is a bit of a hack as well.

Thanks for considering and thanks again for sharing this project!


dwalkes:
> Other required changes for mender support are found in the layer at meta-mender-tegra, however I can move any changes into this layer if requested.

Based on the suggestion from Mirza Krak on the Mender dev team in [mender hub](https://hub.mender.io/t/nvidia-tegra-jetson-tx2/123/4), I'll move any changes you would like from [meta-mender-tegra](https://github.com/Trellis-Logic/meta-mender-tegra) into meta-tegra on this PR and put the rest in [meta-mender-community](https://github.com/mendersoftware/meta-mender-community).

madisongh:
Thanks for this!

On the flash layout customization: I'd like to keep this layer as close to stock L4T as I can, so keeping a custom XML file in the other layer is probably best.  I don't think a patch would work across platforms, since the L4T XML files vary a bit from platform to platform.  NVIDIA also changes the layouts from version to version of L4T, which further complicates things.

The KERNEL_ROOTSPEC change you suggest sounds right - I'd be happy to take a PR for that, or I'll make the change myself when I get a chance.

I'm fine with taking the u-boot-fw-utils recipe addition.  I don't think you need to allocate a separate partition for the u-boot environment, though.  On both the TX1 and TX2, the NVIDIA u-boot places the environment at the end of the `/dev/mmcblk0boot1` partition, and I already have the kernel patches in place to make that partition visible to Linux.  All you need is an `/etc/fw_env.config` file for use by u-boot-fw-utils:
```
/dev/mmcblk0boot1    0x3fe000   0x2000
```
I've used this on my custom TX1 and TX2 devices.  The only thing to watch out for is that on the TX2 (but not the TX1), the NVIDIA bootloaders also put a pseudo-GPT into `boot1` that maps their layout of the boot partitions in `boot0` and `boot1`. I only ran into problems with this when I enabled environment redundancy for my customized u-boot - the second environment area started overwriting that pseudo-GPT, bricking my TX2.

As for the other configuration customization... some of the customizations are machine-specific, some are distro-specific.  For the machine-specific ones, you could add a `jetson-tx2-mender.conf` machine config file to your mender-tegra layer that does a `require conf/machine/jetson-tx2.conf` and then does the other additions/modifications.  For the distro-related config changes, just create a `.inc` file and have people include that in either their `local.conf` or distro config file to have it apply.

Finally, for handling the flashing of the data partition, I'd recommend creating an `image_types_mender-tegra.bbclass` (or whatever) in your mender-tegra layer that `inherit`s `image_types_tegra.bbclass` and then add a `tegraflash_create_flash_config_append_tegra186` that does a `sed -i` to make your modifications.  Then use the reference the new bbclass in your `IMAGE_CLASSES` setting.

Thanks again!

dwalkes:
Thank you very much for the comments and suggestions @madisongh, especially the notes about u-boot environment and mmcblk0boot1 partition which I didn't catch in my first implementation.  I'll work on updating the PR for u-boot-fw-utils and KERNEL_ROOTSPEC changes this week and move the rest into my layer and eventually meta-mender-community.

> On the flash layout customization: I'd like to keep this layer as close to stock L4T as I can, so keeping a custom XML file in the other layer is probably best. I don't think a patch would work across platforms, since the L4T XML files vary a bit from platform to platform. NVIDIA also changes the layouts from version to version of L4T, which further complicates things

What do you think about using xmlstarlet or a similar XML command line tool to add relevant partition sections?  I worry keeping standalone files working with new platforms and L4T versions will be difficult to maintain.

madisongh:
> What do you think about using xmlstarlet or a similar XML command line tool to add relevant partition sections? I worry keeping standalone files working with new platforms and L4T versions will be difficult to maintain.

Unfortunately, just about everything about L4T is difficult to maintain, since NVIDIA sometimes makes incompatible and seemingly gratuitous changes from release to release, making upgrades/downgrades a bit of a pain.

No matter which approach you take - modifying the XML on the fly, patching it, or just providing a customized copy - it will have to be reviewed with each L4T update to make sure it's still going to be compatible.  And NV's flashing tools are different between TX1 and TX2, and seem to interpret the specific XML sections differently, so I'm not sure you could do this generically across platforms. Looks like Xavier shares the same tools with the TX2, but the layouts and specifics of the XML used are different. (Also, the TK1 doesn't use XML at all.)

recipes-kernel/linux/linux-tegra_4.4.bb
```diff
@@ -22,7 +22,7 @@ SRC_URI = "git://${KERNEL_REPO};branch=${SRCBRANCH} \
 "
 S = "${WORKDIR}/git"
 
-KERNEL_ROOTSPEC ?= "root=/dev/mmcblk\${devnum}p1 rw rootwait"
+KERNEL_ROOTSPEC ?= "root=/dev/mmcblk\${devnum}p${distro_bootpart} rw rootwait"
```
madisongh: I think you need a backslash before `${distro_bootpart}` here.

dwalkes: Thanks, still working on testing these changes, will post a comment when done.

recipes-bsp/u-boot/u-boot-common_2016.07.inc
```diff
@@ -0,0 +1,12 @@
+LICENSE = "GPLv2+"
```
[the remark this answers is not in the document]
dwalkes: Done, renamed to u-boot-tegra-common-2016.07.inc, thanks for the suggestion.

recipes-bsp/u-boot/u-boot-fw-utils-tegra_2016.07.bb
```diff
@@ -0,0 +1,43 @@
+require u-boot-common_${PV}.inc
+
+SUMMARY = "U-Boot bootloader fw_printenv/setenv utilities"
+DEPENDS = "mtd-utils"
+
+INSANE_SKIP_${PN} = "already-stripped"
+
+FILESEXTRAPATHS_prepend := "${THISDIR}/patches:"
```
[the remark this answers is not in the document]
dwalkes: Done, thanks for the suggestion

recipes-kernel/linux/linux-tegra_4.4.bb
dwalkes: Thanks for catching this, fixed and tested on hardware.

dwalkes:
Thank you for the suggestions @madisongh.  I believe I've completed all requested changes and I've tested on the sumo branch with T2 dev board and my meta-mender-community changes, see [this pull request](https://github.com/mendersoftware/meta-mender-community/pull/15) for associated mender-community changes.

Let me know if you'd like me to do any other testing of this standalone without mender layers and with my TX2 development board, or if you want me to do a new PR for a different branch.

madisongh:
Turns out that the use of the `${distro_bootpart}` variable in an `extlinux.conf` file doesn't actually work for the default case, because that variable is not actually in the environment - it's a local variable in U-Boot's "hush" shell, and the function that parses the `APPEND` line is the "simple" parser that only understands environment variables.

I have added a patch to our U-Boot sources so we can use it - if `distro_bootpart` is not set in the environment, but is present in the hush local variables, it will set it in the environment before parsing the `APPEND` line, then clear it afterwards.

It looks like you are setting `distro_bootpart` in the environment as part of your Mender integration, so this change shouldn't have any effect on how that works.

I have also added a suitable `/etc/fw_env.config` file for the TX1 and TX2, so `fw_printenv` and `fw_setenv` locate the environment correctly, and I had to work in an addition to the kernel command line for the TX1 so that the boot partitions would be accessible.

With these tweaks, I think we're good to go.  I've tested the changes on both TX1 and TX2 dev kits, and they look like they're doing the right things.

madisongh:
Merged manually, with the above-mentioned tweaks.

dwalkes:
> The NVIDIA bootloaders also put a pseudo-GPT into boot1 that maps their layout of the boot partitions in boot0 and boot1. I only ran into problems with this when I enabled environment redundancy for my customized u-boot - the second environment area started overwriting that pseudo-GPT, bricking my TX2.

I ran into this same issue over the weekend, and created a post [https://devtalk.nvidia.com/default/topic/1063652/jetson-tx2/mmcblk0boot1-usage-at-address-4177408-and-u-boot-parameter-storage-space-availability/](on the forum) to ask about this.

PR #115, opened 2018-11-06
tegra-binaries: Make sure glibc is used with nvidia libraries

AlbanBedel:
The libraries provided by nvidia are linked against glibc, so they
can't be used in builds using another libc. To prevent the user from
finding this the hard way add a check for the libc to all the recipes
that install some .so file from nvidia.


madisongh:
Thanks!

PR #122, opened 2019-03-06
RFS: u-boot-tegra /boot/Image conflict fix

martingkelly:
RFS: I'm not sure if this is correct, as I don't fully understand why we are installing /boot/Image in the u-boot package. I know that this fixes the build issue I am seeing, but I don't know if it may introduce some other issue. @madisongh could you advise?

Background:
Currently, I'm seeing an RPM conflict between the kernel-image package and the u-boot-tegra package because they both install the (identical) file /boot/Image. Removing the /boot/Image installation in u-boot-tegra fixes the issue.


madisongh:
Looks right to me.  dpkg must be more forgiving, since it never complained about the duplicate install.

Thanks!

martingkelly:
Maybe dpkg saw that the hashes are the same and allowed it through. Thanks!

PR #135, opened 2019-06-10
Use wayland with no x11 in distro features

kekiefer:
There's not much that needed to change to enable this. There is some minor patching to libglvnd so that it doesn't try to build glx and it has a couple of typedefs that it needs. Then a minor fixup to the mesa bbappend.

I'm sending this PR to warrior since that's where I've tested it (and where I'll be using it), but I've rebased the commits to [master+no_x](/kekiefer/meta-tegra/tree/master+no_x) if you'd rather I submit that. I don't anticipate any problems.


madisongh:
Worked OK for me, and I've merged this into both warrior and master.  Thanks!

PR #142, opened 2019-07-05
Fix u-boot-fw-utils build for mender support

dwalkes:
Broken with changes in d377317bfc3bd3e9365602d91e3fdeb5960de07d where
u-boot specific install steps were placed in common file shared with
fw-utils-tegra.  Fix naming of u-boot-common file and add a note that
this is shared with fw-utils.


dwalkes:
@madisongh the mender build is working for me on the warrior branch with this commit see [this project](https://github.com/Trellis-Logic/mender-tegra-warrior).  Let me know if you'd like me to test any other branches or would like any changes to the PR.

madisongh:
Thanks.  I tweaked this a bit to remove the gratuitous filename change and merged manually.

And I'll add building u-boot-fw-utils to my tests, which would have caught this sooner.

PR #149, opened 2019-07-16
cuda.bbclass: Add kernel-modules to CUDA RDEPENDS

leon-anavi:
Add kernel-modules to CUDA run-time dependencies. Without it
CUDA 10 is not properly working and CUDA-capable devices are
NOT detected.

### DIFF
--- a/classes/cuda.bbclass
+++ b/classes/cuda.bbclass
@@ -36,7 +36,7 @@ OECMAKE_CUDA_LINK_FLAGS ?= "${OECMAKE_CXX_LINK_FLAGS}"
 OECMAKE_CUDA_LIBRARIES ?= "-lcudadevrt -lcudart_static -lrt -lpthread -ldl"
 
 PACKAGE_ARCH_cuda = "${SOC_FAMILY_PKGARCH}"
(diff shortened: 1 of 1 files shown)


madisongh:
The dependency is only on the GPU driver, so `kernel-modules` is a bit too general.  The GPU driver is already marked as needed via MACHINE_EXTRA_RDEPENDS in the machine configs, so it should already be there unless you're building an image that omits the `packagegroup-machine-base` package group (such as a core-image-minimal).

That said, it wouldn't hurt to have the dependency in here, too, but if you could make it more specific, that would be better.  And I'll have to double-check on the TK1 that the kernel module gets the same name as the other platforms.

leon-anavi:
@madisongh OK, thank you for the feedback. I will make a few tests and make it more specific. Pls, give me a while to update this pull request.

Thanks,
Leon

leon-anavi:
@madisongh, following your recommendations, I have updated both GitHub PR for master and warrior. Now only kernel-module-nvgpu is in the run-time dependencies set by cuda.bbclass. I've successfully test the CUDA examples on TX2.

Best regards,
Leon

PR #150, opened 2019-07-16
cuda.bbclass: Add kernel-modules to CUDA RDEPENDS

leon-anavi:
Add kernel-modules to CUDA run-time dependencies. Without it
CUDA 10 is not properly working and CUDA-capable devices are
NOT detected.


leon-anavi:
Hi,

This patch fixes issues like this:

```
/usr/bin/cuda-samples/deviceQuery Starting...

 CUDA Device Query (Runtime API) version (CUDART static linking)

cudaGetDeviceCount returned 38
-> no CUDA-capable device is detected
Result = FAIL
```

The output above is from deviceQuery provided by recipe cuda-samples and tested on NVIDIA Jetson TX2. In my opinion it doesn't make sense to have CUDA without the kernel modules therefore I have added them to the CUDA run-time dependencies. The same patch is required for both branches master and [warrior](https://github.com/madisongh/meta-tegra/pull/149). 

Best regards,
Leon

PR #156, opened 2019-07-24
Fix transaction check errors with libvisionworks

dwalkes:
Attempts to build an SDK with -c populate_sdk and with libvisionworks included along with other
packages which use pkgconfig fails with:

> Error: Transaction check error:
  file /usr/lib/pkgconfig from install of libvisionworks-dev-1.6.0.500n-r0.armv8a_tegra186 conflicts with file from package spdlog-dev-1.3.1-r0.aarch64
  file /usr/lib/pkgconfig conflicts between attempted installs of libvisionworks-dev-1.6.0.500n-r0.armv8a_tegra186 and librealsense2-dev-2.13.0-r0.aarch64
  file /usr/lib/pkgconfig conflicts between attempted installs of cuda-command-line-tools-dev-10.0.166+1-r0.armv8a_tegra186 and libvisionworks-dev-1.6.0.500n-r0.armv8a_tegra186

Solution is to use install -d for ${libdir}/pkgconfig instead of cp -R --preserve copy

### DIFF
--- a/recipes-devtools/visionworks/libvisionworks_1.6.0.500n.bb
+++ b/recipes-devtools/visionworks/libvisionworks_1.6.0.500n.bb
@@ -25,9 +25,10 @@ do_compile() {
 }
 
 do_install() {
-    install -d ${D}${prefix} ${D}${libdir} ${D}${datadir}
+    install -d ${D}${prefix} ${D}${libdir} ${D}${datadir} ${D}${libdir}/pkgconfig
     cp -R --preserve=mode,timestamps ${B}/usr/include ${D}${prefix}/
-    cp -R --preserve=mode,timestamps ${B}/usr/lib/* ${D}${libdir}/
+    cp -R --preserve=mode,timestamps ${B}/usr/lib/pkgconfig/* ${D}${libdir}/pkgconfig
+    cp --preserve=mode,timestamps ${B}/usr/lib/libvisionworks.so ${D}${libdir}/
     cp -R --preserve=mode,timestamps ${B}/usr/share/visionworks ${D}${datadir}/
 }
 


dwalkes:
@madisongh it looks like libvisionworks-tracking, sfm and 1.4.3 version will have the same problem.  If this is the fix you want to make for these I can update these as well.

madisongh:
What does `rpmquery --dump` show for the owner and mode bits for `/usr/lib/pkgconfig` in these RPM packages?  Looking at the original deb packages, it all looks sane.

dwalkes:
`rpmquery --dump tmp/deploy/rpm/armv8a_tegra186/libvisionworks-dev-1.6.0.500n-r0.armv8a_tegra186.rpm | grep pkgconfig
/usr/lib/pkgconfig 0 1550797236 0000000000000000000000000000000000000000000000000000000000000000 040775 root root 0 0 0 X
`

`rpmquery --dump tmp/deploy/rpm/armv8a_tegra186/cuda-command-line-tools-dev-10.0.166+1-r0.armv8a_tegra186.rpm | grep pkgconfig
/usr/lib/pkgconfig 0 1562599705 0000000000000000000000000000000000000000000000000000000000000000 040755 root root 0 0 0 X`

`rpmquery --dump tmp/deploy/rpm/aarch64/spdlog-dev-1.3.1-r0.aarch64.rpm | grep pkgconfig
/usr/lib/pkgconfig 0 1562601919 0000000000000000000000000000000000000000000000000000000000000000 040755 root root 0 0 0 X`

`rpmquery --dump tmp/deploy/rpm/aarch64/librealsense2-dev-2.13.0-r0.aarch64.rpm | grep pkgconfig
/usr/lib/pkgconfig 0 1563284507 0000000000000000000000000000000000000000000000000000000000000000 040755 root root 0 0 0 X`

So libvisionworks /usr/lib/pkgconfig directory is 7**7**5 permissions and the rest are 7**5**5 which is the default for install -d and no mode argument.

madisongh:
Thanks, I missed that last time.  Looks good then.

PR #167, opened 2019-08-13
Fix issue affecting multi-label extlinux.conf files

compenguy:
The value for UBOOT_EXTLINUX_DEFAULT_LABEL needs to be the menu
description for the label (UBOOT_EXTLINUX_MENU_DESCRIPTION), not the
internal identifier for that label.

In the common case of a single entry, it's not necessary to specify a
default, and in the case of multiple entries, it's probably better to
require the distribution developer to provide a value for
UBOOT_EXTLINUX_DEFAULT_LABEL, to explicitly specify the description of
the label that should be default, if a default is required.


madisongh:
The settings currently there should work as-is, though, since the menu description defaults to being the label if not explicitly set, right?  I don't object to removing the default setting, or using `?=` to make it easier to override.

compenguy:
> The settings currently there should work as-is, though, since the menu description defaults to being the label if not explicitly set, right? I don't object to removing the default setting, or using `?=` to make it easier to override.

It would, if there weren't an [upstream bug in uboot-extlinux-config.bbclass](https://bugzilla.yoctoproject.org/show_bug.cgi?id=13469), where they define a default menu description for the label "linux", but because of how they do label processing using OVERRIDES, and because OVERRIDES by default contains ${OS}, which pretty much always evaluates to "linux", the menu description not only defaults to "${DISTRO}", it's actually almost impossible to override currently.

OE has [accepted my patch for those issues into master](https://git.openembedded.org/openembedded-core/commit/?id=42b44fdd9e63b8d39cf83baecec08328275558ff), but I'm not sure they're going to backport it to warrior.

PR #175, opened 2019-09-05
gstreamer1.0-omx-tegra: remove unneeded build-dependency

Ecordonnier:
This removes a significant amount of build dependencies for the image I am building (based on Jetson TX2).


madisongh:
Thanks!  Now I'm wondering whether plugins-bad was ever necessary and I just missed that, or if something changed in either gstreamer or L4T that made it obsolete...

Ecordonnier:
I also wondered and had a quick look at the git history, but I didn't find what could be the reason it used to be a dependency in the past.

PR #176, opened 2019-09-05
cmake: refresh findcuda-dont-reset-cflags.patch

sashko:
Refresh findcuda-dont-reset-cflags.patch for thud, so it doesn't hunk.

Signed-off-by: Oleksandr Kravchuk <oleksandr.kravchuk@pelagicore.com>


madisongh:
Thanks!

PR #184, opened 2019-10-02
libdrm-tegra-shim: replace to enable non-Tegra displays

kekiefer:
In meta-tegra, the libdrm-tegra-shim takes the place of the normal
libdrm. The shim passes through all api calls directly to libdrm-nvdc,
and acts as a build-time dependency where libdrm-nvdc can not. This
approach works fine when the system only uses the Tegra display
drivers, but if a system has other, non-Tegra displays (such as evdi,
udl, or custom DRM drivers), this approach will fail since
libdrm-nvdc can't handle these non-Tegra drivers on its own.

As it turns out, libdrm-nvdc can work in the opposite way. This is
how it is deployed in the JetPack Ubuntu releases. If loaded first,
libdrm-nvdc will internally process calls to any DRM device with
the name tegra-udrm, drm-nvdc, or nvidia-drm (yes, the desktop
driver), but will pass through calls made to access devices with any
other name to the system libdrm.

To enable this functionality, we make sure that libdrm-nvdc is
loaded first by putting an entry in /etc/ld.so.conf.d, and installing
it in addition to libdrm. We let libdrm be the build-time dependency
of any package that needs it, and thus eliminate the need for the
libdrm-tegra-shim. Now Tegra and non-Tegra displays can coexist on
the system.


madisongh:
I like this idea, but I'll need to hold off on pulling it in until I sort out how to get weston to work without the shim.  That should be doable, too, based on a thread I saw on the Jetson forum.

kekiefer:
That's funny -- this is something I checked was still working properly before I put together the change set. I'll double-check again with a fresh build on Monday.

madisongh:
Thanks.  I ran my test on a Nano, and while the sato image came up fine, weston had problems initializing.

kekiefer:
Confirmed suspicions that Weston seems to be ignoring the ld config and getting drmOpen from the libdrm in /usr/lib:

```
(gdb) info sharedlibrary
From                To                  Syms Read   Shared Object Library
...
0x0000007fa0b3b5b0  0x0000007fa0b433d4  Yes         /usr/lib/libweston-7/../libdrm.so.2
...
```

```
root@jetson-xavier:/usr/lib# ldconfig -p | grep libdrm
	libdrm.so.2 (libc6,AArch64) => /usr/lib/tegra/libdrm.so.2
	libdrm.so.2 (libc6,AArch64) => /usr/lib/libdrm.so.2
```

Maybe you have some ideas on the top of your head. If not, I'll keep digging.

kekiefer:
OK, I've figured out why this is happening. `drm-backend.so` has RPATH set to `${ORIGIN}/..` so that's taking precedence. I can get around this by adding `chrpath --delete ${D}${libdir}/libweston*/drm-backend.so` to `do_install_append_tegra`. I see something similar is done in the recipe for mesa, but for different reasons. Can you think of a better way to handle this, maybe across different recipes which might be doing something similar?

kekiefer:
Native meson builds strip rpaths at install, but this is removed by an oe patch to meson (see openembedded/openembedded-core@98a76c3171f9080d6246aac0188ef05c40852adc). So the built libraries keep the rpath used during build, and this gets fixed up later by oe.

As such, another approach that works and is somewhat simpler is to add a weston path that sticks `install_rpath: '../tegra'` into the meson build for the drm-backend.

I don't think either approach smells too bad, but looking forward to feedback. 

I was hoping to find something like adding an additional system-wide library path that gets used during build as well and would magically get picked up by meson (adding another rpath to `BUILD_LDFLAGS` didn't work). It would be nice to have something that could be added to the machine config, but I can't see how to do this.

madisongh:
Hmm.  `BUILD_LDFLAGS` would only affect native builds.  Did you try adding it to `HOST_LDFLAGS` (or `TARGET_LDFLAGS`)?

kekiefer:
Ack, yeah. OK, well that could work. I do notice that TARGET_LDFLAGS no longer contains any rpaths like it used to, so this might be a bigger hammer than we need.

How about this... what about adding an rpath for ${libdir}/tegra to the pkgconfig for the system libdrm? This would fix up any build that leverages the pc for libdrm that doesn't get rpath stripped (like weston), and everything else will use the ld.so.conf.

madisongh:
That sounds like a good approach.

kekiefer:
This is working with QT and Wayland. I am just now seeing that the eglstreams-kms-example is getting hung up in a strange way:

```
#0  __lll_lock_wait (futex=futex@entry=0x7fb7fa7f58, private=0) at lowlevellock.c:52
#1  0x0000007fb7b8ee58 in __GI___pthread_mutex_lock (mutex=0x7fb7fa7f58)
    at pthread_mutex_lock.c:80
#2  0x0000007fb7f826b0 in drmMalloc () from /usr/lib/tegra/libdrm.so.2
#3  0x0000007fb7a7dd28 in drmGetVersion (fd=7) at ../libdrm-2.4.99/xf86drm.c:866
#4  0x0000007fb7f86598 in ?? () from /usr/lib/tegra/libdrm.so.2
#5  0x0000007fb7f870a4 in drmGetVersion () from /usr/lib/tegra/libdrm.so.2
#6  0x0000007fb78cff70 in ?? () from /usr/lib/libnvidia-glsi.so.32.2.1
#7  0x0000007fb7911b7c in ?? () from /usr/lib/libnvidia-glsi.so.32.2.1
#8  0x0000007fb790494c in _nv017glsi () from /usr/lib/libnvidia-glsi.so.32.2.1
#9  0x0000007fb79e8c8c in ?? () from /usr/lib/libEGL_nvidia.so.0
#10 0x0000007fb799f884 in ?? () from /usr/lib/libEGL_nvidia.so.0
#11 0x0000000000401fa0 in GetDrmFd (device=device@entry=0x7fb7a13348) at egl.c:148
#12 0x0000000000401c2c in main () at main.c:46
```

This works on the JetPack release. I need to troubleshoot this and see what is going on here.

kekiefer:
Well, that was fun to track down. It turns out JetPack Ubuntu builds by default (including libdrm) with the -Bsymbolic-functions linker flag. This flag makes sure libdrm's internally-defined non-static functions (like `drmMalloc` in the trace above) are resolved internally instead of possibly calling back into libdrm-nvdc (which is problematic). Here's an article I found very helpful to explain what this flag is doing: https://flameeyes.blog/2012/10/07/symbolism-and-elf-files-or-what-does-bsymbolic-do/

This is working everywhere I've tried now -- QT eglfs, Weston, gstreamer, and eglstreams-kms-example -- consistently for both drm-nvdc and my custom display driver.

madisongh:
Excellent work on tracking that down.  Thanks!

PR #185, opened 2019-10-03
GCC 7 support

yossizap:
I went ahead and merged GCC 7 with the latest patches from openembedded-core/thud with the required fixes for it to work. I have tested this solution on Jetson TX2 with a basic image that contains CUDA recipes and everything seems to work well.

Since thud is still being updated it shouldn't be too challenging to keep the recipe up-to-date with security/bug fixes until there's a new version of CUDA that supports GCC 8/9. I don't mind taking that responsibility.

Hopefully the commits are detailed enough for the porting process to be clear.

Rationale:

Now that GCC 7 is no longer maintained in master/warrior while being required by CUDA, which is most likely essential to most most people, I think that meta-tegra should provide a GCC 7 solution. It's not trivial for an average user to take care of this dependency, yet necessary for those who want newer SDK versions or Xavier support. It might also discourage people from using the latest layer version/openembedded builds at all, I almost gave up on moving to warrior because of it myself.

Thanks in advance!


madisongh:
I know that the CUDA/gcc version compatibility issue makes it hard to stay current with the latest OE-Core.  There are two main reasons why I haven't pulled in the toolchain recipes here.  First, this is really meant to be a BSP layer, so generic toolchain recipes are out of place.  Second, supporting toolchains is time-consuming - you can run into compatibility issues with building for other platforms, or recipes in other layers which folks include in their builds, which might require bbappends and/or patches to work around... it can be a big time-sink.

That said, we could either create a new layer repo, or add in a new layer in, say, a `contrib` subdirectory here, to hold these recipes (and any others), and give folks the option of including that in their layer configuration to make things somewhat easier.

Alternatively, if someone were willing to put together some documentation on the wiki here on using, for example, the `meta-linaro-toolchain` layer to bring in recipes for the older toolchain, that could help, too.

dwalkes:
> if someone were willing to put together some documentation on the wiki here on using, for example, the meta-linaro-toolchain layer to bring in recipes for the older toolchain, that could help, too.

Good idea.  I've taken a first pass at this at https://github.com/madisongh/meta-tegra/wiki/Using-linaro-gcc7-for-CUDA-support

yossizap:
> Alternatively, if someone were willing to put together some documentation on the wiki here on using, for example, the meta-linaro-toolchain layer to bring in recipes for the older toolchain, that could help, too.
While it's useful to have that tutorial(and Dan's wiki page seems to be throughout), I am not a big fan of prebuilt external toolchains since they don't allow modification/patches. Plus, Linaro only ships ARM toolchains, which means there's less compatibility between target and native/other targets using the same user layers.

I agree that a contrib layer/directory with patches/recipes would be great to get people started and ensure things are 'plug and play' for a regular Jetson build. While I agree that maintaining a set of layers for an unsupported compiler isn't always a walk in the park, I do it myself and know from experience that most errors are unsupported flags and macros that are pretty intuitive and have enough search results.

madisongh:
I've pushed a `contrib` directory with a layer configuration.  Could you rework this PR to move the recipes under there?

yossizap:
Rebased and re-arranged, build passes.

Do you want me to add documentation in the wiki and/or README? We can also add a README in `contrib` to clarify it's purpose.

contrib/conf/layer.conf
```diff
@@ -13,3 +13,13 @@ LAYERVERSION_tegra-contrib = "1"
 LAYERDEPENDS_tegra-contrib = "core"
 
 LAYERSERIES_COMPAT_tegra-contrib = "warrior"
+
+# -fmacro-prefix-map isn't supported in GCC 7
```
madisongh: Rather than adding these settings here, could you create another config file for them, under, say `contrib/conf/include`?  I'd rather the flags additions be opt-in separately from adding the layer (which could expand to include other optional stuff).

madisongh:
Once I pull this in, I'll update the top-level README and add a `contrib` README for some documentation.

yossizap:
Should be good now, reran everything as well.

madisongh:
Great, thanks

Austriker:
@yossizap Thanks it's a great idea.
@madisongh have you updated the Readme ? I have only found the one explaining the linaro version.

madisongh:
@Austriker Done. Thanks for the reminder!

Ecordonnier:
For reference, the wiki page for using the contrib gcc7 is there: https://github.com/madisongh/meta-tegra/wiki/Using-gcc7-from-the-contrib-layer

PR #186, opened 2019-10-06
Update README for Linaro GCC 7 support

dwalkes:
Add a link to the wiki page describing use of Linaro for GCC7
See discussion in #185


madisongh:
Thanks!

PR #211, opened 2019-11-18
Set IMAGE_ROOTFS_EXTRA_SPACE to zero in initramfs

arnstein:
I set IMAGE_ROOTFS_EXTRA_SPACE in my local.conf to have some extra space while developing and testing my build. Unfortunately this gave me this error while building:

```
ERROR: tegra-minimal-initramfs-1.0-r0 do_image_cpio: The initramfs size 5251072(K) overrides INITRAMFS_MAXSIZE: 131072(K)
ERROR: tegra-minimal-initramfs-1.0-r0 do_image_cpio: You can set INITRAMFS_MAXSIZE a larger value. Usually, it should
ERROR: tegra-minimal-initramfs-1.0-r0 do_image_cpio: be less than 1/2 of ram size, or you may fail to boot it.
```

Setting IMAGE_ROOTFS_EXTRA_SPACE to zero in tegra-minimal-initramfs.bb seems to have fixed this, and now I have lots of glorious extra space on my rootfs!


madisongh:
Thanks!

PR #217, opened 2019-12-03
Properly put the nvargus-daemon in the tegra-libraries-argus package.

nielsavonds:
This currently made the tegra-libraries package depend on the tegra-libraries-argus package, because the binary would require the libraries. It was impossible to not include tegra-libraries-argus


madisongh:
Thanks!

PR #221, opened 2019-12-10
fixed potential typo in tegra186-flash-helper.sh

danielwangksu:
It seems to me `boardrev` should be `board_version`


madisongh:
Thanks... missed that one when I renamed that variable.

PR #222, opened 2019-12-10
back port signed image support from master branch

danielwangksu:
I tested on a TX2 and a TX2i board. The patch works but not sure if it is the same for tegra210.  

Also, to my best knowledge, I cannot find a canonical URI for secureboot_r32.1.tbz2 (not part of jetpack for some reason). Also the download requires Nvidia account login, so I changed the SRC_URI to a local mirror instead. It works for me, but if this is not desired I can change it.


recipes-bsp/tegra-binaries/tegra-binaries-32.1.0.inc
```diff
@@ -2,13 +2,21 @@ LICENSE = "Proprietary"
...
+L4T_SHA256SUM = "e7ecaa32928484398d2492e6e9911262b4663e720b4867404d8fe1928fe483aa"
+SB_MD5SUM = "8c61778838533cc20a9a248882091a46"
+SB_SHA256SUM = "f1b8dedf6544c0588997430333553a5b3973a0b826cf21771d84eabce50a31d6"
+SRC_URI[l4t.md5sum] = "${L4T_MD5SUM}"
+SRC_URI[l4t.sha256sum] = "${L4T_SHA256SUM}"
+SRC_URI[sb.md5sum] = "${SB_MD5SUM}"
+SRC_URI[sb.sha256sum] = "${SB_SHA256SUM}"
+
+# L4T 32.1 does not support Jetson Nano secureboot (L4T 32.2 support Jetson Nano secureboot)
+SRC_URI_tegra210 = "${L4T_URI_BASE}/${L4T_BSP_PREFIX}_Linux_R${PV}_aarch64.tbz2;"
```
madisongh: I don't think this is going to work - SRC_URI checksums won't be there as needed.  We might need some python to handle this.

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
```diff
@@ -40,13 +54,26 @@ do_install() {
     install -m 0755 ${S}/bootloader/nv_smd_generator ${D}${BINDIR}
     install -m 0755 ${S}/bootloader/BUP_generator.py ${D}${BINDIR}
     sed -i -e's,^#!/usr/bin/python,#!/usr/bin/env python,' ${D}${BINDIR}/BUP_generator.py
-    install -m 0755 ${S}/bootloader/rollback/rollback_parser.py ${D}${BINDIR}
-    sed -i -e's,^#!/usr/bin/python,#!/usr/bin/env python,' ${D}${BINDIR}/rollback_parser.py
+    install -d ${D}${BINDIR}/rollback
+    install -m 0755 ${S}/bootloader/rollback/rollback_parser.py ${D}${BINDIR}/rollback
+    sed -i -e's,^#!/usr/bin/python,#!/usr/bin/env python,' ${D}${BINDIR}/rollback/rollback_parser.py
```
madisongh: In master, the script just goes into `${BINDIR}` instead of a subdirectory.  Any reason for changing this?

classes/image_types_tegra.bbclass
```diff
@@ -587,23 +615,34 @@ oe_make_bup_payload_common() {
     rm -f ./slot_metadata.bin
     cp ${STAGING_DATADIR}/tegraflash/slot_metadata.bin ./
     mkdir ./rollback
-    ln -sf ${STAGING_BINDIR_NATIVE}/tegra186-flash/rollback_parser.py ./rollback/
+    ln -sf ${STAGING_BINDIR_NATIVE}/tegra186-flash/rollback/rollback_parser.py ./rollback/
     ln -snf ${STAGING_DATADIR}/nv_tegra/rollback/t${@d.getVar('NVIDIA_CHIP')[2:]}x ./rollback/
-    ln -sf ${STAGING_BINDIR_NATIVE}/tegra186-flash/BUP_generator.py ./
+    # ln -sf ${STAGING_BINDIR_NATIVE}/tegra186-flash/BUP_generator.py ./
```
madisongh: Just go ahead and delete this line

classes/image_types_tegra.bbclass
```diff
@@ -587,23 +615,34 @@ oe_make_bup_payload_common() {
     rm -f ./slot_metadata.bin
     cp ${STAGING_DATADIR}/tegraflash/slot_metadata.bin ./
     mkdir ./rollback
-    ln -sf ${STAGING_BINDIR_NATIVE}/tegra186-flash/rollback_parser.py ./rollback/
+    ln -sf ${STAGING_BINDIR_NATIVE}/tegra186-flash/rollback/rollback_parser.py ./rollback/
```
madisongh: See my comment below about where this gets installed - if that gets changed, this should get changed back to the original

recipes-bsp/tegra-binaries/tegra-binaries-32.1.0.inc
danielwangksu: Sorry for the delay. I'm sorry I did not get what you mean?

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
danielwangksu: The reason I move it into `${BINDIR}/rollback` folder, is because I want to run `doflash.sh` and `tegra186-flash-helper.sh` in the generated `core-image-tx2-minimal-jetson-tx2i.tegraflash.zip` to sign BUP-payload.

The l4t_bup_gen.func used by the `tegra186-flash-helper.sh` assume the path `/rollback/rollback_parser.py`. here is the function:

l4t_bup_gen.func line 30-39:
```
_parse_rollback_info()
{
	local tegra_soc="${1}";
	local _rollback_bin=rollback.bin
	local _rollback_cfg="${OUT}"/rollback/"${tegra_soc}"/rollback.cfg
	if [ -f "${_rollback_bin}" ]; then
		rm "${_rollback_bin}"
	fi
	"${OUT}"/rollback/rollback_parser.py "${PRODUCT_NAME}" "${_rollback_cfg}"
}
```

classes/image_types_tegra.bbclass
danielwangksu: sure

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
madisongh: OK, I see.  There's still no need to move rollback_parser.py here - if you look at the line you are also changing in image_types_tegra.bbclass, you'll see that the symlink that gets created goes into the expected `rollback` subdirectory.

To make it possible to do BUP generation from a tegraflash package, there are other changes needed, though.

recipes-bsp/tegra-binaries/tegra-binaries-32.1.0.inc
madisongh: This patch is replacing the single-element SRC_URI with a multiple-element one, but with the `tegra210` overrides you aren't matching up with the new structure.

You need to add `name=l4t` to this line to match up with the `SRC_URI[l4t.md5sum]` and `SRC_URI[l4t.sha256sum]` settings, and you need to rename the settings in the next two lines to be `L4T_MD5SUM_tegra210` and `L4T_SHA256SUM_tegra210` as well.

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
```diff
@@ -7,6 +7,20 @@ SSTATE_SWSPEC = "sstate:tegra-binaries-native::${PV}:${PR}::${SSTATE_VERSION}:"
...
+           file://nvphs.init \
+           file://nvphs.service \
+           file://nvs-service.init \
+           file://nvs-service.service \
+           file://nvstartup.init \
+           file://nvstartup.service \
+           file://tegra186-flash-helper.sh \
+           file://tegra194-flash-helper.sh \
+           file://tegra210-flash-helper.sh \
+           "
```
madisongh: Why are you adding these files to this recipe?

recipes-bsp/tegra-binaries/tegra-binaries-32.1.0.inc
danielwangksu: Got it I will change accordingly. Thank you!

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
danielwangksu: Oh because I saw you added those files to this recipe,  recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.2.1.bb

in this commit https://github.com/madisongh/meta-tegra/commit/c45dc8bae40f533aa7cfd062f3b5d86069f85259

recipes-bsp/tegra-binaries/tegra-binaries-32.1.0.inc
danielwangksu: For tegra210 I think SRC_URI is replaced by the following, because `SRC_SHA256SUM_tegra210` I thought bitbake should ignore `L4T_MD5SUM`. Is my assumption correct:

```
SRC_URI_tegra210 = "${L4T_URI_BASE}/${L4T_BSP_PREFIX}_Linux_R${PV}_aarch64.tbz2"
SRC_MD5SUM_tegra210 = "15fda16d83e9a189ea5534108f57af2f"
SRC_SHA256SUM_tegra210 = "e4902a952245217227486322b3e8cef47942f87f2ea91f9e67d700314ae3ade3"
```

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
madisongh: Hmm, now I have to ask myself why I did that. :)

recipes-bsp/tegra-binaries/tegra-binaries-32.1.0.inc
madisongh: No, because we aren't setting `SRC_URI[md5sum]` and `SRC_URI[sha256sum]`.  I don't think you can put overrides on flag settings like that.  That's why I recommended the other approach.

danielwangksu: Thank you I will change it

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
danielwangksu: If I remember correct, I think this create the `STAGING_BINDIR_NATIVE` variable and use in this line, but not very sure :):

in image_types_tegra.bbclass
```
cp -R ${STAGING_BINDIR_NATIVE}/tegra186-flash/* .
```

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
danielwangksu: Hi @madisongh I revert the this changes back, bitbake can generate payload becasue of oe_make_bup_payload_common() creates symbolic link (through `ln -sf ${STAGING_BINDIR_NATIVE}/tegra186-flash/rollback_parser.py ./rollback/`)

But the final tegraflash.zip does not have `rollback_parser.py` in rollback folder. Is it OK to put `rollback_parser.py` in the rollback folder instead? So that we can run BUP-payload off bitbake?

recipes-bsp/tegra-binaries/tegra186-flashtools-native_32.1.0.bb
madisongh: Yeah, I think this was an oversight on my part... I think I was using `devtool`, which snuck in the duplicate entries on me.  No real harm in keeping them here to match, and I have a bigger overhaul of these recipes in the works coming soon to master anyway, so this is fine as-is.

danielwangksu:
Thank you very much! Let me know if anything need to be changed. Happy new year!

madisongh:
Squashed and merged manually, after fixing a couple of issues that cropped up in my testing.

PR #227, opened 2019-12-20
opencv: update CUDA_ARCH_BIN to work for all tegra models

kekiefer:
Rather than have a per-architecture CUDA_ARCH_BIN, this variable can
be set as a list, based on the nvidia build script found here:
https://github.com/AastaNV/JEP/blob/master/script/install_opencv4.1.1_Jetson.sh

Additionally, this change adds the CUDA_ARCH_BIN "7.2" for t194.


kekiefer:
Note that I have tested this change as working on the Xavier. It should work for others, but I have not tried it. It's a simple check:
```
python3 -c "import cv2; cv2.cuda.printCudaDeviceInfo(0)"
```
```

madisongh:
We definitely need to fix the settings for Xavier -- thanks for catching this -- but it should follow the existing pattern so just the Xavier-specific CUDA code is included for that platform to keep the target footprint smaller.  Otherwise all of the platforms will include binaries for GPUs they don't have.

Can you update this to just add an override for `tegra194`?

kekiefer:
Updated

madisongh:
Thanks!

PR #231, opened 2020-01-07
Correct nvpmodel support for xavier 8gb

kekiefer:
The xavier 8gb module uses a different nvpmodel.conf than its big brother. This change installs the right version, based on a specific `MACHINEOVERRIDES` for this module.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-nvpmodel_32.2.3.bb
+++ b/recipes-bsp/tegra-binaries/tegra-nvpmodel_32.2.3.bb
@@ -17,6 +17,7 @@ do_compile[noexec] = "1"
 NVPMODEL = ""
 NVPMODEL_tegra186 ?= "nvpmodel_t186"
 NVPMODEL_tegra194 ?= "nvpmodel_t194"
+NVPMODEL_xavier-8gb ?= "nvpmodel_t194_8gb"
 NVPMODEL_tegra210 ?= "nvpmodel_t210"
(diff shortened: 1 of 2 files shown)


recipes-bsp/tegra-binaries/tegra-nvpmodel_32.2.3.bb
```diff
@@ -17,6 +17,7 @@ do_compile[noexec] = "1"
 NVPMODEL = ""
 NVPMODEL_tegra186 ?= "nvpmodel_t186"
 NVPMODEL_tegra194 ?= "nvpmodel_t194"
+NVPMODEL_xavier-8gb ?= "nvpmodel_t194_8gb"
```
madisongh: Any reason you couldn't just have used `jetson-xavier-8gb` as the override here?

kekiefer: This serves to identify the SOM regardless of the MACHINE variable. A customizing layer will set a new MACHINE to identify a unique base board, but we still need this override to apply.

If you have a better idea for the name that identifies the SOM I’m not married to this. I’ll just blame NVIDIA marketing for the lazy naming that doesn’t attempt to separate out modules and their own base boards :)

madisongh: OK, makes sense. "xavier-8gb" is certainly more understandable than "galen-8gb" or "p2888-0006".

PR #234, opened 2020-01-16
Add source build for most current cboot (32.2.3)

kekiefer:
I'd like to build cboot from source because the latest binary included with 32.3.1 has support for extlinux enabled, which adds a long delay to boot that I'd rather avoid for production builds. It might make sense to add extlinux support to meta-tegra xavier builds, but that's a different issue.

There are newer versions of cboot available in /some/ L4T releases, but unfortunately not the latest. The latest source release of cboot is in 32.2.3, so we have to add back the tegra-sources for this release.

A common build infrastructure is created to reduce code duplication.

Also, the 31.1.0 build was broken, so a change was made to ignore those errors in this PR. Though for me, this version dies with a `DATA ABORT` just before loading linux.

The 32.2.3 build works against 32.3.1 on Xavier 8GB and I can disable extlinux support. I don't have a TX2 handy right now, so I can't test the 28.3.0 version.


madisongh:
Looks good.  Just dropping the 31.1.0 recipe altogether is probably a good idea at this point. I expect the 28.3 from-source build for TX2 won't work properly with the 32.x BSP, either, so we could probably drop that as well.

Thanks for this, it's something I'd been meaning to do for a while!

PR #235, opened 2020-01-16
Renamed machine folder to match new jetson machine name

MikeLoder:
Small correction to the jetson-nano machine-specific files. They now match the new jetson-nano-qspi-sd machine name.


madisongh:
Thanks!

PR #248, opened 2020-02-07
Add nvidia-docker environment

paroque28:


paroque28:
All these commits were needed to run this:
```bash
root@jetson-nano-qspi-sd:~# docker run -v /usr/bin/cuda-samples/:/usr/bin/cuda-samples/ -it --runtime nvidia nvcr.io/nvidia/l4t-base:r32.3
root@85f6ce9ce7d5:/# /usr/bin/cuda-samples/deviceQuery
/usr/bin/cuda-samples/deviceQuery Starting...

 CUDA Device Query (Runtime API) version (CUDART static linking)

Detected 1 CUDA Capable device(s)

Device 0: "NVIDIA Tegra X1"
  CUDA Driver Version / Runtime Version          10.0 / 10.0
  CUDA Capability Major/Minor version number:    5.3
  Total amount of global memory:                 3956 MBytes (4148076544 bytes)
  ( 1) Multiprocessors, (128) CUDA Cores/MP:     128 CUDA Cores
  GPU Max Clock rate:                            922 MHz (0.92 GHz)
  Memory Clock rate:                             1600 Mhz
  Memory Bus Width:                              64-bit
  L2 Cache Size:                                 262144 bytes
  Maximum Texture Dimension Size (x,y,z)         1D=(65536), 2D=(65536, 65536), 3D=(4096, 4096, 4096)
  Maximum Layered 1D Texture Size, (num) layers  1D=(16384), 2048 layers
  Maximum Layered 2D Texture Size, (num) layers  2D=(16384, 16384), 2048 layers
  Total amount of constant memory:               65536 bytes
  Total amount of shared memory per block:       49152 bytes
  Total number of registers available per block: 32768
  Warp size:                                     32
  Maximum number of threads per multiprocessor:  2048
  Maximum number of threads per block:           1024
  Max dimension size of a thread block (x,y,z): (1024, 1024, 64)
  Max dimension size of a grid size    (x,y,z): (2147483647, 65535, 65535)
  Maximum memory pitch:                          2147483647 bytes
  Texture alignment:                             512 bytes
  Concurrent copy and kernel execution:          Yes with 1 copy engine(s)
  Run time limit on kernels:                     Yes
  Integrated GPU sharing Host Memory:            Yes
  Support host page-locked memory mapping:       Yes
  Alignment requirement for Surfaces:            Yes
  Device has ECC support:                        Disabled
  Device supports Unified Addressing (UVA):      Yes
  Device supports Compute Preemption:            No
  Supports Cooperative Kernel Launch:            No
  Supports MultiDevice Co-op Kernel Launch:      No
  Device PCI Domain ID / Bus ID / location ID:   0 / 0 / 0
  Compute Mode:
     < Default (multiple host threads can use ::cudaSetDevice() with device simultaneously) >

deviceQuery, CUDA Driver = CUDART, CUDA Driver Version = 10.0, CUDA Runtime Version = 10.0, NumDevs = 1
Result = PASS
root@85f6ce9ce7d5:/# /usr/bin/cuda-samples/UnifiedMemoryStreams
GPU Device 0: "NVIDIA Tegra X1" with compute capability 5.3

Executing tasks on host / device
Task [0], thread [0] executing on device (363)
Task [1], thread [1] executing on device (791)
Task [3], thread [2] executing on device (414)
Task [2], thread [3] executing on device (753)
Task [4], thread [0] executing on device (762)
Task [5], thread [2] executing on device (756)
Task [6], thread [1] executing on device (952)
Task [7], thread [3] executing on device (763)
Task [8], thread [0] executing on device (942)
Task [9], thread [2] executing on device (853)
Task [10], thread [1] executing on device (463)
Task [11], thread [3] executing on device (375)
Task [12], thread [0] executing on device (102)
Task [13], thread [2] executing on device (937)
Task [14], thread [1] executing on host (64)
Task [15], thread [3] executing on device (538)
Task [16], thread [0] executing on device (922)
Task [17], thread [1] executing on host (86)
Task [18], thread [2] executing on device (298)
Task [19], thread [3] executing on device (693)
Task [20], thread [1] executing on device (686)
Task [21], thread [0] executing on device (673)
Task [22], thread [2] executing on device (758)
Task [23], thread [3] executing on device (911)
Task [25], thread [0] executing on device (899)
Task [24], thread [1] executing on device (272)
Task [26], thread [2] executing on device (918)
Task [27], thread [3] executing on host (64)
Task [28], thread [0] executing on device (697)
Task [29], thread [1] executing on device (321)
Task [30], thread [2] executing on device (923)
Task [31], thread [3] executing on device (591)
Task [32], thread [0] executing on device (248)
Task [33], thread [1] executing on device (948)
Task [34], thread [2] executing on device (853)
Task [35], thread [3] executing on device (570)
Task [36], thread [0] executing on device (680)
Task [37], thread [1] executing on device (165)
Task [38], thread [2] executing on device (225)
Task [39], thread [3] executing on device (970)
All Done!



```

paroque28:
Sorry, I just found out that I modified libvisionworks_1.6.0.500 but didn't test it. I will  report back as soon as I have more information

paroque28:
It works good now! 👍

madisongh:
@paroque28 Thanks for all the hard work.  I've looked at these changes a bit and have some proposed modifications on [this branch](https://github.com/madisongh/meta-tegra/tree/wip-container-32.3.1).  The big change is to move the generation of the CSV files into a bbclass and package the CSV files separately from the main packages... still trying to figure out a few things, like how to get the packages with the CSV files included in an image when they're needed.  I also tweaked some dependencies to make things buildable without having to add in a bunch of layers that might not be wanted.

When I'm looking at what's in the resulting container (the same one downloaded from NVIDIA's container repository), though, I'm a bit concerned about the fact that the NVIDIA container uses Debian-style library paths which aren't mapping to the non-Debian paths we're using.  Not sure if it's really an issue, since programs do appear to run properly in the container, but if I'm understanding the intent of the libnvidia-container "Jetson" support, it looks like they're trying to keep the container size down by providing pass-through mounts for specific paths... and while at runtime the dynamic loader is finding the underlying libraries all right, there may be other use cases that won't work due to the different paths.

paroque28:
@madisongh Great, thanks :) I wasn't quite sure how to implement the CSV files generation but your way looks neater.

Regarding the library paths, you are right. The containers don't like much the fact that I have multilib enabled. I need to set manually the LD_PATH env variable: `-e LD_LIBRARY_PATH=/usr/local/cuda-10.0/lib64/` so it is a good idea to set a fixed path.

wremie:
Many thanks to both of you @paroque28 and @madisongh for supporting this. We have been very interested in this feature for quite some while now. In the effort to get a clean build i stumble upon some errors compiling docker-ce. It has something to do with GCCVERSION = "7.%" as build completes without this perfectly fine. Following compilation task is failing;
`meta-virtualization/recipes-containers/tini/tini_0.18.0.bb:do_compile`
Seems like I'm missing some stuff on the build host but I'm unable to find out the exact cause. `aarch64-poky-linux/gcc/aarch64-poky-linux/7.3.0/ld: cannot find rcrt1.o: No such file or directory
collect2: error: ld returned 1 exit status`. Could it be I'm missing some C libraries? 

[log.do_compile.txt](https://github.com/madisongh/meta-tegra/files/4212476/log.do_compile.txt)

triblex:
I found that these lines of the bottom of the `l4t.csv` file:

```
sym, ${base_libdir}/libv4l2.so.0.0.999999
sym, ${base_libdir}/libv4lconvert.so.0.0.999999
sym, ${base_libdir}/libv4l/plugins/libv4l2_nvvidconv.so
sym, ${base_libdir}/libv4l/plugins/libv4l2_nvvideocodec.so
```

Needs to be changed to:

```
sym, ${libdir}/libv4l2.so.0
sym, ${libdir}/libv4lconvert.so.0
lib, ${libdir}/libv4l/plugins/libv4l2_nvvidconv.so
lib, ${libdir}/libv4l/plugins/libv4l2_nvvideocodec.so
```

And that the line just above it:

`sym, /usr/lib/libdrm_nvdc.so`

Is a symlink that needs to be created to:

`/usr/lib/tegra/libdrm.so.2`

paroque28:
Good finding! :)  Do you want me to create the commit for this or can you do it?

triblex:
I'll investigate a bit more and create a commit :)

madisongh:
>I found that the nvidia-container-runtime needs ldconfig for starting the container.

It's one of the "backfill" features, so it should get included by default unless you've explicitly added it to DISTRO_FEATURES_BACKFILL_CONSIDERED.

Since it's a distro-wide setting, the best we can do is have the recipe `inherit distro_features_check` and set `REQUIRED_DISTRO_FEATURES = "ldconfig"`.

paroque28:
Hi @madisongh ,
So I found that cuda.csv, cuddn.csv, tensorrt.csv and visionworks-sfm.csv are not added by default when you add nvidia-container runtime and these packages (cuda, tensorrt and visionworks).

This is my proposal:
For instance, add into cuda.bb
```
PACKAGECONFIG ??= "${PN}-container-csv"
PACKAGECONFIG[container-csv] = ",,, ${PN}-container-csv"
```
Note that I am adding by default the container-csv. 
It does not break cuda to have this csv file but it does break nvidia-container-runtime not having these csv files.

What do you think? The idea is to make the same with tensorrt and so on...

madisongh:
>So I found that cuda.csv, cuddn.csv, tensorrt.csv and visionworks-sfm.csv are not added by default when you add nvidia-container runtime and these packages (cuda, tensorrt and visionworks).

How about we just add an RDEPENDS to the `nvidia-container-runtime` recipe on all of the `-container-csv` packages?

paroque28:
I don't like this approach as that would force the build for tensorrt, cddn and more... We don't want this as the user might not have all the .deb files downloaded and would waste time and resources in build time.
And also, that would make nvidia-container-runtime look for all these libraries each time it looks creates a container.

What do you think?

madisongh:
> I don't like this approach as that would force the build for tensorrt, cddn and more... We don't want this as the user might not have all the .deb files downloaded and would waste time and resources in build time.

That is a downside, but since we can't know in advance what containers will need which libraries, making all of the them available as the default would be less likely to cause problems caused by some being missing when an app in the container starts running.  Sigh.

How about a modification of your proposed approach - instead of using PACKAGECONFIG, we automatically add `${PN}-container-csv` to `RDEPENDS_${PN}` if `virtualization` is in `DISTRO_FEATURES`?  That could probably be done in the bbclass.

paroque28:
Yes, that actually makes sense Madison, and we can document that

madisongh:
I've updated by `wip-container-32.3.1` branch to implement the RDEPENDS additions as we discussed above, and I also added a patch to libnvidia-container-tools to handle the mapping between the debian-style paths inside the container and our non-debian-style paths.  There was logic do that in the code already, it just wasn't being applied to the jetson-specific addon mappings from the CSV files.

There's still some adjustments needed - the l4t CSV needs cleanup, at least, and other differences between the from-source builds we do and NVIDIA's pre-packaged binaries have to be accounted for as well.

madisongh:
I ran some more tests and have pushed more changes to `wip-container-32.3.1` to fix the rest of the path mappings that were broken.  The `l4t.csv` file now just maps the `/dev` nodes, and the rest of the recipes generate their own CSVs.

There are still a couple of issues that I can see:
* The `ldconfig` configuration files in the NVIDIA containers prioritize `/usr/lib/aarch64-linux-gnu` over `/usr/lib/aarch64-linux-gnu/tegra`, so the version of `libdrm.so.2` you get is the wrong one.  You can work around that by manually running `ldconfig /usr/lib/aarch64-linux-gnu/tegra` inside the container once it starts.
* Loading the Deepstream container, I see a couple of the gstreamer plugins get blacklisted because they can't find `libcublas.so.10.0` and `libnppig.so.10.0`.  I don't know whether that's a problem unique to our setup, or whether it also affects folks using that container on stock L4T, but `ldconfig /usr/local/cuda-10.0/lib` cleared up the issue.

I'd appreciate it if other folks could give this a try.

madisongh:
I have pushed more changes to the branch to resolve the issues I saw yesterday.  I've reworked the libdrm/libdrm-nvdc integration so it's more like how NVIDIA does it in the stock BSP, which is more compatible with the container setup.  I also added a config file so that ldconfig will configure in the CUDA libraries directory.

The only other thing I'm seeing now is the mismatch between the gstreamer plugins we build from source and the version of gstreamer they use in the container.  So if you want to use gstreamer in the NVIDIA containers, forward porting the gstreamer 1.14 recipes from an older version of OE-Core will be required.

Otherwise, I think the `wip-container-32.3.1` branch is now handling everything like it should.

madisongh:
After a bit more testing, I found and fixed a few more issues - particularly with `libv4l2`, which was using a compiled-in path to locate plugins.

After forward-porting the gstreamer 1.14 recipes from OE-Core warrior branch, I've now been able to successfully run some Deepstream SDK tests within the container, as well as run gstreamer pipelines to capture video from my TX2's camera and display video.

My test distro setup is [here](https://github.com/madisongh/test-distro/tree/zeus-container-32.3.1).

paroque28:
Thanks for the hard work @madisongh  ! 
I will try all these changes right away, Do you still need help trying the issue with ldconfig?

madisongh:
Thanks @paroque28. I think I got the ldconfig issue straightened out with my latest changes.  Give it a try, I'm not 100% sure it will work properly in a multilib config like yours.

dremsol:
Hi @madisongh & @paroque28, thanks for the efforts. We will try on Jetson Nano eMMC with Raspi v2 cam first thing tomorrow.

madisongh:
I've tested on my Nano and Xavier dev kits and it's working there as well.  I've pushed a couple more fixes to pull in some kernel modules needed for docker and to fix another occurrence of compiled-in paths (in gstreamer1.0-omx-tegra).

For accessing the USB camera on my Nano and Xavier dev kits, I added `--device /dev/video0:/dev/video0` to the docker command line.

Commands I used to run the Deepstream test apps:
```
$ sudo docker run -it --rm --net=host --runtime nvidia  -e DISPLAY=$DISPLAY --device /dev/video0:/dev/video0 -v /tmp/.X11-unix/:/tmp/.X11-unix -v /tmp/argus_socket:/tmp/argus_socket nvcr.io/nvidia/deepstream-l4t:4.0.2-19.12-samples
# cd deepstream_sdk_4.0.2_jetson/sources/apps/deepstream-test1
# /opt/nvidia/deepstream/deepstream-4.0/bin/deepstream-test1-app $HOME/deepstream_sdk_v4.0.2_jetson/samples/streams/sample_720p.h264
# cd ../deepstream-test2
# /opt/nvidia/deepstream/deepstream-4.0/bin/deepstream-test2-app $HOME/deepstream_sdk_v4.0.2_jetson/samples/streams/sample_720p.h264
```
The test2 app in particular takes a very long time to run (less so on Xavier), so be patient.

paroque28:
Now that you mention the multilib, I was thinking we should move all the Nvidia paths to lib and forget about lib64.

For instance insted of running:
`docker run --runtime nvidia -e LD_LIBRARY_PATH=/usr/local/cuda-10.0/lib64/ -it tianxiang84/l4t-base:all`

I could run:

`docker run --runtime nvidia -it tianxiang84/l4t-base:all`

What do you think should be the right fix?
Or maybe making a disclaimer that you should't use multilib and disabling that from the yocto build.

madisongh:
With the change ea3bcb46c0b6443d502ba171a61a3c3297080b9a you shouldn't need the LD_LIBRARY_PATH setting any longer, and it should work with a multilib build.

That said, I don't think there's much support upstream (i.e., from NVIDIA) for running 32-bit apps any more (there was some, a few years ago), so just documenting that multilib isn't recommended would be a good idea.

madisongh:
I've cleaned up the patch series from my WIP branch and have PRs pending for both master and zeus-l4t-r32.3.1.  I'll merge them over the weekend unless I hear any objections.

paroque28:
As you mentioned, the LD_LIBRARY_PATH setting is no longer needed. I have tested it and works great. I think we can close this pull request

simabuaboy:
hi guys can i chime in a little, i've been building the recipe for some time from this [test](https://github.com/madisongh/test-distro/tree/zeus-container-32.3.1) of yours @madisongh , and got this error when the finish build trying to run a docker image
 `root@jetson-tx2:~# docker run hello-world 
Unable to find image 'hello-world:latest' locally
latest: Pulling from library/hello-world
256ab8fe8778: Pull complete 
Digest: sha256:f9dfddf63636d84ef479d645ab5885156ae030f611a56f3a7ac7f2fdd86d7e4e
Status: Downloaded newer image for hello-world:latest
docker: Error response from daemon: failed to create endpoint angry_mirzakhani on network bridge: failed to add the host (veth000663a) <=> sandbox (veth96a2687) pair interfaces: operation not supported.
ERRO[0008] error waiting for container: context canceled `

my local.conf :
`MACHINE ?='jetson-tx2'
PREFERRED_PROVIDER_virtual/kernel = "linux-tegra"
GCCVERSION = "7.%"
require contrib/conf/include/gcc-compat.conf
IMAGE_CLASSES += "image_types_tegra"
IMAGE_FSTYPES = "tegraflash"
NVIDIA_DEVNET_MIRROR ?= "file:///mnt/datani/nvidia/tx2"
IMAGE_INSTALL_append = " openssh nvidia-container-runtime cuda-samples nvidia-docker cudnn tensorrt libvisionworks libvisionworks-sfm libvisionworks-tracking"
CORE_IMAGE_EXTRA_INSTALL += "libgomp libgomp-dev libgomp-staticdev"

DISTRO_FEATURES_append = " x11 systemd virtualization"
IMAGE_INSTALL_append = " cuda-samples usbutils" `

can i know the local.conf that you use? or some best practice ? thanks

madisongh:
@simabuaboy I ran into the same error; docker needed `kernel-module-veth` to be present.  I don't think I committed that back to the test-distro branch, though.  I ran through some further tests and found several more kernel modules that need to be present for certain docker features to work. The latest changes in my PRs should cover everything, but when in doubt, just add `kernel-modules` to your image.

madisongh:
See #266 and #267

PR #269, opened 2020-03-22
Small updates for the zeus-l4t-r32.3.1 branch

focusware-develop:


contrib/recipes-devtools/gcc/gcc-7.3.inc
```diff
@@ -72,6 +72,7 @@ SRC_URI = "\
            file://0047-sync-gcc-stddef.h-with-musl.patch \
            file://0048-gcc-Enable-static-PIE.patch \
            file://0049-gcc-override-TARGET_LIBC_PROVIDES_SSP.patch \
+           file://0050-gcc-ustat-compiler-error-fix.patch \
```
madisongh: I'm fine with adding a patch for this, but I'd rather back-port the patch that went into gcc 7.4.0 upstream.

recipes-bsp/tegra-binaries/tegra-redundant-boot_32.3.1.bb
```diff
@@ -50,12 +52,13 @@ PACKAGES = "${PN}-nvbootctrl ${PN} ${PN}-dev"
 FILES_${PN}-nvbootctrl = "${sbindir}/nvbootctrl"
 FILES_${PN} += "/opt/ota_package"
 FILES_${PN}-dev = "${datadir}/nv_tegra/rollback"
-SYSTEMD_SERVICE_${PN} = "nv_update_verifier.service"
+SYSTEMD_SERVICE_${PN} = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'nv_update_verifier.service', '', d)}"
```
madisongh: This shouldn't be necessary; systemd.bbclass should handle this for us.

recipes-bsp/tegra-binaries/tegra-redundant-boot_32.3.1.bb
```diff
@@ -20,9 +20,11 @@ do_install() {
 	install -m 0755 ${B}/usr/sbin/nvbootctrl ${D}${sbindir}
 	install -m 0755 ${B}/usr/sbin/nv_bootloader_payload_updater ${D}${sbindir}
 	install -m 0755 ${B}/usr/sbin/nv_update_engine ${D}${sbindir}
-	install -d ${D}${systemd_system_unitdir}
-	install -m 0644 ${B}/etc/systemd/system/nv_update_verifier.service ${D}${systemd_system_unitdir}
-	sed -i -e's,^After=nv\.service,After=nvstartup.service,' ${D}${systemd_system_unitdir}/nv_update_verifier.service
+	if ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'true', 'false', d)}; then
```
madisongh: This shouldn't be needed - systemd.bbclass should clean this up for us if systemd isn't being used.  What would be needed is the addition of an initscript equivalent for this systemd service... although I'm about at the point of dumping the initscripts completely and just making systemd a required distro feature.  NVIDIA stopped providing initscripts for their services when they started basing their BSP on Ubuntu 16.04, and I don't really have the bandwidth to maintain and test sysvinit-only builds.

recipes-bsp/tegra-binaries/tegra-redundant-boot_32.3.1.bb
```diff
@@ -50,12 +52,13 @@ PACKAGES = "${PN}-nvbootctrl ${PN} ${PN}-dev"
...
-SYSTEMD_SERVICE_${PN} = "nv_update_verifier.service"
+SYSTEMD_SERVICE_${PN} = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'nv_update_verifier.service', '', d)}"
 SYSTEMD_SERVICE_${PN}_tegra210 = ""
 RDEPENDS_${PN}-nvbootctrl = "setup-nv-boot-control"
 RDEPENDS_${PN}-nvbootctrl_tegra210 = ""
 ALLOW_EMPTY_${PN}-nvbootctrl_tegra210 = "1"
-RDEPENDS_${PN} = "${PN}-nvbootctrl setup-nv-boot-control-service"
-RDEPENDS_${PN}_tegra210 = "setup-nv-boot-control-service python-argparse python-core python-subprocess python-textutils python-shell"
+RDEPENDS_${PN} = "${PN}-nvbootctrl ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'setup-nv-boot-control-service', '', d)}"
+RDEPENDS_${PN}_tegra210 = "${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'setup-nv-boot-control-service', '', d)} \
```
madisongh: See above.  The right way to fix this is to provide an initscript for the setup-nv-boot-control service, so it will work on both systemd and sysvinit builds.

contrib/recipes-devtools/gcc/gcc-7.3.inc
focusware-develop: Using the official 7.4.0 upstream patch would be better indeed.

recipes-bsp/tegra-binaries/tegra-redundant-boot_32.3.1.bb
focusware-develop: This is indeed not necessary. I checked it without the change and it build fine for sysvinit.

recipes-bsp/tegra-binaries/tegra-redundant-boot_32.3.1.bb
focusware-develop: This is indeed not necessary. I've checked building without the change, and it builds fine for sysvinit.

recipes-bsp/tegra-binaries/tegra-redundant-boot_32.3.1.bb
focusware-develop: This change is needed, otherwise it will create a dependency on "mender-tegra-bup-payload-install" that is not available on a sysvinit system.

madisongh:
I've manually merged in the changes to `tegra194.inc` and the flash helper script, back-ported the needed patch for gcc 7.3, and have provided initscripts that should solve problem you were having with `/etc/nv_boot_control.conf`.  I ran some tests building without systemd, both with and without mender, and they look OK to me.  Update to latest and give them a try.

focusware-develop:
Thanks again for the updates I'll try them this weekend!

focusware-develop:
I've check the latest version and it works fine (after fixing a build problem for a sysvinit only build). I've created a pull request for the fix.

PR #274, opened 2020-03-31
setup-nv-boot-control: fixed the sysvinit build

focusware-develop:
When building an image with sysvinit, the final build step results in an error: "mender-tegra-bup-payload-install missing". This fix makes sure for a sysvinit build that the "setup-nv-boot-control-service" package also contains a file. Otherwise this package will not be created resulting in a failure since "mender-tegra-bup-payload-install" is in the end dependent on this package.

Signed-off-by: Boris Nagels <develop@focusware.nl>


madisongh:
Thanks!

PR #275, opened 2020-04-01
Added Nvidia Container API to warrior-l4t-r32.2 branch

huberph:
I added the container API from the master branch to the warrior-l4t-r32.2 branch. I could successfully run the CUDA samples and tensorflow in a docker container. I added also the changes for gstreamer and v4l but I didn't test it because I don't need this in our project.


madisongh:
Thanks!  I ran a quick test with gstreamer using the v4l2 plugins, and they worked as well.

PR #279, opened 2020-04-01
libgcc-initial: Code cleanup

leon-anavi:
Code cleanup to remove dummy source which is not needed.


madisongh:
Thanks!

PR #283, opened 2020-04-06
tegra-flashvars: enable customization of pinmux and pmc configs

kekiefer:
For custom boards, L4T documentation walks through the process of
generating pinmux and pmc configurations. These need to be used when
the board is flashed.

This change moves the pinmux and pmc config files into the flashvars
file so that this file can be easily overridden by other layers.


madisongh:
Thanks!

PR #307, opened 2020-04-27
Change module loading to prepend instead of set.

dljsjr:
Changes kernel module load in the include files to prepend instead of set. This allows for appending to these variables in the `local.conf` without needing to create a new layer.


madisongh:
How about just using `<variable>_append` in your `local.conf` file to make additions?

dljsjr:
I tried that and it didn't seem to work, I may have had a typo though.

madisongh:
Yeah, give it another try.  It should work.  Don't forget to include a leading space in the string.

dljsjr:
Yeah it's working, I accidentally typed `KERNEL_MODULES` not `MODULE` 😅

PR #315, opened 2020-05-12
add support for rootfstype kernel param when mounting rootfs from initrd

shlomow:
When we mount the root device we used ext4 hardcoded. In this commit I changed the root file system type to inherit from the `rootfstype` kernel argument. If this fails we try to mount using the `auto` type.


madisongh:
Thanks!

PR #317, opened 2020-05-20
linux-tegra: Use features from linux-yocto.inc

wlindauer:
Including linux-yocto.inc simplifies the recipe file. This also allows
the kernel recipe (and bbappends) to take advantage of KERNEL_FEATURES.
Using KERNEL_FEATURES allows distro layers to modify the kernel config
by using kernel config fragments and .scc files. Most importantly, this
commit keeps from overwriting the .config file in ${B} which makes it
difficult to modify the kernel config from a distro layer.

I am creating a pull request for the thud-l4t-r32.3.1 branch since that is the branch I have tested it on, but I am hoping this can be ported forward to other branches like master.

### DIFF
--- a/recipes-kernel/linux/linux-tegra_4.9.bb
+++ b/recipes-kernel/linux/linux-tegra_4.9.bb
@@ -4,33 +4,22 @@ DESCRIPTION = "Linux kernel from sources provided by Nvidia for Tegra processors
 LICENSE = "GPLv2"
 LIC_FILES_CHKSUM = "file://COPYING;md5=d7810fab7487fb0aad327b76f1be7cd7"
 
-inherit kernel
+require recipes-kernel/linux/linux-yocto.inc
 
-PV .= "+git${SRCPV}"
+LINUX_VERSION ?= "4.9.140"
+PV = "${LINUX_VERSION}+git${SRCPV}"
 FILESEXTRAPATHS_prepend := "${THISDIR}/${BPN}-${@bb.parse.BBHandler.vars_from_file(d.getVar('FILE', False),d)[1]}:"
 EXTRA_OEMAKE += 'LIBGCC=""'
 
-L4T_VERSION = "l4t-r32.3.1"
-SCMVERSION ??= "y"
-export LOCALVERSION = ""
+LINUX_VERSION_EXTENSION ?= "-l4t-r32.3.1"
 
-SRCBRANCH = "patches-${L4T_VERSION}"
+KBRANCH = "patches${LINUX_VERSION_EXTENSION}"
 SRCREV = "47e7e1cb0b492487faa6258a4f3efe91676568b7"
 KERNEL_REPO = "github.com/madisongh/linux-tegra-4.9"
-SRC_URI = "git://${KERNEL_REPO};branch=${SRCBRANCH} \
+SRC_URI = "git://${KERNEL_REPO};branch=${KBRANCH} \
 	   file://defconfig \
 "
 S = "${WORKDIR}/git"
-
-do_configure_prepend() {
-    localversion="-${L4T_VERSION}"
-    if [ "${SCMVERSION}" = "y" ]; then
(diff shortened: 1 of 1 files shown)


madisongh:
This has been proposed before.  I'm not crazy about pulling in all the linux-yocto baggage, mainly because it assumes that the kernel sources are managed the same way the Yocto kernel team manages linux-yocto.

Is it just the kernel config that you're looking to modify, or did you want the whole feature definition setup with patches plus associated config fragments?

wlindauer:
Sorry about the duplicate PR, I should have looked before submitting another one. The silver lining is that this version cleans up some lines instead of just adding more code :-) I have to admit, I am a bit perplexed by the resistance to this. I don't understand the "it assumes that the kernel sources are managed the same way" because this PR seems to work fine without needing to change anything about how you manage your kernel repo. Can you elaborate more on what your concerns are? Are you already managing your kernel the same as linux-yocto and just don't want to be locked in? I thought this would actually be less duplicating of code since you already were doing a similar modification with CONFIG_LOCALVERSION and using a version extension.

To answer your question, I am looking to use the whole feature setup with .scc files because that is how our company distro layer currently works with our company BSP layer (which has used linux-yocto.inc for years). This allows applying the same feature changes to linux-yocto kernels, company BSP kernels, and your kernels all without needing to duplicate and maintain a defconfig for each one. The alternative to duplicating your defconfig is for every client of your layer to run merge_config.sh in a bbappend after the .config has already been overwritten and that seems sub-optimal in my opinion. Using merge_config.sh manually also limits you to just config fragments without the ability to group patches in the .scc files. I hope this answered your question adequately.

madisongh:
> I am a bit perplexed by the resistance to this.  I don't understand the "it assumes that the kernel sources are managed the same way" 

TBH, part of my resistance probably stems from my attempt to customize a linux-yocto kernel for a past project, and getting frustrated with the complexity involved in making the simplest of changes, particularly when it didn't work as I expected. Definitely left a bad taste for me.  Of course, that was some years ago now, and it's likely that the tooling works better now than it did then.

By "the same way" I meant the way the linux-yocto team manages its kernel, with the extra metadata in a separate repo, the branching to represent platform variants, etc., plus the kernel code base that's much closer to upstream than we'll ever get to. I guess I'm afraid of getting stuck at some point in the future having to debug build failures, or work around broken configuration processes, to accommodate a years-old, Androidized and vendor-hacked kernel that is outside the realm of possibilities considered by the linux-yocto developers as they evolve their tooling.

That said, I don't want to prevent folks from using linux-yocto-style customization if it works better for them.  I'm just not sure that I'm keen on making it the default.  If tweaking some variables and replacing the `do_configure_prepend()` with something that is easier to override makes it easier to use `linux-yocto.inc` from a bbappend in your layer, that would be more than OK with me.

madisongh:
OK, so I spent some more time with this and changed my mind.  After some test builds, I've merged a slightly modified version of this into master, and backproted it to thud-l4t-r32.3.1.

The main difference is adding back the SCMVERSION setting.

Hopefully the modified version will work for you.  If it does, I can port the changes to the other current branches.

Thanks!

wlindauer:
I appreciate you considering this more. Your modified commit works just fine for my use case and the SCMVERSION improvement makes it even better. This change will make it much easier for distro layers to modify the kernel config, especially in my case, where I need to work with multiple BSP layers at the same time.
I imagine that this change will work fine on already existing branches of poky releases. The compromise can be that if a future poky commit to linux-yocto.inc tooling breaks something, you can always just replace the "require linux-yocto.inc" line with "inherit kernel" again on your master branch. The other tweaks will likely keep working in both cases, just without applying the cfg fragment.
I am going to leave this PR open, assuming you will close it when you are done back-porting to the appropriate branches. Again, thank you for your speedy response.

madisongh:
Closing this now, since all branches based off L4T 32.3.1 and later now have the updated kernel recipe.

PR #324, opened 2020-05-26
cboot: update sha256 hash for 32.4.2

kekiefer:
I don't know if NV did something sneaky and changed the archive on us, or if you pulled a corrupted archive when you did the work, but a fresh fetch of cboot sources this morning needed a new hash. I confirmed on a different machine with a second fetch of these sources that this is indeed correct.


madisongh:
Looks like they posted a new copy.  The sources are identical, but they've added a license file at the top of the tree.

There is another issue with the recipe, though, in that it's not renaming the fetched tarball to include the version number in the name of the file.  That could lead to some conflicts in shared downloads directories if you start building cboot from different L4T versions. I should fix that.

PR #326, opened 2020-05-27
enable at24 module

shlomow:
enable at24 module so eeprom tool would work


madisongh:
I would prefer not to make this change.  To keep the kernel build as close to the stock L4T kernel, the `defconfig` file is a direct copy extracted from the kernel headers included in the L4T package, and I'd rather keep it that way.

You should be able to make this change in your own builds by adding a `bbappend` for the kernel recipe with the `CONFIG_EEPROM_AT24=m` (or `=y`) setting in a `.cfg` file that you add to the SRC_URI.  See (here)[https://github.com/madisongh/test-distro/blob/master/layers/meta-testdistro/recipes-kernel/linux/linux-tegra_4.9.bbappend] for an example from my test distro.

shlomow:
okay thanks!

PR #333, opened 2020-06-03
cuda.bbclass, cuda-environment: add some environment variables

madisongh:
for better compatibility with the old FindCUDA module.


kekiefer:
Now with this change, the good news is that opencv build is actually finding cuda properly and trying to build using the version in the sysroot. This brings two pieces of bad news:

First, it's trying to pull in the download for `NVIDIA_OPTICAL_FLOW`. The opencv recipe patches out online downloads, so the archive it's looking for just needs to be cached during fetch. I did this manually and it works, so I should be able to fix up the recipe to do it automatically. Not so bad.

The second piece of bad news is that in 10.2 (actually, 10.1), cublas isn't in the cuda tree anymore, so even the current opencv can't find it. I can get around it by adding `-DWITH_CUBLAS=OFF` to the EXTRA_OECMAKE. The other possibility is either a patch to the cmake recipes or a symlink of cublas back into the cuda tree (https://forums.developer.nvidia.com/t/cuda-blas-libraries-not-installed/107908/17).

I'm happy to take these fixes on since I need to get opencv working again for my image, but if you have a preference for how to fix the cublas issue, let me know.

madisongh:
@kekiefer see #337 for my take on this.

PR #335, opened 2020-06-03
tegra-common: add machine extra rdepends for egl-wayland when required

kekiefer:
Currently, egl-wayland is included at runtime only when the weston
compositor is used, or when egl-wayland is explicitly included in
the image. But really, the compositor doesn't need to use it -- the
compositor renders with egldevice platform or some other mechanism,
and it's the clients who need the egl-wayland platform if they want
to create a wayland egl surface. To be fair, weston does actually use
the library, but only for an optional interface.

It's not valid to assume that weston will always be the compositor,
and regardless of the compositor, clients may still need this backend.

This change puts egl-wayland in the MACHINE_EXTRA_RDEPENDS inside of
tegra-common.inc so all tegra machines will have egl-wayland if the
distro uses wayland, regardless of the compositor being used.


madisongh:
It looks like egl-wayland is called on by `libEGL_nvidia.so.0`.  How about adding the conditional dependency to RDEPENDS_${PN} in the tegra-libraries recipe instead?

kekiefer:
That makes a lot more sense. I've actually made it just an RRECOMMENDS, since it's not a strong dependency.

PR #342, opened 2020-06-08
machine: Add a default value for IMAGE_FSTYPES

wlindauer:
Yocto documentation recommends setting this variable in your machine
configuration file. This has the added benefit of making it simpler to
start using this layer and is needed when building the same image using
two different BSP layers with custom IMAGE_CLASSES.


madisongh:
I think there was a reason why I didn't do this initially, but I can no longer remember what that was... regardless, it's long past time we did this, so thanks.  I've got a modified version of this that extends this change to all the tegra platforms that I'm testing in master.  If that checks out OK, I'll back-port it here and to the other current branches.

madisongh:
I've pushed the more general fix.  Thanks again!

PR #346, opened 2020-06-11
linux-tegra: add patch for devtool compatibility

mozcelikors:
When "devtool modify linux-tegra" is invoked, get_staging_kver function in poky/scripts/lib/devtool/standard.py results in a parse error. Devtool parses first few lines for the kernel version with its current implementation. Therefore to make linux-tegra Makefile compatible with devtool, first few lines should contain kernel version information.
This commit introduces a patch that makes linux-tegra usable with general Yocto devtool workflow.
Patch could be cherry-picked to other branches that suffer the same problem.

Signed-off-by: Mustafa Özçelikörs <mozcelikors@gmail.com>


madisongh:
Yeah, I ran into this myself the other day.  I'd rather fix this directly in [linux-tegra](https://github.com/madisongh/linux-tegra-4.9), and will gladly take a PR there for this.

mozcelikors:
Sure, I will do that soon, Thanks.

madisongh:
Thanks for the linux-tegra PR.  I've merged it and have pushed 12c8ba45967771fd2e1b77da4a96f386c03aa942 here to bump the SRCREV to point to your change.

PR #348, opened 2020-06-12
remove kernel-devicetree from MACHINE_ESSENTIAL_EXTRA_RDEPENDS

shlomow:
In the jetson, cboot is used to load device tree from its dedicated partition.

Thus, it is not needed to install it into rootfs

Signed-off-by: Shlomi Vaknin <shlomi.39sd@gmail.com>

I guess it is needed in other platforms as well, but I don't familliar with them.
By the way, we don't really need the whole u-boot package. What I did in another commit is to 
create a bbappend to u-boot and create a package that includes only the exlinux and initrd. What is your opinion?

### DIFF
--- a/conf/machine/include/tegra186.inc
+++ b/conf/machine/include/tegra186.inc
@@ -9,7 +9,7 @@ KERNEL_IMAGETYPE = "Image"
 KERNEL_ARGS ?= "console=ttyS0,115200 console=tty0 fbcon=map:0 isolcpus=1-2"
 
 MACHINE_FEATURES = "alsa usbhost pci rtc cuda"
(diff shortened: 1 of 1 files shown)


madisongh:
Thanks.  Yes, the same applies to the other platforms now, too, since they all use cboot.  That wasn't always the case... it's been a long time now, but IIRC the TX1 and TX2 both originally let U-Boot load the kernel device tree.

The u-boot repackaging sounds like a good idea, too.  There's no reason to include u-boot in the rootfs for these platforms.  

There was some other cleanup and refactoring needed in the machine config files, too, so I've gone ahead and done a complete sweep through them in #349.  Thanks!

PR #352, opened 2020-06-15
Add libvisionworks 1.6 in FILES

victor-uavia:
The package VisionWorksTargets.cmake set the location to the
libvisionworks library to '/usr/lib/libvisionworks.so.1.6.0' which is
not present in the recipe.

```
/.../recipe-sysroot/usr/share/visionworks/cmake/VisionWorksTargets.cmake:64 (message):
|   The imported target "visionworks" references the file
|
|      "/.../recipe-sysroot/usr/lib/libvisionworks.so.1.6.0"
```

This add the two following files from the package 'libvisionworks_1.6.0.501_arm64.deb'
- libvisionworks.so.1.6 -> libvisionworks.so.1.6.0
- libvisionworks.so.1.6.0

As a matter of fact, the 'libvisionworks.so' redirect to 'libvisionworks.so.1.6', which is not present.

### DIFF
--- a/recipes-devtools/visionworks/libvisionworks_1.6.0.501.bb
+++ b/recipes-devtools/visionworks/libvisionworks_1.6.0.501.bb
@@ -35,7 +35,6 @@ INHIBIT_PACKAGE_STRIP = "1"
 INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
 INHIBIT_SYSROOT_STRIP = "1"
 
-FILES_${PN} = "${libdir}/libvisionworks.so*"
-FILES_${PN}-dev = "${includedir} ${libdir}/pkgconfig ${datadir}/visionworks"
+FILES_${PN}-dev = "${datadir}/visionworks"
 RDEPENDS_${PN} = "libstdc++"
 PACKAGE_ARCH = "${SOC_FAMILY_PKGARCH}"
--- a/recipes-devtools/visionworks/libvisionworks_1.6.0.501.bb
+++ b/recipes-devtools/visionworks/libvisionworks_1.6.0.501.bb
@@ -27,7 +27,7 @@ do_install() {
     install -d ${D}${prefix} ${D}${libdir} ${D}${datadir} ${D}${libdir}/pkgconfig
     cp -R --preserve=mode,timestamps ${B}/usr/include ${D}${prefix}/
     cp -R --preserve=mode,timestamps ${B}/usr/lib/pkgconfig/* ${D}${libdir}/pkgconfig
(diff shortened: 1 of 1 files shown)


recipes-devtools/visionworks/libvisionworks_1.6.0.501.bb
```diff
@@ -27,15 +27,15 @@ do_install() {
     install -d ${D}${prefix} ${D}${libdir} ${D}${datadir} ${D}${libdir}/pkgconfig
     cp -R --preserve=mode,timestamps ${B}/usr/include ${D}${prefix}/
     cp -R --preserve=mode,timestamps ${B}/usr/lib/pkgconfig/* ${D}${libdir}/pkgconfig
-    cp --preserve=mode,timestamps ${B}/usr/lib/libvisionworks.so ${D}${libdir}/
+    cp --preserve=mode,timestamps ${B}/usr/lib/libvisionworks.so* ${D}${libdir}/
```
madisongh, line 30: This should also add `,links` to the `--preserve` and add `--no-dereference`, so the symlinks are preserved as-is.

recipes-devtools/visionworks/libvisionworks_1.6.0.501.bb
```diff
@@ -27,15 +27,15 @@ do_install() {
...
     cp -R --preserve=mode,timestamps ${B}/usr/share/visionworks ${D}${datadir}/
 }
 
 INHIBIT_PACKAGE_STRIP = "1"
 INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
 INHIBIT_SYSROOT_STRIP = "1"
 
-FILES_${PN} = "${libdir}/libvisionworks.so"
+FILES_${PN} = "${libdir}/libvisionworks.so*"
```
madisongh, line 38: We should be able to just delete this line now, since the SONAME-based symlinks are all present.  Normal packaging rules will handle the libraries correctly.

recipes-devtools/visionworks/libvisionworks_1.6.0.501.bb
```diff
@@ -27,15 +27,15 @@ do_install() {
...
 FILES_${PN}-dev = "${includedir} ${libdir}/pkgconfig ${datadir}/visionworks"
```
madisongh, line 39: This line should be changed to
```
FILES_${PN}-dev += "${datadir}/visionworks"
```
the other directories mentioned should already get included by normal packaging, as will the ${SOLIBSDEV} symlinks in ${libdir} for the libraries.

victor-uavia:
@madisongh thanks for the insight, I update the code.

PR #361, opened 2020-06-17
Update dunfell

shlomow:
This pull request only cherry picks the latest commits that applied on the machine section


madisongh:
Thanks.  I held off on this until I had done some testing with these changes in master.  Everything checked out there, so this is good to go in.

PR #364, opened 2020-06-20
tegra-bup-payload: move recipe from meta-testdistro

kekiefer:
This recipe provides a way to bundle a BUP payload into the rootfs,
to be used during initial install.

It is being ported from meta-testdistro, as it is useful for various
kinds of OTA updates.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-bup-payload_1.0.bb
+++ b/recipes-bsp/tegra-binaries/tegra-bup-payload_1.0.bb
@@ -1,0 +1,24 @@
+DESCRIPTION = "Install tegra bup payload file in the location expected by nv_update_engine"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+COMPATIBLE_MACHINE = "(tegra)"
+
+UBOOT_SUFFIX ??= "bin"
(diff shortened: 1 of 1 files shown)


madisongh:
No objection to this, in principle.  We do need to handle the case where the initramfs is not bundled into the kernel, which used to be the default for cboot-based platforms and is still an option.

kekiefer:
It's not a problem to fix up this recipe to handle missing the bup payload in the image deploy directory, it just won't have anything to deploy. I get here by setting `INITRAMFS_IMAGE_BUNDLE = ""` (tried this on a jetson-xavier), but then `do_image_tegraflash` will fail since it doesn't handle this case. So it seems like I might be missing the point here. What else should I be setting to get this case working?

kekiefer:
I've updated this PR. I'm not sure it makes a whole lot of sense, since it doesn't actually deploy anything in the case where `INITRAMFS_IMAGE_BUNDLE = ""`, since `kernel-bup-payload:do_deploy` doesn't generate one in this case.

madisongh:
@kekiefer I manually merged this after adding in the logic for handling the separate initrd/kernel case.  In testing it I found a couple of issues elsewhere, so I've pushed fixes for those, too.

So at this point, all cases should now be handled correctly:
  * u-boot -> depend on u-boot-bup-payload:do_deploy
  * cboot with kernel-bundled initramfs -> depend on kernel-bup-payload:do_deploy
  * cboot with separate initrd + kernel -> depend on initramfs's do_image_complete

I also took the liberty of moving the recipe to a different subdirectory, just to keep it separate from the recipes that are more about unpacking binaries from the L4T BSP package.

PR #391, opened 2020-07-15
Backport meta: use bundled initramfs instead of initrd on tegra186/194

danielwangksu:
This is a backport from master branch, tested those changes won't affect TX2/TX2i

Original Commit: af4492634b6e3c0657973cf20418c95a7d8fe0a4
Original comment:

meta: use bundled initramfs instead of initrd on tegra186/194

cboot appears to mangle the initrd at boot time when it
exceeds a couple of MiB in size, but only if it has
been built separately and combined with the kernel using
mkbootimg.  This does not happen when the initrd is
built into the kernel directly, so work around the cboot
issue by switching to bundling the initramfs into the
kernel on the non-u-boot platforms.

Signed-off-by: Daniel Wang <daniel.wangksu@gmail.com>

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -11,7 +11,7 @@ TEGRA_SIGNING_ARGS ??= ""
 TEGRA_SIGNING_ENV ??= ""
 
 DTBFILE ?= "${@os.path.basename(d.getVar('KERNEL_DEVICETREE').split()[0])}"
-LNXFILE ?= "${@'${IMAGE_UBOOT}-${MACHINE}.bin' if '${IMAGE_UBOOT}' != '' else '${INITRD_IMAGE}-${MACHINE}.cboot'}"
+LNXFILE ?= "${@'${IMAGE_UBOOT}-${MACHINE}.bin' if '${IMAGE_UBOOT}' != '' else '${KERNEL_IMAGETYPE}-initramfs-${MACHINE}.cboot'}"
 LNXSIZE ?= "67108864"
 
 IMAGE_TEGRAFLASH_FS_TYPE ??= "ext4"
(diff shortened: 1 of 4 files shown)


classes/image_types_tegra.bbclass
```diff
@@ -540,7 +540,7 @@ do_image_tegraflash[depends] += "zip-native:do_populate_sysroot dtc-native:do_po
                                  virtual/kernel:do_deploy ${TEGRAFLASHDEPS} \
                                  ${@'${INITRD_IMAGE}:do_image_complete' if d.getVar('INITRD_IMAGE') != '' else  ''} \
                                  ${@'${IMAGE_UBOOT}:do_deploy ${IMAGE_UBOOT}:do_populate_lic' if d.getVar('IMAGE_UBOOT') != '' else  ''} \
-                                 ${@'cboot:do_deploy' if d.getVar('IMAGE_UBOOT') != '' and d.getVar('SOC_FAMILY') == 'tegra186' else  ''}"
+                                 cboot:do_deploy"
```
madisongh, line 543: This change looks like it may break TX1 builds.  Changing the `and d.getVar...` conditional to be `!= 'tegra210'`, instead of removing it altogether, might be the right way to fix it.

danielwangksu: Good point! I changed it

madisongh:
@danielwangksu I have manually merged this in after some testing across the different targets and making some additional changes to fix some things.

* I've added back in the kernel-bup-payload recipe from the original commit to allow for BUP payload creation when using the bundled initramfs
* Some minor changes to ensure compatibility for building without bundling the initramfs
* I did leave tegra186 platforms with the old non-bundled settings.
* I made some minor changes to the tegra194 flash helper script since I ran into issues with it, but those issues appear unrelated to this change.

PR #402, opened 2020-07-30
tegra[186|194]-flash-helper: restore ability to specify flash command

kekiefer:
Allows you to flash a single partition, e.g.:

./doflash.sh -c "write CPUBL-CFG cbo.dtb; reboot"

Regression of changes made in #205


madisongh:
Oh, oops!  Good catch, thanks!

PR #407, opened 2020-08-19
cuda: add compatible hosts

sapiippo:
Use COMPATIBLE_HOST to limit the native and nativesdk builds to the
two supported architectures. Otherwise bitbake will throw the
"unrecognized HOST_ARCH" errors during parsing of the recipes,
even when the recipe is not selected to be built.


madisongh:
Thanks!

PR #417, opened 2020-09-02
added missing HOMEPAGE to recipes

lfdmn:
Please check.

I'm not sure for :
  - u-boot-tegra-common-2016.07.inc
  - linux-tegra_4.9.bb


madisongh:
These look OK.  Interestingly, the kernel recipes in OE-Core don't set HOMEPAGE, either.
For the u-boot recipe, we could use the DENX home page set in the OE-Core recipes, the L4T home page, or what you have here... not sure it really matters that much.

PR #425, opened 2020-09-17
Deepstream Support

mmontero:
Add deepstream recipes support and deepstream docker fix

- Backport GStreamer 1.14 from warrior for deepstream docker support
- Add recipe for deepstream-5.0 (core libraries, GStreamer plugins, samples, python binding)
- Add recipe for deepstream python apps
- Add python 3.6 recipe from python 3.7 recipe as base, since deepstream python binding requires python3.6

### DIFF
--- a/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
+++ b/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
@@ -12,7 +12,7 @@ SRC_COMMON_DEBS = "${BPN}_${PV}_arm64.deb;subdir=${BPN}"
 SRC_URI[sha256sum] = "a7a7015515883ac88c7587c7a2acfcf78510e539b84b702afd05f4f330faa55e"
 
 COMPATIBLE_MACHINE = "(tegra)"
-PACKAGE_ARCH = "${SOC_FAMILY_PKGARCH}"
+PACKAGE_ARCH = "${TEGRA_PKGARCH}"
 
 DEPENDS = "gstreamer1.0 gstreamer1.0-rtsp-server tensorrt cudnn libcublas cuda-cudart tegra-libraries"
 
--- a/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
+++ b/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
@@ -36,8 +36,8 @@ do_install() {
     install -d ${D}${DEEPSTREAM_PATH}/lib/tensorflow
     install -m 0644 ${S}${DEEPSTREAM_PATH}/lib/tensorflow/* ${D}${DEEPSTREAM_PATH}/lib/tensorflow
 
-    install -d ${D}/etc/ld.so.conf.d/
-    echo "${DEEPSTREAM_PATH}/lib" > ${D}/etc/ld.so.conf.d/deepstream.conf
+    install -d ${D}/${sysconfdir}/ld.so.conf.d/
+    echo "${DEEPSTREAM_PATH}/lib" > ${D}/${sysconfdir}/ld.so.conf.d/deepstream.conf
 
     install -d ${D}${libdir}/gstreamer-1.0/deepstream
     install -m 0644 ${S}${DEEPSTREAM_PATH}/lib/gst-plugins/* ${D}${libdir}/gstreamer-1.0/deepstream/
@@ -61,7 +61,7 @@ RDEPENDS_${PN}-sources = "bash"
 
 PACKAGES += "${PN}-samples ${PN}-sources"
 
-FILES_${PN} = "/etc/ld.so.conf.d/  \
+FILES_${PN} = "${sysconfdir}/ld.so.conf.d/  \
 	       ${libdir}/gstreamer-1.0/deepstream \
 	       ${DEEPSTREAM_PATH}/lib \
 	      "
--- a/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
+++ b/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
@@ -5,6 +5,7 @@ LIC_FILES_CHKSUM = " \
     file://opt/nvidia/deepstream/deepstream-5.0/LICENSE.txt;md5=c614b345088f989a9c05a7393419bf55 \
     file://opt/nvidia/deepstream/deepstream-5.0/doc/nvidia-tegra/LICENSE.iothub_client;md5=4f8c6347a759d246b5f96281726b8611 \
     file://opt/nvidia/deepstream/deepstream-5.0/doc/nvidia-tegra/LICENSE.nvds_amqp_protocol_adaptor;md5=8b4b651fa4090272b2e08e208140a658"
+HOMEPAGE = "https://developer.nvidia.com/deepstream-sdk"
 
 inherit l4t_deb_pkgfeed
 
--- a/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
+++ b/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
@@ -16,8 +16,6 @@ PACKAGE_ARCH = "${TEGRA_PKGARCH}"
 
 DEPENDS = "gstreamer1.0 gstreamer1.0-rtsp-server tensorrt cudnn libcublas cuda-cudart tegra-libraries"
 
-inherit pkgconfig
-
 S = "${WORKDIR}/${BPN}"
 B = "${WORKDIR}/build"
 
--- a/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
+++ b/recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
@@ -1,6 +1,10 @@
 DESCRIPTION = "NVIDIA Deepstream SDK"
 LICENSE = "Proprietary"
-LIC_FILES_CHKSUM = "file://usr/share/doc/deepstream-5.0/copyright;md5=f635f9f375e764ce281a2070599e2457"
+LIC_FILES_CHKSUM = " \
+    file://usr/share/doc/deepstream-5.0/copyright;md5=f635f9f375e764ce281a2070599e2457 \
+    file://opt/nvidia/deepstream/deepstream-5.0/LICENSE.txt;md5=c614b345088f989a9c05a7393419bf55 \
+    file://opt/nvidia/deepstream/deepstream-5.0/doc/nvidia-tegra/LICENSE.iothub_client;md5=4f8c6347a759d246b5f96281726b8611 \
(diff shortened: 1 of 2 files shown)


madisongh:
Thanks for the work on this. A couple of questions...

We already have gstreamer 1.14 recipes in meta-tegra/contrib, so we don't need those recipes repeated.  Can you try deepstream with the contrib recipes?

Is the older Python 3.6.x really a requirement?  Does deepstream not interoperate with newer versions? It would be good to see if we can drop this dependency, especially since as a BSP layer, we'd be breaking compatibility rules with the python3-dir and gpgme changes.

If down-revving is a hard requirement, that complicates things.  I'd be OK with adding, say, a `python36_3.6.9.bb` recipe (to the contrib sub-layer) that installs its interpreter as `${bindir}/python3.6`,  and then force the deepstream-related python scripts to use `python3.6` instead of `python3` in their shebang lines.  Then instead of replacing `python3-dir.bbclass`, we could have a `python36-dir.bbclass` that points variables to the 3.6 paths, and reference that where there is the version-specific dependency. As long as it can be installed alongside the normal python3 and not conflict, and we can keep the blast radius for that change restricted to just the deepstream-related dependents, that should work OK and not require retesting every other python package in all layers to see if they all still work.

mmadrigal:
Hi @madisongh , thanks for reviewing this, we are excited to contribute to your project. To answer some of your questions:

1. Yes, we are going to try with the contrib recipes. At the time we added GStreamer 1.14 to the fork the support was not in place yet. If it works with the contrib that would be great!

2. Regarding the Python 3.6.x support. We found that deepstream python bindings depend on Python 3.6.x version. We agree this is not ideal and we were kind of surprised about this as well. The problem is not related with the python scripts per se, but with the bindings file (pyds.so) which is a dependency of the python apps (https://github.com/NVIDIA-AI-IOT/deepstream_python_apps). 

Let us know your thoughts.

madisongh:
Yeah, it's a shame that NVIDIA didn't open-source the python bindings. That doesn't leave us with a whole lot of options.

One option would be to just omit the Python bindings library altogether, or at least package it separately from the main deepstream package (which may require some packaging workarounds to ignore the missing dependency on libpython3.6m.so).  That would at least make deepstream usable for C/C++ applications.  Folks that want to use the python bindings could go down the path of using the DS-5.0 NGC container, or could in their own distro/layer switch out to python 3.6 and include the python bindings package in their builds that way.

Another option would be to do what I mentioned earlier, creating a python 3.6 that could co-reside with normal python.  We could still package the deepstream python bindings separately, and then only if someone needed them would the extra python be included in their builds.  I think it's doable, but there could be some complications in this approach that I've missed.

A third option, which I expect would be quite a bit of work, would be to read through the NVIDIA documentation on the python bindings and write your own implementation of them that we could then build against whatever version of python is current.  From a brief glance at the DS docs, it looks like the bindings are more than just simple wrappers around the DS C API, though, so this could be a big hassle to maintain.

My preference, I guess, would be to start with the first option, just to get the base functionality in place, then figure out the python bindings issue later.  How does that sound?

mmadrigal:
It sounds like a good plan. It would help to absorb the support sooner and later focus on the python support. Unfortunately, since most of the ML and AI work is usually done in python this would become a must at some point. We are going to prepare the changes as discussed and be back to you soon. 

I believe the best way to go would be adding the additional instance of Python 3.6 along with the default version, so both can co-exist somehow or at least leave Python 3.6 as an optional selection. I agree with you than trying to re-do the bindings might be more time consuming and likely quite difficult to maintain on later versions of Jetpack. 

We are really interested in merging these changes to your project and improve the Jetson support. We have prepared this wiki for the community and we want it to point to the original project only :)

https://developer.ridgerun.com/wiki/index.php?title=Yocto_Support_for_NVIDIA_Jetson_Platforms

madisongh:
@mmadrigal Sounds good. Hopefully, NVIDIA will open-source the bindings in future versions of the SDK, so we won't have to have janky workarounds like this for too long.

Thanks!

mmadrigal:
Hi @madisongh , the team has done the changes and it is ready for you to review. We left the python bindings support on a separate branch in our fork, we will work on this later. 

Also, we tested with the contrib recipes for GStreamer 1.14 and everything seems to work as expected :) we are going to update our wiki to use the contrib.

recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
```diff
@@ -0,0 +1,73 @@
+DESCRIPTION = "NVIDIA Deepstream SDK"
+LICENSE = "Proprietary"
+LIC_FILES_CHKSUM = "file://usr/share/doc/deepstream-5.0/copyright;md5=f635f9f375e764ce281a2070599e2457"
```
madisongh, line 3: Not a showstopper for this, but (not surprisingly) there's a bunch of open-source code included in the package as well, and additional license texts:
```
/opt/nvidia/deepstream/deepstream-5.0/LICENSE.txt
/opt/nvidia/deepstream/deepstream-5.0/doc/nvidia-tegra/LICENSE.nvds_amqp_protocol_adaptor
/opt/nvidia/deepstream/deepstream-5.0/doc/nvidia-tegra/LICENSE.iothub_client
```
It would be good to enumerate them all and add the texts to the checksum.

recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
```diff
@@ -0,0 +1,73 @@
...
+
+SRC_URI="https://repo.download.nvidia.com/jetson/common/pool/main/d/deepstream-5.0/deepstream-5.0_5.0.0-1_arm64.deb;subdir=${BPN}"
```
madisongh, line 5: We have a bbclass that will compute the SRC_URI for the packages NVIDIA serves via its package feed, you can replace this line with
```
inherit l4t_deb_pkgfeed

SRC_COMMON_DEBS = "${BPN}_${PV}_arm64.deb;subdir=${BPN}"
```
and that way if the base URL of the package feed changes, we won't have to change all the recipes.

recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
```diff
@@ -0,0 +1,73 @@
...
+SRC_URI[sha256sum] = "a7a7015515883ac88c7587c7a2acfcf78510e539b84b702afd05f4f330faa55e"
+
+COMPATIBLE_MACHINE = "(tegra)"
+PACKAGE_ARCH = "${SOC_FAMILY_PKGARCH}"
```
madisongh, line 9: A recent addition to the branch is a TEGRA_PKGARCH package architecture for packages that are tegra-specific but not SoC-specific.  This is one of those, so replacing SOC_FAMILY_PKGARCH with TEGRA_PKGARCH would be helpful for folks that build for multiple platforms.

recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
```diff
@@ -0,0 +1,73 @@
...
+
+DEPENDS = "gstreamer1.0 gstreamer1.0-rtsp-server tensorrt cudnn libcublas cuda-cudart tegra-libraries"
+
+inherit pkgconfig
```
madisongh, line 13: The build for this recipe doesn't appear to use `pkg-config`, so this inherit isn't needed.

recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
```diff
@@ -0,0 +1,73 @@
+DESCRIPTION = "NVIDIA Deepstream SDK"
+LICENSE = "Proprietary"
```
madisongh, line 2: Could you add a home page setting?  Looks like
```
HOMEPAGE = "https://developer.nvidia.com/deepstream-sdk"
```
would be appropriate here.

madisongh (commented):
I've left a few comments in-line; if you could take care of those (although I'm willing to defer the license one for now), this should be good to go.

recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
```diff
@@ -0,0 +1,73 @@
...
+    for f in ${S}${DEEPSTREAM_PATH}/lib/*; do
+        [ ! -d "$f" ] || continue
+        install -m 0644 "$f" ${D}${DEEPSTREAM_PATH}/lib/
+    done
+    
+    install -d ${D}${DEEPSTREAM_PATH}/lib/tensorflow
+    install -m 0644 ${S}${DEEPSTREAM_PATH}/lib/tensorflow/* ${D}${DEEPSTREAM_PATH}/lib/tensorflow
+
+    install -d ${D}/etc/ld.so.conf.d/
+    echo "${DEEPSTREAM_PATH}/lib" > ${D}/etc/ld.so.conf.d/deepstream.conf
```
madisongh, line 35: Missed this earlier - could you replace the hard-coded `/etc` with `${sysconfdir}` in the two lines above?

recipes-devtools/deepstream/deepstream-5.0_5.0.0-1.bb
```diff
@@ -0,0 +1,73 @@
...
+    install -m 0755 ${S}${DEEPSTREAM_PATH}/bin/* ${D}${bindir}/
+
+    install -d ${D}${DEEPSTREAM_PATH}/lib/
+    for f in ${S}${DEEPSTREAM_PATH}/lib/*; do
+        [ ! -d "$f" ] || continue
+        install -m 0644 "$f" ${D}${DEEPSTREAM_PATH}/lib/
+    done
+    
+    install -d ${D}${DEEPSTREAM_PATH}/lib/tensorflow
+    install -m 0644 ${S}${DEEPSTREAM_PATH}/lib/tensorflow/* ${D}${DEEPSTREAM_PATH}/lib/tensorflow
```
madisongh, line 32: Unfortunately, this doesn't preserve the symlinks, so you end up with multiple copies of the (really large) `libtensorflow_trtis.so.1` library.  This needs to be changed to preserve the symlinks, or install just the library and re-create the symlinks under `${D}`.

madisongh:
Thanks again for putting this together! Some refinements to consider for the future:
* It would be nice to split the SDK up further, if possible, even if NVIDIA continues to release it as one giant blob.  That might take some work, and probably more understanding than I have about how the plugins and libraries are used.
* I see NVIDIA provides sources for at least some of the libraries and gstreamer plugins.  At some point, it would be great to build what we can from source instead of using their prebuilt binaries.  That might help with the gstreamer version problem, too, if the plugins can be compiled against whatever the current version of gstreamer is in the tree.  That's worked out well for the other plugins NVIDIA provides in the BSP (most of the time, at least).
* Likewise for the samples - build them from source instead of installing the prebuilts, maybe as a separate recipe.

madisongh (approved):
Thanks for making the requested changes.  Looks good to go.

PR #431, opened 2020-09-29
Support csv file reference removal globs

dwalkes:
The list of files in the csf for v4l-utils does not match the list
in /etc/nvidia-container-runtime/host-files-for-container.d/l4t.csv
on a Jetpack 4.4 based stock NVIDIA image.  This is problematic
for the libv4l2rds libraries when attempting to install v4l-utils
in a container.  The install works fine if the container is running
on an NVIDIA image, but when running on OE4T you see an error
message:
```
unable to make backup link of './usr/lib/aarch64-linux-gnu/libv4l2rds.so.0.0.0' before installing new version: Invalid cross-device link
```

To avoid this, add support for removing container runtime files
in the container-runtime-csv class using a glob pattern, and
use this pattern to avoid placing libv4l2rds in the container
runtime.

### DIFF
--- a/classes/container-runtime-csv.bbclass
+++ b/classes/container-runtime-csv.bbclass
@@ -1,4 +1,5 @@
 CONTAINER_CSV_FILES ??= "${libdir}/*.so*"
+CONTAINER_CSV_EXCLUDE_FILES ??= ""
 CONTAINER_CSV_BASENAME ??= "${PN}"
 CONTAINER_CSV_PKGNAME ?= "${CONTAINER_CSV_BASENAME}-container-csv"
 CONTAINER_CSV_EXTRA ??= ""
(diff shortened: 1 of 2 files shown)


dwalkes:
@madisongh a better way to do this might be to try to match the exact csv file content NVIDIA uses, however this is an easier approach, depending on how many one-off instances we find.  Let me know what you think.

madisongh:
It looks like Debian or Ubuntu puts `libv4l2rds.so.0.0.0` into a separate package, whereas the recipe in meta-oe doesn't do that, so that library isn't present in the stock L4T image (and hence not in their csv). Hopefully we won't have other cases like this, but adding the general exclusion mechanism makes sense to me.

PR #433, opened 2020-10-07
Fix Nvidia Board ref for Jetson xavier nx.

ichergui:


ichergui:
Hi @madisongh 
I was not sure about the destination branch, I chose the `dunfull-l4t-r32.4.3`
Let me know if I have to change it.
Have a nice day.

madisongh:
That variable is used to look up the directory under `Linux_for_Tegra/bootloader` under which the configuration files reside, which really is `t186ref` even for Xavier modules.  It might not be the best name for that variable, but I don't think you really want to be changing its value.

ichergui:
This mean that no need to update the variable defined in `xavier-nx.inc `
I will close the this PR, right ?

madisongh:
> This mean that no need to update the variable defined in xavier-nx.inc

Right.  If making this change is an attempt to fix some kind of failure, open an issue and let's discuss there.

ichergui:
OK.
Thanks.

PR #436, opened 2020-10-12
Fix CUDA include directories

eboskma:
The variable `CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES` is not correctly initialized in the generated `${WORKDIR}/toolchain.cmake`. 

The `CMAKE_CUDA_TOOLKIT_ROOT_DIR` and `CMAKE_CUDA_TOOLKIT_TARGET_DIR` variables are evaluated when the toolchain file is generated from `cuda.bbclass`, which are undefined at that time and the incude dirs become `/include /include`.

Also, no list separator was used, this results in an include path of `-I<root_dir> <target_dir>` instead of `-I<root_dir> -I<target_dir>`.

### DIFF
--- a/classes/cuda.bbclass
+++ b/classes/cuda.bbclass
@@ -59,8 +59,9 @@ cmake_do_generate_toolchain_file_append_cuda() {
     cat >> ${WORKDIR}/toolchain.cmake <<EOF
 set(CMAKE_CUDA_TOOLKIT_ROOT_DIR "${STAGING_DIR_NATIVE}/usr/local/cuda-${CUDA_VERSION}" CACHE PATH "" FORCE)
 set(CMAKE_CUDA_TOOLKIT_TARGET_DIR "${STAGING_DIR_HOST}/usr/local/cuda-${CUDA_VERSION}" CACHE PATH "" FORCE)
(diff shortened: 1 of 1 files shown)


madisongh:
Good catch, although I haven't yet run into a build failure because of this long-standing mistake - CMake still seems to be figuring out the correct include paths in spite of that variable being set incorrectly.  Maybe we should just drop the setting of that variable completely instead?

eboskma:
I just tried removing `CMAKE_CUDA_TOOLKIT_INCLUDE_DIRECTORIES` from `cuda.bbclass`, and that only sets the `recipe-sysroot-native` include dir. 

I'm testing it by building darknet, and then it fails to find `curand.h`, which is in the `recipe-sysroot`.

madisongh:
OK, got it.  The packages I've been building use `pkg-config` to get the CFLAGS (and thus the include paths) for curand and other libraries, that must be the difference.

madisongh:
I modified your changes a bit, fixed both the bbclass and the the toolchain file that gets installed in SDK builds, and have manually merged this fix (along with another CUDA build environment fix I needed to make) into all applicable branches.

Thanks, @eboskma !

PR #438, opened 2020-10-12
Update u-boot SRCREV to include allow equal sign as key value separation patch

arnstein:
Signed-off-by: Arnstein Kleven <arnsteinkleven@gmail.com>


madisongh:
Thanks for taking care of these!

PR #452, opened 2020-10-26
Update to L4T R32.4.4/JetPack 4.4.1

madisongh:
Lots of files touched with this, but mainly due to renaming.

Kernel recipe points to a new `oe4t-patches-l4t-r32.4` branch, as I couldn't get the merges right when trying to update the existing branch. I re-imported the NVIDIA repos as subtrees into the new branch using a slightly different approach, which I hope will work better for future merges, if they are necessary.  Our patches from the old `patches-l4t-r32.4` branch were rebased onto the new one with no conflicts.


madisongh:
Ran into a couple of minor issues with SDcard formatting during my final round of tests, so the tegra-helper-scripts commit needed some tweaking.

Everything looks good now - tested on TX1, TX2, AGX Xavier, Xavier NX (SD), Xavier NX (eMMC), Nano (SD), Nano (eMMC), both Mender and non-Mender.

PR #460, opened 2020-10-28
Tegra 210 Secure Boot Enhancements - Master Branch

timesys-nathan:
NVIDIA asked us to test secure boot out in conjunction with meta-tegra.  These are our notable findings, which they have asked us to push into meta-tegra.


madisongh:
Thanks for this.  On the pkc_disable problem - the release notes for L4T R32.4.4 state (on page 9) that that problem is supposed to be fixed in that release.  I was going to try it myself to see (I think I still have a Nano module around that didn't get messed up by my previous attempts with the broken nvtboot), but is the statement in the release notes incorrect?

timesys-nathan:
I was testing on the Dunfell branch, which looks like 32.3.1.  So, the problem is probably fixed on your master branch (and you can therefore disregard the fourth commit).

madisongh:
The flash helper script change looks good to me.  Still working through the other changes, although the one that packages up the entire contents of the Linux_for_Tegra tree looks dubious.

We probably need a wiki page on how to integrate fuse burning into a Yocto/OE-based build.  What I set up for TX2-based platforms is:

1. Use the regular L4T BSP and run `odmfuse.sh --no-burn` to generate the `odmfuse_pkc.xml` file which burns all required fuses.
2. Add the `odmfuse_pkc.xml` file via a bbappend to the `tegra-bootfiles` recipe, and set the ODMFUSE_FILE variable to  point to it.
3. When the tegraflash package is generated, if the XML file is present, a `burnfuses.sh` script gets generated for doing the initial fusing, and the package is set up so you can run that script, power cycle the device, then flash the signed binaries.

Extending this same workflow to t210 (and t194) platforms would be great. Making it possible to re-run `odmfuse.sh` on an already-fused device is kind of outside our scope, though.

nathan546:
So, I do agree the do_deploy_tarball() change is probably overkill.  There's some backstory for what it's worth:

We (Timesys) have a separate internal security layer which adds various features (encryption/verification/etc) to multiple platforms.  We were porting the Tegra 210 into it and these were the relevant patches I used that I _thought_ may be useful to you.  So, if you want to skip that one as well, I think that's fine.  The patches may just be a point of discussion or something for you to consider in the future.

From what we've done with other customers, we like to provide them with all of the necessary tools in the DEPLOY_DIR_IMAGE.  So, the intent was to provide odmfuse.sh in that same directory.  Bundling it into the tegraflash tarball alongside doflash.sh would probably be a better solution (in my opinion).  Ultimately, navigating to the tmp/work-shared/l4t directory isn't the end of the world either... just seemed a bit confusing to me.

Looking back through those patches, it looks like I accidentally ripped a space out of one of the commands in _0001-Add-in-u-argument-for-fusing-board-that-has-already.patch_.  I'm sure you'll notice that too.  It was not intentional.

As far as workflow goes for flashing the fuses, I took the opposite approach.  Historically, we've left the key file outside of Yocto completely.  So, when I ran through the T210, I passed the key file argument in externally through the whole process.  I do like that with your approach, you could store the key hash (XML file) inside a customer's Yocto layer without exposing the private key.

Hopefully we'll get some bandwidth to work on other items/boards (T194/TX2/...) in the future.  The T210s are all we've used so far.

Here was my entire workflow for secure boot:

# Secure Boot

Generate secure boot key:
```
SECURE_BOOT_KEY="/somewhere/secure"
openssl genrsa -out ${SECURE_BOOT_KEY}/rsa_priv.pem 2048
```

Extract odmfuse.sh (fuse burning tool)
```
cd <Yocto Directory>/build/tmp/deploy/images/timesys-jetson-nano-qspi-sd
mkdir efuses
tar -xf efuses.tar.gz -C efuses
cd efuses
```

We've found that some versions of the NVIDIA efuse tool accidentally burn the PKC disable bit (disabling secure boot permanently).  You may want to check that the SHA256 sum of these 4 files matches what we have included in our bbappend:
```
sha256sum bootloader/t210ref/nvtboot.bin bootloader/nvtboot_recovery.bin bootloader/nvtboot_cpu.bin bootloader/nvtboot_recovery_cpu.bin 

886ab5cd85a0efd96c28364f94532a9eb3880e9760398cebb8ea1fa7acfdf902  bootloader/t210ref/nvtboot.bin
e97ab424305cb0e9852c3aa3871eb79bf9bd50dd58a75b2236db5bbcbb1c7851  bootloader/nvtboot_recovery.bin
3810bb36c3b1df9ccf06c6a0e10e660c074062562d8941d322efc9fa0af9fe74  bootloader/nvtboot_cpu.bin
a370833064ce14fab2948c79cc595bd30533a7f8eeffdd26adff73f1fd6256ca  bootloader/nvtboot_recovery_cpu.bin
```

Place board in to recovery mode by shorting the two related pins and powering on

Burn the fuses (this will also disable JTAG).  You may also want to set the device key parameter via "-D <DK_file>" during this step.
```
SECURE_BOOT_KEY="/somewhere/secure"
sudo ./odmfuse.sh -i 0x21 -p -c PKC -k ${SECURE_BOOT_KEY}/rsa_priv.pem
```

Now, if you attempt to boot again at this stage, the board should refuse to boot, as the binaries are not signed.

Power off the board, then place board in to recovery mode by shorting the two related pins and powering on again.

Proceed to flash (securely)
```
sudo ./doflash.sh -u ${SECURE_BOOT_KEY}/rsa_priv.pem
```

madisongh:
Thanks for the background.  I have pulled in the fix for the helper script and have, at least for now, added support for using an `odmfuse_pkc.xml` file generated using the L4T BSP's `odmfuse.sh --no-burn` process to match the existing support for TX2 platforms.  I'd be willing to revisit this later to cover initial fusing if we can get the package trimmed down.

I've tested signing and flashing with a fused Nano and a fused Xavier NX module, so we now have coverage for all three SoC types.  I've also updated [our wiki page](https://github.com/OE4T/meta-tegra/wiki/Secure-Boot-Support) on secure boot support to cover the latest updates.

Issue #486, opened 2020-11-07
dunfell-l4t-r32.4.3: the gstreamer nvvideo4linux2 plugin are failing in some scenarios


quaresmajose:
I have found this with the test of the PR #471
I test this without the PR #471 with gstreamer **1.16.3** and **1.16.2** using OE core  `dunfell`  and  meta-tegra `dunfell-l4t-r32.4.3` branches

This pipeline don't work:
```
gst-launch-1.0 videotestsrc ! nvvidconv ! nvv4l2h264enc ! h264parse ! nvv4l2decoder ! fakesink dump=1
```

This one work:
```
gst-launch-1.0 videotestsrc ! nvvidconv ! nvv4l2h264enc ! nvv4l2decoder ! fakesink dump=1
```

I will try to find more information about this issue and come back.
but apparently the `nvvideo4linux2` is broken on `dunfell-l4t-r32.4.3`

madisongh:
Yes, I see a hang with the first pipeline, and no hang with the second.  If I remove `nvv4l2decoder` from the first pipeline, I get no hang, so it may be that element (or one of the binary-blob libraries) that's the problem.  I tried a couple of the patches from the NVIDIA dev forum for gstreamer issues (see [this page](https://elinux.org/Jetson/L4T/r32.4.x_patches) for links), specifically the replacements for `libnvtvmr.so`, since they appeared to be related to encode/decode issues, but they didn't help.

quaresmajose:
I already discovered the problem, there's a dead lock on `nvv4l2decoder` and I am trying to address it.

I think this problem affect the Ubuntu as well and the gstreamer 1.14 on the contrib layer.
Can anyone test the pipelines on the issue description on Ubuntu to check the results?

The dead lock is because the `gstv4l2bufferpool` only set  `capture_plane_stopped` if the object that use the poll are `V4L2_TYPE_IS_OUTPUT` and in this pipeline it have an encoder in front of the decoder. Because of this the `gstv4l2bufferpool` don't set the `capture_plane_stopped` and `nvv4l2decoder`  will waiting forever for this.

This only happens because the `h264parse` change the caps when the pipeline goes to running that will trigger a restart of the `nvv4l2decoder` in the `gst_v4l2_video_dec_set_format`

`gstv4l2bufferpool.c` code that don't set `capture_plane_stopped`:
```c
#ifdef USE_V4L2_TARGET_NV
  /* Make sure, there are no outstanding buffers to be unreffed by
[...]

PR #488, opened 2020-11-08
gstreamer1.0-plugins-nvvideo4linux2: check if we have a pool before t…

quaresmajose:
gstreamer1.0-plugins-nvvideo4linux2: check if we have a pool before the locking in video decoder set format

There's a dead lock on nvv4l2decoder.
The dead lock is because the gstv4l2bufferpool only set capture_plane_stopped if the object
that use the poll are V4L2_TYPE_IS_OUTPUT and in this pipeline we have an encoder and h264parse
in front of the decoder.
This disable the pool on the nvv4l2decoder and because of this the gstv4l2bufferpool don't
set the capture_plane_stopped and nvv4l2decoder will waiting forever for this.

This only happens because the h264parse change the caps when the pipeline goes to running
that will trigger a restart of the nvv4l2decoder in the gst_v4l2_video_dec_set_format.

So disable the the check for capture_plane_stopped on nvv4l2decoder if we don't have a pool
otherwise nvv4l2decoder will hang forever waiting for capture_plane_stopped.

https://github.com/OE4T/meta-tegra/issues/486

Signed-off-by: Jose Quaresma <quaresma.jose@gmail.com>

### DIFF
--- a/recipes-multimedia/gstreamer/gstreamer1.0-plugins-nvvideo4linux2_1.14.0-r32.4.3.bb
+++ b/recipes-multimedia/gstreamer/gstreamer1.0-plugins-nvvideo4linux2_1.14.0-r32.4.3.bb
@@ -15,6 +15,9 @@ SRC_URI += "file://build-fixups.patch"
 SRC_URI += "file://0001-v4l2videoenc-Fix-negotiation-caps-leak.patch"
 SRC_URI += "file://0002-v4l2allocator-Fix-data-offset-bytesused-size-validat.patch"
 SRC_URI += "file://0003-v4l2bufferpool-Avoid-set_flushing-warning.patch"
+# https://github.com/OE4T/meta-tegra/issues/486
+SRC_URI += "file://0001-gstv4l2videodec-use-ifdef-macro-for-consistency-with.patch"
+SRC_URI += "file://0002-gstv4l2videodec-check-if-we-have-a-pool-before-the-l.patch"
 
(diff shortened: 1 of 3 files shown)


recipes-multimedia/gstreamer/gstreamer1.0-plugins-nvvideo4linux2_1.14.0-r32.4.3.bb
```diff
@@ -15,6 +15,9 @@ SRC_URI += "file://build-fixups.patch"
 SRC_URI += "file://0001-v4l2videoenc-Fix-negotiation-caps-leak.patch"
 SRC_URI += "file://0002-v4l2allocator-Fix-data-offset-bytesused-size-validat.patch"
 SRC_URI += "file://0003-v4l2bufferpool-Avoid-set_flushing-warning.patch"
+# https://github.com/OE4T/meta-tegra/issues/486
+SRC_URI += "file://0002-v4l2allocator-Fix-data-offset-bytesused-size-validat.patch"
+SRC_URI += "file://0003-v4l2bufferpool-Avoid-set_flushing-warning.patch"
```
madisongh, line 20: This doesn't look quite right... the patches mentioned here are repeats of the ones above, rather than the new ones being added.

quaresmajose: My bad! sorry for that. I will fix it

madisongh:
I've also ported this to master and gatesgarth, adjusting the patch for the changes in R32.4.4.    Thanks!

PR #489, opened 2020-11-10
gstreamer1.0-plugins-nvvideo4linux2: fix assertion on nvv4l2decoder a…

quaresmajose:
gstreamer1.0-plugins-nvvideo4linux2: fix assertion on nvv4l2decoder allocation query

Avoid run the allocation query on the decoder source pad if the caps has not yet been negotiated.
Runs the the allocation query with a null caps causes an assert and this patch fix this.

It is intended to fix the following:

export G_DEBUG=fatal-criticals GST_DEBUG=7 GST_DEBUG_FILE=/tmp/gst.log
gst-launch-1.0 --gst-debug-no-color videotestsrc ! nvvidconv ! nvv4l2h264enc ! h264parse ! nvv4l2decoder ! nvvidconv ! fakesink

(gst-launch-1.0:4720): GStreamer-CRITICAL **: 16:00:27.748: gst_mini_object_unref: assertion 'mini_object != NULL' failed
Trace/breakpoint trap

Signed-off-by: Jose Quaresma <quaresma.jose@gmail.com>


madisongh:
Thanks!

PR #491, opened 2020-11-11
Add support for the USB device instance to update

sweng5080:
This commit adds support for specifying the USB device instance to update as a new command line argument. This allows supporting multiple NVIDIA devices that may be connected simultaneously with the flash script. The command line usage is as follows:
--usb-instance <id> -- Specify the USB instance to connect to; integer ID (e.g. 0, 1), bus/dev (e.g. 003/091), or USB port path (e.g. 3-14). The latter is best.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra186-flash-helper.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra186-flash-helper.sh
@@ -6,8 +6,9 @@ no_flash=0
 flash_cmd=
 imgfile=
 dataimg=
+inst_args=""
 
-ARGS=$(getopt -n $(basename "$0") -l "bup,no-flash,datafile:" -o "u:v:c:" -- "$@")
+ARGS=$(getopt -n $(basename "$0") -l "bup,no-flash,datafile:,usb-instance:" -o "u:v:c:" -- "$@")
 if [ $? -ne 0 ]; then
     echo "Error parsing options" >&2
     exit 1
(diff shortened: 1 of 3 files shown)


madisongh:
This looks OK to me, thanks... any chance you could update the other flash helper scripts with the same change?

sweng5080:
> any chance you could update the other flash helper scripts with the same change?
I can look into doing that... However, I don't have those platforms to be able to verify that the change works on them properly as it does for the Jetson TX2/t186.

sweng5080:
> > any chance you could update the other flash helper scripts with the same change?
> 
> I can look into doing that... However, I don't have those platforms to be able to verify that the change works on them properly as it does for the Jetson TX2/t186.

Done. I've updated all three SoC variant's flash helper scripts.

madisongh:
Thanks for the additions.  I've verified that the additional option works for t194 and t210 platforms as well.

sweng5080:
> Thanks for the additions. I've verified that the additional option works for t194 and t210 platforms as well.

I'm not familiar with the github tools and workflow. Can I get this change cherry-picked over to the dunfell-l4t-r32.4.3 branch? Or is the normal process workflow to just do another pull request that targets that branch? Or is there github-type cherry-pick option?

madisongh:
I already cherry-picked the change to the gatesgarth and dunfell-l4t-r32.4.3 branches.  I usually do that (manually, as github doesn't really deal with the concept of having multiple release branches, as far as I can tell) when it's appropriate.

PR #494, opened 2020-11-17
cboot recipe refactoring and fixes

madisongh:
* Refactors the cboot recipe (for building from source) into per-SoC recipes for t18x and t19x
* Add a patch for make 4.3 compatibility (for Ubuntu 20.10 and other distros that have updated to the new make)
* Add PACKAGECONFIG mechanism to make it easier to disable some features where needed (particularly for t19x)
* Add a patch to fix A/B support for kernel and device tree loading on SDcard-based Xaviers.
* Fix up insertion of the version string during the build to make the version more identifiable at runtime
* Other minor build cleanup
* Switch to building cboot from sources by default


madisongh:
* All platforms build
* Verified the PACKAGECONFIG mechanism works
* Validated that booting TX2 and Xavier platforms with various configurations of cboot works

Issue #503, opened 2020-11-25
Dunfell / Xavier AGX: do_image_tegraflash fails if the kernel DTS does not contain a bootargs entry


lcintrat:
Hi,

If the kernel DTS does not contain a DTS entry, fdtget returns with exit code 1.

Excerpt from log:
```
+ cp -L /home/build/yocto/build/tmp/deploy/images/evpu/tegra194-p2888-0001-p2822-0000.dtb ./tegra194-p2888-0001-p2822-0000.dtb
+ fdtget ./tegra194-p2888-0001-p2822-0000.dtb /chosen bootargs
+ bootargs=
+ bb_exit_handler
+ ret=1
```
Best regards,
Ludovic

PR #504, opened 2020-11-25
image_tegraflash: fix no bootargs in DTS

lcintrat:
For #503 

If the kernel DTS does not contain an entry for /chosen bootargs,
fdtget returns with exit code 1, then the image task fails.

Signed-off-by: Ludovic Cintrat <ludovic.cintrat@nexvision.fr>


madisongh:
Thanks for this.  I think the right way to fix it is to have the logic for tegra194 platforms match the changes I made in 5f904d946ea48d7fe97938f66d1386eeb2db5c73 for the others - having the `KERNEL_ARGS` setting completely replace whatever might be in the device tree file for the `bootargs` property.

madisongh:
I've gone ahead and merged the change I mentioned, and have back-ported it from master to the gatesgarth, dunfell-l4t-r32.43, and dunfell branches.  Thanks again for finding this!

PR #526, opened 2020-12-21
cuda: add compatible hosts

jasuonpe:
Use COMPATIBLE_HOST to limit the native and nativesdk builds to the
two supported architectures. Otherwise bitbake will throw the
"unrecognized HOST_ARCH" errors during parsing of the recipes,
even when the recipe is not selected to be built.


jasuonpe:
This is cherry-picked https://github.com/OE4T/meta-tegra/commit/660c2c8f2d19ea65dc59364b798dc55d0b5ff265 for zeus. This is needed for getting updated latest meta-tagra changes under QT 5.15.

PR #536, opened 2021-01-21
Adds BUPGEN image names stripping from TEGRA_BUPGEN_STRIP_IMG_NAMES variable

lfdmn:
Adds a built-in way to strip image names from partition layout while generating the BUP payload.

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -536,7 +536,17 @@ tegraflash_generate_bupgen_script() {
     local outfile="${1:-./generate_bup_payload.sh}"
     local spec__ sdramcfg fab boardsku boardrev
     rm -f $outfile
-    cat <<EOF > $outfile
+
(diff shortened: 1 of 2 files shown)


madisongh (changes requested):
Something like this is good to have, but it needs some changes.

Since the BUP generation script is included in the tegraflash package, the changes being inserted into the script will render the flash.xml.in file unusable for regular flashing after creating a BUP payload, and that's not good.

Instead of rewriting flash.xml.in, how about generating a new copy of the file with the filenames stripped out and changing the invocation of the flash-helper script (added down at the bottom of  tegraflash_generate_bupgen_script) to use the name of the copy?

lfdmn:
OK, I'll implement the requested change

madisongh:
I tweaked these changes a bit further and merged them manually into dunfell-l4t-r32.4.3, gatesgarth, and master.

Thanks @lfdmn !

PR #572, opened 2021-02-12
tegra-wifi_1.0.bb: increase udev rules priority

### LINKED ISSUE #571: Wrong name for wlan interface in systemd on Xavier NX
On the Xavier NX, systemd gets the wrong name for the wlan0 interface. This is because the systemd udev rules are applied before the wlan name is set by the tegra-wifi rules.

```
nobi-48eab18f-9fa6-4070-a8d6-2a38f4bc1fc5:/lib/udev/rules.d# ls -la | grep 99
-rw-r--r--    1 root     root           399 Mar  9  2018 90-alsa-restore.rules
-rw-r--r--    1 root     root          4713 Mar  9  2018 99-systemd.rules
-rw-r--r--    1 root     root           251 Mar  9  2018 99-tegra-bluetooth.rules
-rw-r--r--    1 root     root           122 Mar  9  2018 99-tegra-wifi.rules
```

This leads to the following attributes that are set incorrectly:

```
nobi-48eab18f-9fa6-4070-a8d6-2a38f4bc1fc5:/lib/udev/rules.d# udevadm test $(udevadm info -q path -p /sys/class/net/wlan0) 2>&1 | grep wlP4
ID_NET_NAME_PATH=wlP4p1s0
ID_NET_NAME=wlP4p1s0
SYSTEMD_ALIAS=/sys/subsystem/net/devices/wlP4p1s0
run: '/lib/systemd/systemd-sysctl --prefix=/net/ipv4/conf/wlP4p1s0 --prefix=/net/ipv4/neigh/wlP4p1s0 --prefix=/net/ipv6/conf/wlP4p1s0 --prefix=/net/ipv6/neigh/wlP4p1s0'
```

Systemd services that use lines like `After=sys-subsystem-net-devices-%i.device` won't work because of this.

To fix this, the 99-tegra-wifi.rules file should be renamed to 98-tegra-wifi.rules.

nielsavonds:
Make sure tegra-wifi rules are applied before applying systemd rules.
Otherwise, systemd will use an incorrect name for the wifi interface.

Fixes #571 

Signed-off-by: Niels Avonds <niels@nobi.life>

### DIFF
--- a/recipes-bsp/tegra-wifi/tegra-wifi_1.0.bb
+++ b/recipes-bsp/tegra-wifi/tegra-wifi_1.0.bb
@@ -13,7 +13,7 @@ S = "${WORKDIR}"
 do_install() {
     if [ -s ${S}/tegra-wifi.rules ]; then
 	install -d ${D}${nonarch_base_libdir}/udev/rules.d
-	install -m 0644 ${S}/tegra-wifi.rules ${D}${nonarch_base_libdir}/udev/rules.d/99-tegra-wifi.rules
+	install -m 0644 ${S}/tegra-wifi.rules ${D}${nonarch_base_libdir}/udev/rules.d/98-tegra-wifi.rules
     fi
 }
 


madisongh:
Good catch.. thanks!

PR #588, opened 2021-02-17
jetson-tx2-devkit: add BUP support for FAB D02

ichergui:
Jetson TX2 devkit modules got a version bump, so to make
sure that we can build compatible BUP payloads, add that FAB to the
list of specs is needed.


ichergui:
This update in the machine conf file need to be ported to the branches `dunfell-l4t-r32.5.0`, `dunfell-l4t-r32.4.3`, `dunfell-l4t-r32.4.2`, ...

madisongh:
I've cherry-picked this to the other branches, except dunfell-l4t-32.4.2, which is not actively supported.

ichergui:
> I've cherry-picked this to the other branches, except dunfell-l4t-32.4.2, which is not actively supported.

Thanks

PR #600, opened 2021-02-24
recipes-devtools: python: Added recipe for pycuda

nandra:
Signed-off-by: Marek Belisko <marek.belisko@open-nandra.com>


madisongh:
This package has dependencies on packages not in OE-Core, so it can't go in the main part of the layer. Are all of those packages in meta-python? If so, this recipe could go into, say, `external/meta-python/recipes-devtools/python`, so it is made available automatically if the meta-python layer is present (look under `external` for other examples).

madisongh:
@nandra Could you resubmit this over at the (new meta-tegra-community)[https://github.com/OE4T/meta-tegra-community] repo?

PR #611, opened 2021-03-03
tegra-binaries: nvpmodel: support setting NVPMODEL_CONFIG_DEFAULT

kekiefer:
This change allows setting a variable in e.g. your machine configuration
to specify the default setting that nvpmodel will make at first boot.

If left unspecified, the default power model set by nvidia will be used,
as before.

To change the default config to 0 (MAXN), for example, you can add the
following to your machine configuration:

NVPMODEL_CONFIG_DEFAULT = "0"


madisongh (changes requested):
One small thing... could you add a default initialization of the new variable at the top of the recipe?

kekiefer:
Done. Also, I think I'm satisfied that I've put the builder through its paces :P

madisongh (approved):

PR #614, opened 2021-03-04
Refinements to l4t_deb_pkgfeed.bbclass

madisongh:
Integrating VPI is presenting a problem due to the way we compose the SRC_URIs for packages coming from the L4T feeds.

Reworking the logic to calculate the download path on a per-package basis works better for VPI and some of the other L4T packages, and should be more future-proof.  This does require the addition of an L4T_DEB_GROUP setting for some of the recipes, since it's no longer set by default, but we can also drop that setting from some other recipes.


ichergui (approved):
Perfect.
Well done @madisongh 
I tried my recipe concerning VPI, it is working like a charm.

PR #615, opened 2021-03-05
libnvvpi1: add recipe

ichergui:
Add recipe for installing the Nvidia VPI library.


madisongh (changes requested):
* Adding a recipe to build a program against the library that we can use for testing would be best
* Not sure we need the container-runtime-csv support for this package
* The devso-symlink packaging is probably not needed, either

ichergui:
> * Adding a recipe to build a program against the library that we can use for testing would be best
> * Not sure we need the container-runtime-csv support for this package
> * The devso-symlink packaging is probably not needed, either

* Agree 100% that we need at least a sample to test the library
* I was not sure tbh, I will remove it
* I will remove the `devso-symlink` package

Thanks a lot for your feedback

PR #623, opened 2021-03-09
Script fixes for SDcard creation with R32.5.x

### LINKED ISSUE #621: dosdcard.sh does not receive input on prompt
When running the `dosdcard.sh` script, it seems to hang and do nothing when the user enter `yes` or `no`. The issue started happening in the 32.5.0 branch.

The workaround is to press "return" key five times, like shown in this screencast https://asciinema.org/a/NY5HR2kxPeQwohAFUma91PFGi

madisongh:
Fixes #621 by redirecting input for `tegraflash.py` to `/dev/null` so whatever it is in the NVIDIA tools that's messing with the input stream doesn't affect the terminal input, so the confirmation prompt in `make-sdcard` works again.  Only Xavier flash helper needs this, as the problem doesn't manifest itself for Nano (t210).

Also fixes a minor issue in handling unrecognized responses to the confirmation prompt in make-sdcard.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
@@ -215,7 +215,6 @@ confirm() {
 		    ;;
 		*)
 		    echo "Please answer 'yes' or' no'."
-		    break;
 		    ;;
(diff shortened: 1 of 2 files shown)


ichergui (approved):
Looks good

PR #643, opened 2021-03-27
rootfs and kernel on SDcard for AGX Xavier

madisongh:
* Updates the `nvflashxmlparse` script with a function to split a flash layout into two, with the rootfs, kernel, kernel-dtb, and UDA partitions put on a separate SDcard device and the rest of the partitions retained on the on-board eMMC
* Updates the tegraflash bundle to optionally include a `dosdcard.sh` script.

* Only supports the Jetson AGX Xavier dev kit (`MACHINE = "jetson-agx-xavier-devkit"`)
* Does **not** support A/B redundancy.  Boots from the B slot always fail during MB1.

To use, set the following variables in your build configuration:
```
TEGRA_ROOTFS_AND_KERNEL_ON_SDCARD = "1"
ROOTFSPART_SIZE = "15032385536"
```
The rootfs partition size above is for a 16GB SDcard. If you use a different size card, adjust accordingly (but be warned that the SDcard flash XML file may also need modification - the script currently assumes 16GB).

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -56,6 +56,7 @@ IMAGE_TEGRAFLASH_DATA ??= ""
 
 BL_IS_CBOOT = "${@'1' if d.getVar('PREFERRED_PROVIDER_virtual/bootloader').startswith('cboot') else '0'}"
 TEGRA_SPIFLASH_BOOT ??= ""
+TEGRA_ROOTFS_AND_KERNEL_ON_SDCARD ??=""
 
 CBOOTFILENAME = "cboot.bin"
 CBOOTFILENAME_tegra194 = "cboot_t194.bin"
@@ -69,6 +70,7 @@ FLASHTOOLS_DIR_tegra194 = "tegra186-flash"
 
 TEGRAFLASH_PACKAGE_FORMAT ??= "tar"
 TEGRAFLASH_PACKAGE_FORMATS = "tar zip"
+TEGRAFLASH_SDCARD_SIZE ??= "16G"
 
 python() {
     pt = d.getVar('TEGRAFLASH_PACKAGE_FORMAT')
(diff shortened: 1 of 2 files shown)


madisongh:
The sequence for using this, which I forgot to mention is:
1. bitbake the image
2. Unpack the tegraflash package
3. Use `./dodscard.sh` to generate the SDcard image (or burn it directly to the SDcard)
4. Use `./doflash.sh` to flash the bootloader and other boot partitions to the Xavier

After going through the above once, you can then build other images, and as long as you haven't changed the kernel, just `dd` the rootfs onto the APP partition of the SDcard.  If you do change the kernel, or device tree, you only need to re-generate the SDcard image.

Unless you switch between L4T versions, or change cboot, you don't have to re-flash via USB again.

ichergui (approved):
No comments, I already test it and it works fine to me.

madisongh:
Final change: I updated the `nvflashxmlparse` script to set the number of sectors based on the SDcard size, if specified, rather than assuming 16G card size, so if you set `TEGRAFLASH_SDCARD_SIZE` in your build configuration, it will be honored.

Once merged, I'll work on updating/expanding the wiki page on how to flash your Jetson.

chhg82:
Great, it works! Thank you very much!!!
Two questions:
- `MACHINE = "jetson-agx-xavier-devkit"` doesn't exist!? I just configured `MACHINE = "jetson-xavier"` in dunfell-l4t-r32.4.3
- It's a bit confusing with all those size variables with different units for each: `IMAGE_ROOTFS_SIZE`, `ROOTFSPART_SIZE` and `TEGRAFLASH_SDCARD_SIZE`. It would be much easier with just setting `IMAGE_ROOTFS_SIZE` from https://www.yoctoproject.org/docs/current/ref-manual/ref-manual.html#var-IMAGE_ROOTFS_SIZE

madisongh:
> `MACHINE = "jetson-agx-xavier-devkit"` doesn't exist!?

That's the new name in R32.5.0.  If you're on R32.4.3, then yes, it's `jetson-xavier`.

> It's a bit confusing with all those size variables with different units and precedence for each

True enough.  That could probably be cleaned up, or at least better documented.

PR #646, opened 2021-03-31
Support customizable bootlogo blob

kekiefer:
This adds support for packaging a custom bootlogo in the tegraflash package to show at boot. Leaving `PREFERRED_PROVIDER_virtual/bootlogo` set to the default `bootlogo-prebuilt` will package the default `bmp.blob`. An alternative `bootlogo-custom` recipe will use `imagemagick-native` (meta-oe) to resize an arbitrary input image and construct the blob from scratch.

Note that the boot logo is put in the deploy directory. This location is used so that updaters can leverage this artifact to handle as required. The BMP/BMP_b partitions are not part of the bootloader or kernel update payloads.

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -63,6 +63,7 @@ CBOOTFILENAME_tegra194 = "cboot_t194.bin"
 TOSIMGFILENAME = "tos-trusty.img"
 TOSIMGFILENAME_tegra194 = "tos-trusty_t194.img"
 TOSIMGFILENAME_tegra210 = "tos-mon-only.img"
+BMPBLOBFILENAME = "bmp.blob"
 
 BUP_PAYLOAD_DIR = "payloads_t${@d.getVar('NVIDIA_CHIP')[2:]}x"
 FLASHTOOLS_DIR = "${SOC_FAMILY}-flash"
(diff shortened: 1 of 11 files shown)


madisongh:
Nice! Would it make sense to move the OE4T logo and the scripts/code to create custom logos into a separate repo?  We could stash pre-converted BMPs there too, then, so you wouldn't have to have build imagemagick...

kekiefer:
The script to do the compression is pretty light, and I'm comfortable leaving that in here. The logic to make the blob is mostly handled by the nvidia script.

There would be value in being able to utilize pre-converted images to make nicer scaling decisions. The imagemagick build isn't too bad, so I wouldn't get hung up on that.

I'll add something with the option to skip the conversion and use pre canned images if desired. I want to be make sure other layers can easily override this and provide their own logos.

madisongh:
> There would be value in being able to utilize pre-converted images to make nicer scaling decisions. The imagemagick build isn't too bad, so I wouldn't get hung up on that.

OK.  My only two concerns were about keeping binary files in the repo with the metadata and folks needing meta-oe in their layer set to get the logos.  Neither are that big a deal, though.

kekiefer:
OK, I've moved the image files into OE4T/bootlogo-oe4t, and added a PACKAGECONFIG variable that will let you choose to generate from a single image file or use the cooked files, with the cooked as default to avoid needing imagemagick.

kekiefer:
I've added a cboot patch that enables A/B support for the BMP partition. The stock t19x layouts all include BMP/BMP_b.

madisongh:
I've followed this with a couple of additional changes based on some testing I did across the other Jetson models.

Thanks!

PR #655, opened 2021-04-08
Bootloader seamless display transition support

kekiefer:
Patches to support seamless display transition from bootloader.

I assume we will probably apply the linux-tegra patches to the forked repo, but provided here for evaluation.

Taking over the framebuffer requires somewhat delicate handling as there seems to be some inconsistency between the currently set mode, and what the framebuffer emulation thinks the settings are, and the contents of the framebuffer memory don't seem to match what was present in the bootloader (although maybe they should?). Simply mapping the framebuffer won't show anything on the screen until the mode is written back.

First, update the framebuffer contents to match what was written in the bootloader, then:

```
cat /sys/class/graphics/fb0/mode | cat > /sys/class/graphics/fb0/mode
```

From an application, it is also possible, but similarly the initial framebuffer settings read back from the kernel can be incorrect. In particular, the display drivers want to see detailed+cea mode flags, or else it will mark the mode as dirty, even if the timings don't change.

The necessary flags can be seen by looking at the kernel logs. When the framebuffer virtual size is changed, you'll see a line like this:

```
tegradc 15200000.nvdisplay: tegra_dp_get_bpp: vmode=0x10000000 did not specify bpp
```

Then if the other application takes over and the display blinks while the link retrains (at least for displayport), you'll see the following:

```
tegradc 15200000.nvdisplay: blank - powerdown
...
tegradc 15200000.nvdisplay: tegra_dp_get_bpp: vmode=0x10600000 did not specify bpp
```

The vmode flags were what triggered the display update, and those could be added to the framebuffer settings.

```
ioctl(fd, FBIOGET_VSCREENINFO, &vinfo);

vinfo.vmode |= 0x200000;
vinfo.vmode |= 0x400000;

// existing code to mmap and draw any updates, the won't be visible until the following call is made:

ioctl(fd, FBIOPUT_VSCREENINFO, &vinfo);

// continue display progress
```

If using the ioctl method to "match" the display mode set by the bootloader, also check the virtual size settings and make sure the pixel clock matches the mode set by the bootloader.

### DIFF
--- a/recipes-bsp/cboot/cboot-t19x/0013-display-nvdisp-sor-enable-seamless-display-transitio.patch
+++ b/recipes-bsp/cboot/cboot-t19x/0013-display-nvdisp-sor-enable-seamless-display-transitio.patch
@@ -1,0 +1,82 @@
+From e750298e345281ae54c286910a184f650b682246 Mon Sep 17 00:00:00 2001
+From: Kurt Kiefer <kekiefer@gmail.com>
+Date: Wed, 7 Apr 2021 20:42:49 -0700
+Subject: [PATCH 1/1] display: nvdisp/sor: enable seamless display transition
+
+When the CMU is enabled in the bootloader, the Linux tegra display
+drivers will set up a limited range output by default. This caused
+a discontinuity in the image when the range changed from cboot's
+default full range output.
+
+Additionally, the gamma correction was broken. Although an sRGB
+ouput gamma is applied through the CMU on the display as a sane
+default, the window settings weren't applying an sRGB degamma to
+the framebuffer, so gamma was getting applied twice.
+
+This change fixes both problems, enabling a truly seamless display
+transition to Linux.
+
+Signed-off-by: Kurt Kiefer <kekiefer@gmail.com>
+---
+ .../common/drivers/display/nvdisp/tegrabl_nvdisp_cmu.c   | 9 +++++++++
+ .../common/drivers/display/nvdisp/tegrabl_nvdisp_win.c   | 4 ++++
+ .../partner/common/drivers/display/sor/tegrabl_sor.c     | 9 +++++++--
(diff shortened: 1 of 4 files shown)


kekiefer:
So that description is a long way of saying that these patches have value to support in the bootloader/kernel, but there _may_ be a bit more userspace footwork required on a per-application basis, at least there was in my displayport case. I'd be curious to hear if hdmi needs similar handling.

kekiefer:
I should also note that I have the linux kernel console framebuffer disabled. I don't know how this will factor in.

madisongh:
I tried patches on a TX2 with an HDMI monitor attached, and they didn't break anything, but the display isn't seamless, unfortunately.  With HDMI, at least, it looks one of the drivers powers off and on the display a couple of times during startup.  It will take a bit of digging to figure out exactly where that's being triggered.

kekiefer:
The system I'm working with right now only has DisplayPort. I hope to find some time to take a look at the HDMI side soon on a devkit. I suspect that similar changes to these will be required. https://github.com/OE4T/meta-tegra/blob/628d16c0e04c967b071970548c42d529e00b0d1b/recipes-bsp/cboot/cboot-t19x/0013-display-nvdisp-sor-enable-seamless-display-transitio.patch#L62-L80

kekiefer:
I finally got around to looking at this on a devkit with HDMI. It looks like these patches are still doing what they are supposed to do -- they keep the csc and output range settings consistent between the bootloader and kernel. You can see during the initial transition to Linux that the image doesn't change in brightness anymore... just before it gets reset anyway and goes blank. Sigh. I'll spend a bit more time when I have the energy.

kekiefer:
I'm going to close this PR rather than let it languish for longer, and just pull these patches into my custom layer. Hopefully someone will find this useful for their own work in the future.

PR #665, opened 2021-04-14
jetson-nano-2gb-devkit: update PARTITION_LAYOUT_TEMPLATE

madisongh:
which I missed during the update to L4T R32.5.0.


madisongh:
See discussion #661

PR #669, opened 2021-04-16
nsight-systems-cli: add recipe

ichergui:
Add recipe for installing libraries and tools of Nvidia Nsight Systems CLI.

### DIFF
--- a/recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17.bb
+++ b/recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17.bb
@@ -1,0 +1,49 @@
+DESCRIPTION = "NVIDIA Nsight Systems is a multi-core CPU sampling profiler that \
+    provides an interactive view of captured profiling data, helping improve \
+    overall application performance."
+HOMEPAGE = "https://developer.nvidia.com/nsight-systems"
+LICENSE = "Proprietary"
+BASE_VERSION = "${@'.'.join(d.getVar('PV').split('.')[0:3])}"
+LIC_FILES_CHKSUM = " \
+    file://opt/nvidia/${BPN}/${BASE_VERSION}/NVIDIA_SLA.pdf;md5=f7646eea4f29c4f708ea0aa5bc6778a6 \
+"
+
+inherit l4t_deb_pkgfeed
+
+SRC_COMMON_DEBS = "${BPN}-${BASE_VERSION}_${PV}-0256620_arm64.deb;subdir=${BPN}"
+SRC_URI[sha256sum] = "8a1d1bc76ee4a88793e1df26214e0df9aa69e0b435f9e79acdca84a41e7b3c8f"
+
+S = "${WORKDIR}/${BPN}"
+B = "${S}"
+
+COMPATIBLE_MACHINE = "(tegra)"
+
+do_configure() {
+    :
+}
+
+do_compile() {
+    :
+}
+
+do_install() {
+    install -d ${D}/opt/nvidia/${BPN}/${BASE_VERSION}/
+    cp -R --preserve=mode,timestamps,links --no-dereference ${B}/opt/nvidia/${BPN}/${BASE_VERSION}/target-linux-armv8/ ${D}/opt/nvidia/${BPN}/${BASE_VERSION}/
+    cp -R --preserve=mode,timestamps,links --no-dereference ${B}/opt/nvidia/${BPN}/${BASE_VERSION}/host-linux-armv8/ ${D}/opt/nvidia/${BPN}/${BASE_VERSION}/
+    cp -R --preserve=mode,timestamps,links --no-dereference ${B}/opt/nvidia/${BPN}/${BASE_VERSION}/bin/ ${D}/opt/nvidia/${BPN}/${BASE_VERSION}/
+}
+
+INHIBIT_PACKAGE_STRIP = "1"
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
+INHIBIT_SYSROOT_STRIP = "1"
+EXCLUDE_FROM_SHLIBS = "1"
+
+PACKAGES += "${PN}-qdstrmimporter"
+FILES_${PN} = " \
+    /opt/nvidia/${BPN}/${BASE_VERSION}/target-linux-armv8 \
+    /opt/nvidia/${BPN}/${BASE_VERSION}/bin \
+"
+FILES_${PN}-qdstrmimporter = "/opt/nvidia/${BPN}/${BASE_VERSION}/host-linux-armv8"
+INSANE_SKIP_${PN} = "ldflags file-rdeps"
+INSANE_SKIP_${PN}-qdstrmimporter = "ldflags file-rdeps"
+PACKAGE_ARCH = "${TEGRA_PKGARCH}"


ichergui:
* Tests are done on Jetson Xavier AGX and Nano (eMMC verison)
* Here is interesting logs when using the `Nsight systems CLI`

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
```diff
@@ -0,0 +1,53 @@
...
+do_configure() {
+    :
+}
+
+do_compile() {
+    :
+}
+
+do_install() {
+    install -d ${D}/opt/nvidia/nsight-systems-cli/2020.5.3/target-linux-armv8
```
kekiefer, line 29: You should use the `BPN` variable here instead of `nsight-systems-cli`.

Considering that the `2020.5.3` part of the PV gets used in a few places, is there a straightforward way to factor this out?

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
```diff
@@ -0,0 +1,53 @@
...
+    cp -R --preserve=mode,timestamps,links --no-dereference ${B}/opt/nvidia/nsight-systems-cli/2020.5.3/target-linux-armv8/* ${D}/opt/nvidia/nsight-systems-cli/2020.5.3/target-linux-armv8/
```
kekiefer, line 30: You can probably install the whole `target-linux-armv8` directory insteady of using cp here, unless there's just a few things you need out of it.

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
```diff
@@ -0,0 +1,53 @@
...
+    overall application performance."
+HOMEPAGE = "https://developer.nvidia.com/nsight-systems"
+LICENSE = "Proprietary"
+LIC_FILES_CHKSUM = " \
+    file://opt/nvidia/nsight-systems-cli/2020.5.3/NVIDIA_SLA.pdf;md5=f7646eea4f29c4f708ea0aa5bc6778a6 \
+"
+
+inherit nvidia_devnet_downloads
+
+SRC_URI = "${NVIDIA_DEVNET_MIRROR}/nsight-systems-cli-2020.5.3_${PV}_arm64.deb;subdir=nsight-systems-cli"
```
madisongh, line 12: It looks like this package is available off NV's debian package feed.  The name is a bit different, but the sha256sum is identical.  So you could instead use
```
inherit l4t_deb_pkgfeed

SRC_COMMON_DEBS = "${BPN}-2020.5.3_2020.5.3.17-0256620_arm64.deb;subdir=${BPN}"
```
which would make this buildable without having the package in NVIDIA_DEVNET_MIRROR.

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
```diff
@@ -0,0 +1,53 @@
...
+FILES_${PN} = " \
+    /opt/nvidia/nsight-systems-cli/2020.5.3/target-linux-armv8 \
+    /opt/nvidia/nsight-systems-cli/2020.5.3/host-linux-armv8 \
+    /opt/nvidia/nsight-systems-cli/2020.5.3/bin \
+"
+INSANE_SKIP_${PN} = "ldflags"
+
+PRIVATE_LIBS = " \
+    libnvlog.so \
+    libstdc++.so.6 \
```
madisongh, line 50: Ouch... the package includes its own copy of `libstdc++`?  Is it really needed?

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
```diff
@@ -0,0 +1,53 @@
...
+
+inherit nvidia_devnet_downloads
+
+SRC_URI = "${NVIDIA_DEVNET_MIRROR}/nsight-systems-cli-2020.5.3_${PV}_arm64.deb;subdir=nsight-systems-cli"
+SRC_URI[sha256sum] = "8a1d1bc76ee4a88793e1df26214e0df9aa69e0b435f9e79acdca84a41e7b3c8f"
+
+S = "${WORKDIR}/nsight-systems-cli"
+B = "${S}"
+
+COMPATIBLE_MACHINE = "tegra"
```
madisongh, line 18: This should be set to `"(tegra)"` (i.e., add parentheses).

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
```diff
@@ -0,0 +1,53 @@
+SUMMARY = "NVIDIA Nsight Systems is a multi-core CPU sampling profiler that \
```
madisongh, line 1: This is long enough that it should probably be `DESCRIPTION` rather than `SUMMARY`.

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
```diff
@@ -0,0 +1,53 @@
...
+    provides an interactive view of captured profiling data, helping improve \
+    overall application performance."
+HOMEPAGE = "https://developer.nvidia.com/nsight-systems"
+LICENSE = "Proprietary"
+LIC_FILES_CHKSUM = " \
+    file://opt/nvidia/nsight-systems-cli/2020.5.3/NVIDIA_SLA.pdf;md5=f7646eea4f29c4f708ea0aa5bc6778a6 \
```
madisongh, line 7: Is there no plain text copy of the license anywhere in the package?  Plain text would be better.

madisongh (changes requested):

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: Perfect, I tried two days ago with different package name but I didn't successfully download the package.
Thanks.

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: It is not really needed tbh,
There is a tool `QdstrmImporter` taht you can find here `/opt/nvidia/nsight-systems-cli/2020.5.3/host-linux-armv8`
which used to convert files from format `.qdstrm` to `.qdrep`.
For more details, please refer to the [Nvidia documentation](https://docs.nvidia.com/nsight-systems/2020.3/profiling/index.html#importing-qdstrm-files)

kekiefer: Maybe this should be split into a different package?

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: No

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: OK
Agree, I will change it to `DESCRIPTION` instead

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: Should I create a Variable like `PKG_RELEASE_DATE = "2020.5.3"` ? 
What do you think ?

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
madisongh: Yeah, I see.  Oh, well.

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
madisongh: Separate packages for the 'host' tools vs the 'target' sounds like a good idea.

Looking at the package contents, it looks like the target-side directory also contains a bunch of libraries (as well as sqlite3 binary and python interpreter).  I think this recipe is a good candidate for just setting `EXCLUDE_FROM_SHLIBS = "1"` to skip shared library processing for the entire set.

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
madisongh: Maybe extract it from PV?  Something like this:
```
BASE_VERSION = "${@'.'.join(d.getVar('PV').split('.')[0:3])}"
```

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: `host-linux-armv8` is a part of the package `nsight-systems-cli`, it is just a name that nvidia provide.
All libraries and tool are dedicated to be executed in the target.

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: Ah nice

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
kekiefer: That's understood, but you said these so-called host tools don't need to be present for the rest of the package to work. So it makes sense to have this recipe create multiple packages.

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: It works perfectly 
```
$ bitbake -e nsight-systems-cli | grep -i "^BASE_VERSION"
BASE_VERSION="2020.5.3"
$
```

recipes-devtools/nsight-systems/nsight-systems-cli_2020.5.3.17-1.bb
ichergui: Agree, I will create a separate package that contains `/opt/nvidia/nsight-systems-cli/2020.5.3/host-linux-armv8`

ichergui:
* The `nsight-systems-cli` recipe provide 2 main packages: `nsight-systems-cli` and `nsight-systems-cli-qdstrmimporter`
* If only `nsight-systems-cli` is installed within the image the result's files from the nsight systems CLI have an extension `*.qdstrm`
**Example:**
```
root@jetson-nano-devkit-emmc:~# /opt/nvidia/nsight-systems-cli/2020.5.3/bin/nsys profile --output /tmp/deviceQuery.log --show-output false /usr/bin/cuda-samples/deviceQuery          
Collecting data...
Processing events...
Saving temporary "/tmp/nsys-report-1268-251d-5651-84ac.qdstrm" file to disk...

Creating final output files...
Report file moved to "/tmp/deviceQuery.log.qdstrm"

root@jetson-nano-devkit-emmc:~#
```
* when installing the package`nsight-systems-cli-qdstrmimporter` the result's files  have an extension `*.qdrep` as illustrate the above examples.

kekiefer (approved):

kekiefer:
To be clear, is the qdstrmimporter required to run the nsight gui and start/run/analyze profiling sessions from a host computer?

ichergui:
> To be clear, is the qdstrmimporter required to run the nsight gui and start/run/analyze profiling sessions from a host computer?

No, the Nsight systems GUI have this tool integrated in the host package `NsightSystems-linux-public-2020.5.3.17-0256620.deb`.

madisongh (approved):

PR #675, opened 2021-04-21
backport (copy) tegra-eeprom-tool to thud

fei4xu:
I just copied the .bb from `dunfell` (r32.4.3) to `thud` (r32.3.1), it seems to work.

otherwise it cannot compile because `https://github.com/madisongh/tegra-eeprom-tool` link is broken.


madisongh:
Interesting.  The Github automatic redirect from the old URL under my username to the one under OE4T still works for me, but there's no harm in updating.

PR #676, opened 2021-04-23
component: libnvidia-container-tools

FoodWasteWarrior:
added branch specification to SRC_URI of Nvidia modprobe
    this is required as bitbake fails to fetch the recipe otherwise
    Signed-off-by: Fabian Schaltegger fabian.schaltegger@kitro.ch


FoodWasteWarrior:
The branch of the nvidia-modprobe repo has to be specified.
Otherwise bitbake fetching fails.
For modprobe version 396.xx, the branch is called 396 and so on.

madisongh:
We fixed this in later branches by specifying `branch=main`, but the needed commit appears to be in both, so this looks fine.

PR #677, opened 2021-04-24
Switch to downloading CUDA host tools from NVIDIA's deb package feed

madisongh:
NVIDIA recently added and/or documented [host (x86-64) package feeds](https://docs.nvidia.com/jetson/jetpack/install-jetpack/index.html#jetpack-debian-packages) for the CUDA tools. We can use these instead of requiring SDK Manager downloads and the NVIDIA_DEVNET_MIRROR setup.


dwalkes:
Thanks @madisongh I've updated https://github.com/OE4T/meta-tegra/wiki/L4T-Integration-Issues and https://github.com/OE4T/meta-tegra/wiki/NVIDIA_DEVNET_MIRROR-and-SDK-Manager on the wiki.

ichergui:
Well done guys.

PR #685, opened 2021-04-25
classes/nvidia_devnet_downloads: no need for the the NVIDIA_DEVNET_MI…

ichergui:
…RROR setup.

All packages can be downloaded from Nvidia's package feeds.


madisongh:
I wanted to leave this around for a while, in case we run into a situation where it's needed again.

ichergui:
> I wanted to leave this around for a while, in case we run into a situation where it's needed again.

it is a wise decision to keep it (just in case) in the master branch.

I will close the PR.

PR #693, opened 2021-04-27
Thud copy and paste: backport dunfell-r32.4.3 to thud-r32.3.1

fei4xu:
Hi,

this is my dirty backport: copy any new files from dunfell-r32.4.3 to thud-r32.3.1 and make thud-r32.4.3 branch.

Current files are tested to be compilable with yocto-thud. 

I purposely left over some recipes because my original target is to upgrade cude/cudnn/tensorrt package. Maybe that's why the copied-over recipes can compile. I may copy more files but I'd like to stop here.

NOTE:
If some recipe.bbapend file needs the .bb file, or some .bbclass file is missing,  just copy it from yocto-dunfell.
FOR OPENCV:
I'm using opencv 4 for thud https://layers.openembedded.org/layerindex/recipe/97501/ (the only opencv 4 recipe that I can find to compile in thud)


madisongh:
I took this and made some further modifications to complete the R32.4.3 port - see #696.

I reverted the opencv bbappend change to version 4, since meta-oe's opencv recipe is 3.4.3.

I've successfully run test builds on a couple of the Jetson platforms to get some recipe coverage, but haven't tried them all, nor have I tried booting the images that were built.  I'll leave it to you and other folks needing R32.4.3 support thud to submit any further changes/fixes.

Thanks!

PR #695, opened 2021-04-27
weak value assign BMPBLOBFILENAME variable

EasyIP2023:
Signed-off-by: Vincent Davis Jr <vincedav2495@gmail.com>

Added change to BMPBLOBFILENAME variable. If one wants/needs to package up a different bmp.blob file name.

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -63,7 +63,7 @@ CBOOTFILENAME_tegra194 = "cboot_t194.bin"
 TOSIMGFILENAME = "tos-trusty.img"
 TOSIMGFILENAME_tegra194 = "tos-trusty_t194.img"
 TOSIMGFILENAME_tegra210 = "tos-mon-only.img"
-BMPBLOBFILENAME = "bmp.blob"
+BMPBLOBFILENAME ??= "bmp.blob"
 
 BUP_PAYLOAD_DIR = "payloads_t${@d.getVar('NVIDIA_CHIP')[2:]}x"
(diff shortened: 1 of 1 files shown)


kekiefer:
I'm not sure this is going to do what you want. This variable isn't really any more configurable than the `CBOOTFILENAME` or `TOSIMAGEFILENAME` -- the purpose of this variable is internal to the image creation script -- it's not a source of the bmp blob, it's the destination that the blob gets copied into during `oe_make_bup_payload`:

```
oe_make_bup_payload() {
    ...
    export bmpblobfilename=${BMPBLOBFILENAME}
    ...
    cp "${DEPLOY_DIR_IMAGE}/bootlogo-${MACHINE}.blob" ./$bmpblobfilename
    ...
}
```

So unless you're providing your own image creation script, this location is just going to overridden.

The correct way to provide your own bmp blob is to either append one of the existing bootlogo recipes, or create your package that provides `virtual/bootlogo` and deploys the blob to `${DEPLOY_DIR_IMAGE}/bootlogo-${MACHINE}.blob`

EasyIP2023:
@kekiefer I just wanted an easy way to change the destination blob file name. I'm using the updated way for bootlogo blob file creation with a .bbappend. Basically when I extract core-image-minimal-jetson-xavier-nx-devkit.tegraflash.tar.gz. I want to see a .blob file that isn't named bmp.blob. Yes, I have a custom image creation script, with a custom partition configuration file. Although keeping the name bmp.blob is totally fine. Just want an easy way to change that.

EasyIP2023:
I'll just keep the default name. bmp.blob

PR #709, opened 2021-05-12
python3-jetcam: new package

brgl:
This adds a recipe for python3-jetcam - an easy to use Python camera interface for NVIDIA Jetson. It depends on packages available in OE-core and meta-openembedded.

### DIFF
--- a/recipes-devtools/python/python3-jetcam_git.bb
+++ b/recipes-devtools/python/python3-jetcam_git.bb
@@ -1,0 +1,17 @@
+SUMMARY = "Easy to use Python camera interface for NVIDIA Jetson"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://LICENSE.md;md5=295ba21d3c8eb4396383d735495fbe6e"
+
+inherit setuptools3
+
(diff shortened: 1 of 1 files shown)


brgl:
I guess the build fails because the autobuilder doesn't include meta-openembedded & meta-python in the build? Can we make meta-tegra depend on these?

madisongh:
I really don't want to add layer dependencies for the base BSP layer, but we could create either another repo, or create a subdirectory here, with another layer to host these kinds of recipes.  How does that sound?

brgl:
Sounds good! A new repo might be a bit too much for a single recipe. We have `meta-tegra/contrib`. How about `meta-tegra/extra` for this type of jetson-specific additions?

madisongh:
We talked about this at the OE4T monthly meeting today, and decided to go ahead with creating a separate repo.  I've added your commit here and #600 to the [initial population PR](https://github.com/OE4T/meta-tegra-community/pull/1) I've opened in the new repo.

madisongh:
This has been merged into [meta-tegra-community](https://github.com/OE4T/meta-tegra-community).  Thanks!

PR #712, opened 2021-05-17
python3-tensorrt: add recipe

madisongh:

### DIFF
--- a/external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
+++ b/external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
@@ -1,0 +1,59 @@
+SUMMARY = "Python bindings for TensorRT"
+HOMEPAGE = "http://developer.nvidia.com/tensorrt"
+LICENSE = "Proprietary"
+LIC_FILES_CHKSUM = "file://python/packaging/LICENSE.txt;md5=c291e0a531e08d4914e269730ba2f70d"
+
+DEPENDS = "python3-pybind11 tensorrt"
+
+inherit setuptools3 cmake cuda
+
+SRC_REPO = "github.com/NVIDIA/TensorRT.git;protocol=https"
+SRCBRANCH = "master"
+SRC_URI = "git://${SRC_REPO};branch=${SRCBRANCH} \
+           file://0001-Do-not-overwrite-CMAKE_CXX_FLAGS.patch \
+           file://0002-pyOnnx-fixups-for-TensorRT-7.1-API.patch \
+           "
+SRCREV = "ab20a8ac26da03df5e434b4f9d1da0156ca019b6"
+
+S = "${WORKDIR}/git"
+
+OECMAKE_SOURCEPATH = "${S}/python"
+DISTUTILS_SETUP_PATH = "${B}"
+
+EXTRA_OECMAKE = "-DONNX_INC_DIR=${STAGING_INCDIR} -DPYBIND11_DIR=${STAGING_DIR_TARGET} \
+                 -DTARGET_ARCHITECTURE=${HOST_ARCH} -DCMAKE_BUILD_TYPE=Release \
+                 -DPY_INCLUDE=${STAGING_INCDIR}/${PYTHON_DIR}"
+
+CXXFLAGS += "${CUDA_CXXFLAGS}"
+
+do_configure() {
+    cmake_do_configure
+    TRT_MAJOR=$(awk '/^\#define NV_TENSORRT_MAJOR/ {print $3}' ${STAGING_INCDIR}/NvInferVersion.h)
+    TRT_MINOR=$(awk '/^\#define NV_TENSORRT_MINOR/ {print $3}' ${STAGING_INCDIR}/NvInferVersion.h)
+    TRT_PATCH=$(awk '/^\#define NV_TENSORRT_PATCH/ {print $3}' ${STAGING_INCDIR}/NvInferVersion.h)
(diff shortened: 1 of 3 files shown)


external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
```diff
@@ -0,0 +1,56 @@
...
+LICENSE = "Proprietary"
+LIC_FILES_CHKSUM = "file://python/packaging/LICENSE.txt;md5=c291e0a531e08d4914e269730ba2f70d"
+
+DEPENDS = "python3-pybind11 tensorrt"
+
+inherit setuptools3 cmake pkgconfig cuda
+
+SRC_REPO = "github.com/NVIDIA/TensorRT.git;protocol=https"
+SRCBRANCH = "master"
+SRC_URI = "git://${SRC_REPO};branch=${SRCBRANCH} \
```
ichergui, line 12: I think you can remove the `branch=` because will clone the repo by using the `master` branch by default
In this case, no need to define `SRCBRANCH`

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
```diff
@@ -0,0 +1,56 @@
...
+
+    [ -d ${B}/tensorrt ] || mkdir ${B}/tensorrt
+    varsubst ${S}/python/packaging/setup.cfg ${B}/setup.cfg
+    varsubst ${S}/python/packaging/setup.py ${B}/setup.py
+    varsubst ${S}/python/packaging/tensorrt/__init__.py ${B}/tensorrt/__init__.py
+    cp ${S}/python/packaging/LICENSE.txt ${B}/tensorrt
+}
+
+do_compile_append() {
+    distutils3_do_compile
```
ichergui, line 49: I think that `distutils3_do_compile` is not needed because it will be done by default when inheriting `distutils3`

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
```diff
@@ -0,0 +1,56 @@
...
+}
+
+do_install() {
+    distutils3_do_install
```
ichergui, line 53: The same as above

ichergui (commented):
I'm doing a quick test on taking account my comments

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
ichergui: This one is needed

ichergui (changes requested):
I tested the recipe `do_compile_append` is not needed

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
madisongh: I know it's the default, but these are there to make it easier to override the repo location and branch via a bbappend or local.conf setting.

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
madisongh: I've reworked each of the task definitions to explicitly call on the relevant functions from the inherited classes in the right order.  I have left in the distutils3_do_compile function because it does actually do something.  Yes, distutils3_do_install will also execute the compile tasks, but it does so under pseudo context, which could potentially cause some issues.

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_7.1.3-1.bb
ichergui: Agree

ichergui (approved):
Looks good to me.

madisongh:
I'll wait for some feedback from @tirankas before merging.

tirankas:
Hi Matt,
I check before the latest change (since two hours ago) and it worked fine.
I will take the last version now and will verify again

tirankas:
Hi Matt,
With the latest changes (picking the two last commits), I get this:
>>> import tensorrt as trt
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/site-packages/tensorrt/__init__.py", line 35, in <module>
    from .tensorrt import *
ModuleNotFoundError: No module named 'tensorrt.tensorrt'

It worked fine without the last two commits:
>>> import tensorrt as trt
>>> trt.Logger()
<tensorrt.tensorrt.Logger object at 0x7f9d0c6af0>

Not sure what changes broke this...

Thanks



Thanks

madisongh:
Hmm.  OK, thanks, I'll check it out

madisongh:
@tirankas  not sure what happened in your build, but it's working for me.  Can you try a `bitbake -c clean -c cleansstate python3-tensorrt` then build again, just to make sure it's not due to some stale files left around from the earlier build?

tirankas:
Sure, I will check again although I did the same procedure twice ( a to b and then b to a, got consistent results).

tirankas:
Checked again and it is working.
I found one line deleted (within the do_compile function) which lead to the issue, everything works fine now.

Thanks again

ichergui:
Hey @tirankas 
Thanks for the updates. I'm happy that works fine to you.

PR #718, opened 2021-05-27
virglrenderer depends on gbm.h

jwinarske:
Signed-off-by: Joel Winarske <joel.winarske@gmail.com>

This enables the building of qemu package without changing qemu's package config.

### DIFF
--- a/external/virtualization-layer/recipes-graphics/virglrenderer/virglrenderer_%.bbappend
+++ b/external/virtualization-layer/recipes-graphics/virglrenderer/virglrenderer_%.bbappend
@@ -1,0 +1,3 @@
+
+DEPENDS += "virtual/libgbm"
+


ichergui (changes requested):
Could we add only `libgbm-dev` in the dependecy list instead of `virtual/mesa` ?

jwinarske:
> Could we add only `libgbm-dev` in the dependecy list instead of `virtual/mesa` ?

```
ERROR: Nothing PROVIDES 'libgbm-dev' (but /home/joel/tegra-demo-distro/layers/meta/recipes-graphics/virglrenderer/virglrenderer_0.9.1.bb DEPENDS on or otherwise requires it). Close matches:
  libgudev
  mesa RPROVIDES libgbm-dev
ERROR: Required build target 'virglrenderer' has no buildable providers.
Missing or unbuildable dependency chain was: ['virglrenderer', 'libgbm-dev']

Summary: There was 1 WARNING message shown.
```

Mesa will need to come in at some point based on this MR: https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/9902

jwinarske:
Actually I think we want `virtual/libgbm`

ichergui (approved):

madisongh (changes requested):
We can't take this one as is, since we'd be modifying the dependencies for the recipe for all platforms, which is a no-no for a BSP layer.

If it's generally applicable (and the needed header just happens to be present due to transitive dependencies), the fix should be submitted as a patch upstream to OE-Core.

If it's really specific to the tegra targets only, we can do this with `DEPENDS_append_tegra` and then set `PACKAGE_ARCH = "${TEGRA_PKGARCH}"`, so the modification only applies to our platforms.

If it is generally applicable, and you do submit the patch upstream, I'll also accept the modified append approach as a temporary workaround until the patch makes it into OE-Core.

Finally, since `virglrenderer` is in OE-Core, this bbappend belongs in in the core part of the tree, not under `external/virtualization-layer`.

jwinarske:
I think that NVIDIA is the only entity that has avoided Mesa for all these years.  The MR referenced is indicating that is about to change; it enables loading of proprietary driver.

I suspect if it were an issue on other targets it would have already been accounted for upstream.

Given gbm is a hard dependency for virglrenderer, and tegra being the exception perhaps `DEPENDS_append_tegra` is sufficient?

jwinarske:
I'll ask upstream about it

jwinarske:
Patch sent up stream for poky recipe, waiting on review.

madisongh:
Looks like the upstream patch [made it in](https://git.openembedded.org/openembedded-core/commit/?id=fd807dfa2ec25111dd359ec87d197020e747cf0d), so we should be OK to close this.

PR #721, opened 2021-05-31
conf/machine: enable serial uart via the micro-USB connector J501

ichergui:

* here is the logs from the serial UART
[serial_uart_xavier_uefi.log](https://github.com/OE4T/meta-tegra/files/6572138/serial_uart_xavier_uefi.log)

### DIFF
--- a/conf/machine/include/agx-xavier-uefi.inc
+++ b/conf/machine/include/agx-xavier-uefi.inc
@@ -1,4 +1,4 @@
-SERIAL_CONSOLES ?= "115200;ttyS0"
+SERIAL_CONSOLES ?= "115200;ttyTCU0"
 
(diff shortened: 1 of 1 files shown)


madisongh:
Hmm, OK.  It never occurred to me to try this, given NVIDIA's instructions about using UEFI.  `ttyS0` is also available on the microUSB connector, on the third `ttyUSB` port it exposes to the host, rather than the fourth, but it would be nice to have to some consistency.  I'll give it a try.

madisongh:
OK, so I think NVIDIA uses `ttyS0` instead of `ttyTCU0` because ACPI doesn't expose the latter device, so you have to use the other console port in order to switch between devicetree mode and ACPI.

Since the eMMC also shows up differently between the two (at least with the 5.12 kernel we're using), I guess that's not a big deal.

PR #723, opened 2021-06-02
tegra186-flashtools-native: patch to convert BMP_generator_L4T.py to …

kk6ho:
…python3

Change python shebang to python3 shebang and
other python3 conversions (print and string encodes)

bootlogo-custom recipe was unable to build due to python2
build dependency

Signed-off-by: Geoff Parker <geoffrey.parker@arthrex.com>

### DIFF
--- a/recipes-bsp/tegra-binaries/files/0012-BMP_generator_L4T.py-to-Python3.patch
+++ b/recipes-bsp/tegra-binaries/files/0012-BMP_generator_L4T.py-to-Python3.patch
@@ -1,0 +1,71 @@
+--- Linux_for_Tegra/tools/bmp-splash/BMP_generator_L4T.py.orig	2021-06-02 02:26:34.460334063 +0000
++++ Linux_for_Tegra/tools/bmp-splash/BMP_generator_L4T.py	2021-06-02 02:54:01.965137867 +0000
+@@ -1,4 +1,4 @@
+-#!/usr/bin/python
++#!/usr/bin/env python3
+ #
+ # Copyright (c) 2017-2020, NVIDIA CORPORATION.  All rights reserved.
+ #
+@@ -26,7 +26,7 @@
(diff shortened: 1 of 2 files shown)


kk6ho:
I would like this applied to the hardknott branch as well.

Not sure if my 0012-BMP_generator_L4T.py-to-Python3.patch should be applied to the tegra210-flashtools-native_32.5.1.bb recipe as well

kekiefer:
The `BMP_generator_L4T.py` that you've patched here gets copied into the native flash tools package for both platforms, so this same patch should apply for tegra210. Bootlogo is also valid for that platform, so patching it as well would be prudent.

madisongh:
I guess I haven't seen this problem because my build hosts still have Python 2 installed on them as `/usr/bin/python`.

kk6ho:
> The `BMP_generator_L4T.py` that you've patched here gets copied into the native flash tools package for both platforms, so this same patch should apply for tegra210. Bootlogo is also valid for that platform, so patching it as well would be prudent.

Ok, I'll send another PR to add the patch for tegra210.

kk6ho:
> I guess I haven't seen this problem because my build hosts still have Python 2 installed on them as `/usr/bin/python`.

I started seeing this error once I upgraded my build host from Debian 10 to Debian 11 which triggered a rebuild.  I had also tried adding python2 to my Debian 11 host, but the issue persisted. I may have forgotten to clean all the recipes involved.

madisongh:
> Ok, I'll send another PR to add the patch for tegra210.

That's OK... I've already tested with the patch carried over to the tegra210 tools, and it works fine.  I'll get that committed shortly, then port them both over to other branches.

madisongh:
Oh, and thanks!

EasyIP2023:
Prefect!

PR #729, opened 2021-06-07
Add cboot extlinux.conf support

madisongh:
cboot kind of, sort of, supports booting a kernel image and initrd in the rootfs, and on the t194 platforms looks for `/boot/extlinux/extlinux.conf` in the rootfs (or the first partition, if no `root=` is present in the kernel args in the device tree).

This patch series adds a recipe for creating an extlinux setup on t194 platforms, including signing the kernel image and initrd. It also changes the default KERNEL_IMAGETYPE for t194 and t186 (w/cboot) platforms to be `Image.gz`, to reduce the size of the file so cboot can load it successfully.

There are also some adjustments to remove the `cbo.dtb` cboot options file by default, to match L4T's default, which I discovered when testing.


kekiefer:
One of the issues with the t194 + extlinux support was that cboot had broken ext4 support that does not properly handle sparse files, as created by meta-oe ext4 rootfs image creation, which affected the ability for cboot to properly read the kernel image from the rootfs (IIRC). In your testing, is this still an issue?

madisongh:
@kekiefer Yes it was, and thanks for reminding me what the root cause of the problem was. I've added a patch to cboot to fix the sparse file handling (which, works OK, AFAICT).

Using a gzip-compressed kernel image sort of worked around the problem, since the compression should eliminate any all-zeros runs that would result in a gap between extents.

kekiefer:
This is great, as always. I was successful out of the box running this on t194 with the gz kernel image, and I'm playing around with different options a bit more.

One thought. It would be more standard to leave the packaging of the kernel up to the existing kernel classes (`kernel-image-image`, `kernel-image-image.gz`). To keep the signature connected, that may want to be an image type as well, with a dependency chain like `cboot-extlinux` -> `kernel-image-${KERNEL_IMAGETYPE}.sig` -> `kernel-image-${KERNEL_IMAGETYPE}`. It does complicate things a little, but again does have a standard Yocto packaging structure this way. Just thought I should bring this up for discussion.

madisongh:
> One thought. It would be more standard to leave the packaging of the kernel up to the existing kernel classes

I think it's all pretty standard - you select the kernel "image" type via the KERNEL_IMAGETYPE variable, and you can build others by adding them to KERNEL_IMAGETYPES - all handled by the kernel build directly.  I set some defaults in the machine configs, but you can override them easily enough.  We fortunately do not have to turn them into image/conversion types (for `image.bbclass`), as the kernel "image"  doesn't go through that path.

kekiefer:
Yes, I understand. The comment was about the package where the kernel image normally comes from -- the `kernel-image-image.gz` package is already there to install a versioned copy of the kernel into `/boot` and symlink it to `Image`. The idea being that building the virtual/kernel recipe and updating the system with a package manager would naturally update the kernel image and modules, plus the signature if a kernel image class was added for this.

It's really a minor point. Maybe I'll fiddle with it and see if it's worth doing.

I do have one patch to cboot that I found useful, this allows timeouts less than one second to be used:

```
From cf72aa447e7f013b24bd69f1ada82a674236a52d Mon Sep 17 00:00:00 2001
From: Kurt Kiefer <kekiefer@gmail.com>
Date: Tue, 8 Jun 2021 14:16:10 -0700
Subject: [PATCH] extlinux: support timeouts under 1 sec

Signed-off-by: Kurt Kiefer <kekiefer@gmail.com>
---
 bootloader/partner/common/lib/linuxboot/extlinux_boot.c | 2 +-
 1 file changed, 1 insertion(+), 1 deletion(-)

diff --git a/bootloader/partner/common/lib/linuxboot/extlinux_boot.c b/bootloader/partner/common/lib/linuxboot/extlinux_boot.c
index 908a7e6..45ee903 100644
--- a/bootloader/partner/common/lib/linuxboot/extlinux_boot.c
+++ b/bootloader/partner/common/lib/linuxboot/extlinux_boot.c
@@ -193,7 +193,7 @@ static tegrabl_error_t parse_conf_file(void *conf_load_addr, struct conf *extlin
 	if (timeout != NULL) {
 		/* extlinux.conf timeout is 1/10 of a second */
 		user_input_wait_timeout_ms = tegrabl_utils_strtoul(timeout, NULL, BASE_10);
-		user_input_wait_timeout_ms = (user_input_wait_timeout_ms / 10UL) * TIME_1MS;
+		user_input_wait_timeout_ms = (user_input_wait_timeout_ms * TIME_1MS) / 10UL;
 	} else {
 		user_input_wait_timeout_ms = 0UL;
 	}
-- 
2.29.2
```

madisongh:
> It's really a minor point. Maybe I'll fiddle with it and see if it's worth doing.

OK, I see now.  Not sure it's worth the bother, given the limited application of the `.sig` files...

Thanks for the patch.  I did run into the problem of using a timeout < 10 resulting in infinite wait when I was testing. :)

madisongh:
As an aside, I think the reason that you don't run into the sparse-file problem with stock L4T is because the `l4t_sign_image.sh` script is run on the files *after* they're copied into the rootfs, and that script uses `dd` to split apart the contents and signature files after the signing is done.  `dd` doesn't create sparse files.

PR #742, opened 2021-06-22
nvflashxmlparse changes from master + latest tegra-boot-tools

atharvanan1:
* adds changes from master for changes in `nvflashxmlparse` script
* updates tegra-boot-tools to 2.4.0


madisongh:
Have you tested this with the Nano platforms?  The NVIDIA flashing tools in R32.4.3 behave differently than the ones in R32.4.4 and later (see af65d23c6da28177bc6746eaf5823e0f2e2e1b29), and I don't think a straight back-port is going to produce the correct results.

madisongh:
I have added another change to nvflashxmlparse.py to make it compatible with the R32.4.3 tools (which ignore the `id` elements in the XML files) and merged this manually.

atharvanan1:
Thanks @madisongh for taking care of that!

PR #745, opened 2021-06-29
gstreamer1.0-plugins-tegra: Remove libgstbadbase-1.0 dependency.

Lexmark-chad:
This library was removed from the gst-bad-plugins repo in
gstreamer 1.13.1 per this commit:

d01297e115e475760764d9caa4eb37b5f7f9f31a

Signed-off-by: Chad McQuillen <chad.mcquillen@lexmark.com>


madisongh:
Thanks!

PR #753, opened 2021-07-14
Add workaround to fix nvvideo4linux2 build

hatf0:
With GCC 10+, `-fcommon` is no longer the default. This is not a significant issue for the majority of packages, as they have been updated to reflect this change. However, nvvideo4linux2 has not, and requires this work-around to successfully build with GCC 10+.

Please let me know if this isn't the desired way to fix this issue!

### DIFF
--- a/recipes-multimedia/gstreamer/gstreamer1.0-plugins-nvvideo4linux2_1.14.0-r32.5.0.bb
+++ b/recipes-multimedia/gstreamer/gstreamer1.0-plugins-nvvideo4linux2_1.14.0-r32.5.0.bb
@@ -27,6 +27,7 @@ DEPENDS = "gstreamer1.0 glib-2.0 gstreamer1.0-plugins-base virtual/egl tegra-lib
 PACKAGECONFIG ??= "libv4l2"
 PACKAGECONFIG[libv4l2] = ",,v4l-utils,tegra-libraries-libv4l-plugins"
 EXTRA_OEMAKE = "${@bb.utils.contains('PACKAGECONFIG', 'libv4l2', 'USE_LIBV4L2=1', '', d)}"
+CFLAGS += "-fcommon"
 
 S = "${WORKDIR}/gst-v4l2"
 


recipes-multimedia/gstreamer/gstreamer1.0-plugins-nvvideo4linux2_1.14.0-r32.5.0.bb
```diff
@@ -27,6 +27,7 @@ DEPENDS = "gstreamer1.0 glib-2.0 gstreamer1.0-plugins-base virtual/egl tegra-lib
 PACKAGECONFIG ??= "libv4l2"
 PACKAGECONFIG[libv4l2] = ",,v4l-utils,tegra-libraries-libv4l-plugins"
 EXTRA_OEMAKE = "${@bb.utils.contains('PACKAGECONFIG', 'libv4l2', 'USE_LIBV4L2=1', '', d)}"
+TARGET_CFLAGS += "-fcommon"
```
madisongh, line 30: You can just use `CFLAGS` here, instead of `TARGET_CFLAGS`.  Probably doesn't matter much for this recipe, but that's the more common usage.

madisongh:
Could you update the commit and also add a signoff line per our [contribution guidelines](https://github.com/OE4T/meta-tegra/blob/master/CONTRIBUTING.md)?

I suspect this isn't the only recipe that will need something like this. Most folks use an older compiler for CUDA/nvcc compatibility, so building with the latest compiler in OE-Core doesn't happen very often.  Thanks!

hatf0:
Updated commit to add sign-off, and changed the env to `CFLAGS` per suggestion.

> I suspect this isn't the only recipe that will need something like this. Most folks use an older compiler for CUDA/nvcc compatibility, so building with the latest compiler in OE-Core doesn't happen very often. Thanks!

Actually, this was the only recipe that was causing trouble! Every other recipe built just fine -- at least, on ```gcc version 10.2.0 (GCC)```

PR #756, opened 2021-07-16
cboot-*: add patch to fix symlink support in ext2_dir_lookup

ricardosalveti:
ext2_dir_lookup was incorrectly using the original path string while
recursively performing a lookup on a symlink, causing cboot to crash
when trying to recursively identify the link node.

Signed-off-by: Ricardo Salveti <ricardo@foundries.io>


ricardosalveti:
@madisongh found this one while adding support for ostree, which uses syslinux.cfg as a link to the config file managed by the ostree.

There is also one extra patch required for it to work with ostree (using it together with https://github.com/advancedtelematic/meta-updater) that I added at https://github.com/foundriesio/meta-lmp/commit/c4641c92bf21678153f85b85d8abfd6018b9cc0e#diff-e9da64b53da5bbb0af07d898d031dcb4647003967bf66974594dda975399df82 (the distro we maintain that is ostree-based). Let me know if this is something that you would like to see available via meta-tegra as well.

madisongh:
Looks good, thanks!  And yes, I'd be interested in the additional fixes for the extlinux/syslinux handler, too. I know we have some meta-tegra users that would be interested in the ostree integration as well.

PR #775, opened 2021-08-12
l4t_deb_pkgfeed.bbclass: fix for new override syntax

kekiefer:
Fixed a missing update for the new override syntax. This fixes
an issue with l4t .deb packages using zstd compression not extracting
properly.


madisongh:
Thanks!

PR #794, opened 2021-09-08
Backport R32.6.1 updates to Hardknott

ricardosalveti:
I noticed that both master and dunfell are both based on R32.6.1, but hardknott is still on R32.5.2, and since it is the current stable (not LTS) I wanted to open this PR to see if it also makes sense, from the meta-tegra upstream perspective, to align the same changes there.

The only thing missing from the list is the usual u-boot-tegra bump as we first need to land the required changes to the branch used by hardknott. I can propose the changes if you're fine to move hardknott to R32.6.1.

There is one regression I noticed with the latest cboot that I still have to debug (unable to boot my ostree deployment), but it is not related to this backporting PR (will create an additional one with the fix against master once I get to debug it).


madisongh:
My plan was to leave hardknott at 32.5.x and have the imminent honister release track with 32.6.x.  I try not to do major L4T updates on the branches that track stable OE-Core releases once they've been pulled (dunfell has been an exception because of its LTS status, and I'm still trying to figure out how best to handle LTS-tracking branches going forward).

ichergui:
I wish to add something to what @madisongh is saying, `Hardknott` EOL is for  soon (November 2021).
Please refer to this [link](https://wiki.yoctoproject.org/wiki/Releases)

ricardosalveti:
Sure, that is fine, it is just confusing as currently LTS is ahead of the current stable release (hardknott).

Will close the PR then, thanks!

PR #799, opened 2021-09-14
Build Tensorrt Plugins from source

ichergui:
In this PR: 

* Split tensorrt recipe to `tensorrt-core`, `tensorrt-plugins-prebuilt` and `tensorrt-samples` recipes.
* Add new recipe to build `tensorrt-plugins` from source.
* Update dependency list in recipes which depends on `tensorrt-core` and `tensorrt-plugins`.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -51,7 +51,7 @@ SIGGEN_EXCLUDE_SAFE_RECIPE_DEPS += "\
   cuda-nvrtc->tegra-libraries \
   cuda-nvtx->tegra-libraries \
   libcublas->tegra-libraries \
-  tensorrt->tegra-libraries \
+  tensorrt-core->tegra-libraries \
   libdrm->libdrm-nvdc \
   libdrm-nvdc->tegra-libraries \
   tegra-nvs-service->tegra-nvs-base \
--- a/external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
+++ b/external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
@@ -1,0 +1,50 @@
+DESCRIPTION = "NVIDIA TensorRT Plugins for deep learning"
+HOMEPAGE = "http://developer.nvidia.com/tensorrt"
+LICENSE = "Apache-2.0 & BSD-3-Clause & MIT"
+LIC_FILES_CHKSUM = " \
+  file://LICENSE;md5=a55f7353e48c6734c10531db0b040177 \
+  file://third_party/cub/LICENSE.TXT;md5=20d1414b801e2a130d7d546685105508 \
+  file://parsers/onnx/third_party/onnx/LICENSE;md5=efff5c5110f124a1e2163814067b16e7 \
+  file://parsers/onnx/LICENSE;md5=e149467d209874a32715c20e9fdafac5 \
+"
+
+inherit cuda cmake container-runtime-csv
+
+SRC_REPO = "github.com/NVIDIA/TensorRT.git;protocol=https"
+SRCBRANCH = "master"
+SRC_URI = "gitsm://${SRC_REPO};branch=${SRCBRANCH} \
+    file://0001-CMakeLists.txt-fix-cross-compilation-issues.patch \
+"
+SRCREV = "eb5de99b523c76c2f3ae997855ad86d3a1e86a31"
+
+S = "${WORKDIR}/git"
+
+DEPENDS += "zlib libcublas cudnn cuda-cudart cuda-nvrtc protobuf protobuf-native tensorrt-core"
+
+COMPATIBLE_MACHINE = "(tegra)"
+
+CONTAINER_CSV_FILES = "${libdir}/*.so*"
+
+PACKAGECONFIG ??= " \
+    plugin \
+    parsers \
+"
+PACKAGECONFIG[plugin] = "-DBUILD_PLUGINS=ON,-DBUILD_PLUGINS=OFF,"
+PACKAGECONFIG[parsers] = "-DBUILD_PARSERS=ON,-DBUILD_PARSERS=OFF,"
+
+EXTRA_OECMAKE = "-DBUILD_SAMPLES=OFF -DSKIP_GPU_ARCHS=ON -DTRT_PLATFORM_ID="${TARGET_ARCH}" \
+  -DCUDA_VERSION="${CUDA_VERSION}" -DCUDA_NVCC_FLAGS="${CUDA_NVCC_FLAGS}" \
+  -DCUDA_INCLUDE_DIRS="${STAGING_DIR_HOST}/usr/local/cuda-${CUDA_VERSION}/include" \
+"
+
+LDFLAGS += "-Wl,--no-undefined"
+
+do_install:append() {
+    install -d ${D}${includedir}
+    install -m 0644 ${S}/include/NvInferPlugin.h ${D}${includedir}
+    install -m 0644 ${S}/include/NvInferPluginUtils.h ${D}${includedir}
+    install -m 0644 ${S}/include/NvOnnxConfig.h ${D}${includedir}  
+    install -m 0644 ${S}/parsers/onnx/NvOnnxParser.h ${D}${includedir}
+}
+
+RDEPENDS:${PN} += "tegra-libraries"
--- a/recipes-devtools/gie/tensorrt-core_8.0.1-1.bb
+++ b/recipes-devtools/gie/tensorrt-core_8.0.1-1.bb
@@ -1,0 +1,51 @@
+DESCRIPTION = "NVIDIA TensorRT Core (GPU Inference Engine) for deep learning"
+LICENSE = "Proprietary"
+
+inherit l4t_deb_pkgfeed container-runtime-csv
+
+HOMEPAGE = "http://developer.nvidia.com/tensorrt"
+
+L4T_DEB_GROUP = "tensorrt"
+
+SRC_COMMON_DEBS = "\
+    libnvinfer8_${PV}+cuda10.2_arm64.deb;downloadfilename=libnvinfer8_${PV}+cuda10.2_arm64.deb;name=lib;subdir=tensorrt \
+    libnvinfer-dev_${PV}+cuda10.2_arm64.deb;downloadfilename=libnvinfer-dev_${PV}+cuda10.2_arm64.deb;name=dev;subdir=tensorrt \
+    libnvparsers8_${PV}+cuda10.2_arm64.deb;downloadfilename=libnvparsers8_${PV}+cuda10.2_arm64.deb;name=nvp;subdir=tensorrt \
+    libnvparsers-dev_${PV}+cuda10.2_arm64.deb;downloadfilename=libnvparsers-dev_${PV}+cuda10.2_arm64.deb;name=nvpdev;subdir=tensorrt \
+"
+LIBSHA256SUM = "305c4482a315ceb59e514823b359fdeebfbdd5fa2124e277dd176589e2f49aea"
+DEVSHA256SUM = "f750c910a23107715dc2510d360725e46e9072079caacf3cec4255dd38bee849"
+NVPSHA256SUM = "34040352c9f44611928a7d6aa6a7f885b6506a7b3310a8b9fc0782a9ba42037a"
+NVPDEVSHA256SUM = "b09864c351aebf2200fb98f48dc68b4a75260bbcd01423bbf1633acdc115b9be"
+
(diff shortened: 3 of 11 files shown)


ichergui:
#793

recipes-devtools/gie/tensorrt-core_8.0.1-1.bb
```diff
@@ -0,0 +1,49 @@
...
+}
+
+do_install() {
+    install -d ${D}${includedir}
+    install -m 0644 ${S}/usr/include/aarch64-linux-gnu/*.h ${D}${includedir}
+    install -d ${D}${libdir}
+    cp --preserve=mode,timestamps,links --no-dereference ${S}/usr/lib/aarch64-linux-gnu/*.so* ${D}${libdir}
+}
+
+RDEPENDS:${PN} += "libstdc++ cudnn libcublas cuda-cudart cuda-nvrtc cuda-nvtx tegra-libraries libglvnd"
```
madisongh, line 47: This list may need some revising with the split. In fact, it would probably be better to use DEPENDS for the dependencies and let the shlibs processing during packaging figure out the runtime dependencies.

recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
```diff
@@ -0,0 +1,52 @@
+DESCRIPTION = "NVIDIA TensorRT Plugins for deep learning"
+HOMEPAGE = "http://developer.nvidia.com/tensorrt"
+LICENSE = "Apache-2.0"
+LIC_FILES_CHKSUM = "file://LICENSE;md5=a55f7353e48c6734c10531db0b040177"
+
+inherit cuda cmake 
+
+OECMAKE_GENERATOR = "Unix Makefiles"
```
madisongh, line 8: This line shouldn't be necessary.  ninja works perfectly fine for the build.

recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
```diff
@@ -0,0 +1,52 @@
+DESCRIPTION = "NVIDIA TensorRT Plugins for deep learning"
+HOMEPAGE = "http://developer.nvidia.com/tensorrt"
+LICENSE = "Apache-2.0"
```
madisongh, line 3: There are more licenses than just Apache in the included code, from what I can tell.

recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
```diff
@@ -0,0 +1,52 @@
...
+SRCBRANCH = "master"
+SRC_URI = "gitsm://${SRC_REPO};branch=${SRCBRANCH} \
+    file://0001-CMakeLists.txt-fix-cross-compilation-issues.patch \
+"
+SRCREV = "eb5de99b523c76c2f3ae997855ad86d3a1e86a31"
+
+S = "${WORKDIR}/git"
+B = "${S}"
+
+DEPENDS += "zlib libcublas cudnn cuda-cudart cuda-nvrtc protobuf protobuf-native tensorrt-core"
```
madisongh, line 20: The dependency on protobuf, which lives in meta-oe, means that this recipe cannot be located here.  It will have to be moved under `external/openembedded-layer`.

Alternatively, we could revert to using the included protobuf sources and remove this dependency, although I'm not crazy about that idea.

madisongh (changes requested):
See in-line comments

conf/layer.conf
```diff
@@ -51,7 +51,8 @@ SIGGEN_EXCLUDE_SAFE_RECIPE_DEPS += "\
   cuda-nvrtc->tegra-libraries \
   cuda-nvtx->tegra-libraries \
   libcublas->tegra-libraries \
-  tensorrt->tegra-libraries \
+  tensorrt-core->tegra-libraries \
+  tensorrt-plugins->tegra-libraries \
```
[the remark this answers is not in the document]
ichergui: Good catch  :)

recipes-devtools/gie/tensorrt-core_8.0.1-1.bb
ichergui: I need to double check the dependency list, I will update it if needed

recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
ichergui: I moved the recipe under `external/openembedded-layer` for now. Agree, I prefer to use the internal `Protobuf`  to avoid depending on `protobuf` and  `protobuf-native`

recipes-devtools/gie/tensorrt-core_8.0.1-1.bb
ichergui: I updated the `RDEPENDS` list

recipes-devtools/gie/tensorrt-samples_8.0.1-1.bb
```diff
@@ -0,0 +1,49 @@
...
+    TARGET="${TARGET_ARCH}" \
+'
+
+do_configure() {
+    :
+}
+
+do_install() {
+    install -d ${D}${prefix}/src/tensorrt/bin
+    cp --preserve=mode,timestamps ${S}/../bin/sample_* ${D}${prefix}/src/tensorrt/bin
```
madisongh, line 39: It looks like the makefiles build both "release" and "debug" versions of the programs.  Some patching or EXTRA_OEMAKE overrides would be in order so we only build one copy and let packaging handle the debug info separation.  We should also add our LDFLAGS in so you don't have to set `ldflags` in INSANE_SKIP.

You could also just use `install` here, rather than the more risk `cp`.

It would also be good to install `trtexec` (and maybe package that separately from the samples, as NV does).

recipes-devtools/gie/tensorrt-samples_8.0.1-1.bb
```diff
@@ -0,0 +1,49 @@
...
+    install -d ${D}${prefix}/src/tensorrt/data
+    cp -R --preserve=mode,timestamps ${S}/../data/* ${D}${prefix}/src/tensorrt/data
+}
+
+FILES:${PN} += "${prefix}/src/tensorrt/bin ${prefix}/src/tensorrt/data"
+RDEPENDS:${PN} += "tensorrt-core tensorrt-plugins tegra-libraries libglvnd"
+
+INSANE_SKIP:${PN} += "ldflags"
+
+PACKAGE_ARCH = "${TEGRA_PKGARCH}"
```
madisongh, line 49: This PACKAGE_ARCH setting should be removed.  The `cuda` bbclass sets the appropriate value.

external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
```diff
@@ -0,0 +1,50 @@
...
+    install -d ${D}${includedir}
+    install -m 0644 ${S}/include/NvInferPlugin.h ${D}${includedir}
+    install -m 0644 ${S}/include/NvInferPluginUtils.h ${D}${includedir}
+    install -m 0644 ${S}/include/NvOnnxConfig.h ${D}${includedir}  
+    install -m 0644 ${S}/parsers/onnx/NvOnnxParser.h ${D}${includedir}
+}
+
+RDEPENDS:${PN} += "tensorrt-core"
+
+PACKAGE_ARCH = "${TEGRA_PKGARCH}"
```
madisongh, line 50: This PACKAGE_ARCH settings should be removed - it will be set by `cuda.bbclass`.

madisongh (changes requested):
It's getting there.  Just a couple of minor in-line comments.

recipes-devtools/gie/tensorrt-samples_8.0.1-1.bb
ichergui: Done

external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
ichergui: Done

recipes-devtools/gie/tensorrt-samples_8.0.1-1.bb
ichergui: Update the patch and install `trtexec` in a separate package.

recipes-devtools/gie/tensorrt-samples_8.0.1-1.bb
```diff
@@ -0,0 +1,72 @@
...
+    cp -R --preserve=mode,timestamps ${S}/../data/* ${D}${prefix}/src/tensorrt/data
+}
+
+PACKAGES =+ "${PN}-trtexec"
+
+FILES:${PN} += "${prefix}/src/tensorrt/bin ${prefix}/src/tensorrt/data"
+FILES:${PN}-trtexec += "${prefix}/src/tensorrt/bin/trtexec"
+
+RDEPENDS:${PN} += "tensorrt-core tensorrt-plugins tegra-libraries libglvnd"
+RDEPENDS:${PN}-trtexec += "tensorrt-core tensorrt-plugins tegra-libraries libglvnd"
```
madisongh, line 72: I don't think we need these RDEPENDS additions, unless some of the libraries are using dlopen() at runtime to pull libraries in from these packages - shlibs processing during packaging should handle setting the runtime dependencies automatically.

recipes-devtools/gie/tensorrt-plugins-prebuilt_8.0.1-1.bb
```diff
@@ -57,28 +42,27 @@ do_configure() {
...
 }
-PACKAGES += "${PN}-samples"
+
 FILES:${PN} += "${prefix}/src/tensorrt/bin"
-FILES:${PN}-samples = "${prefix}/src"
 
-RDEPENDS:${PN} += "libstdc++ cudnn libcublas cuda-cudart cuda-nvrtc cuda-nvtx tegra-libraries libglvnd"
-RDEPENDS:${PN}-samples += "tegra-libraries bash python3 libglvnd cudnn cuda-cudart libcublas"
-RPROVIDES:${PN}-samples = "${PN}-examples"
+RDEPENDS:${PN} += "tensorrt-core"
```
madisongh, line 59: I think `tegra-libraries` is also needed here, since `libnvinfer_plugin.so` does a `dlopen()` on the `libcuda.so` provided there.  `tensorrt-core` shouldn't be needed, since it will get pulled in due to linking against `libnvinfer.so`.

external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
```diff
@@ -0,0 +1,50 @@
...
+
+do_install:append() {
+    install -d ${D}${includedir}
+    install -m 0644 ${S}/include/NvInferPlugin.h ${D}${includedir}
+    install -m 0644 ${S}/include/NvInferPluginUtils.h ${D}${includedir}
+    install -m 0644 ${S}/include/NvOnnxConfig.h ${D}${includedir}  
+    install -m 0644 ${S}/parsers/onnx/NvOnnxParser.h ${D}${includedir}
+}
+
+RDEPENDS:${PN} += "tensorrt-core"
```
madisongh, line 50: I think `tegra-libraries` is also needed here, since `libnvinfer_plugin.so` does a `dlopen()` on the `libcuda.so` provided there.  `tensorrt-core` shouldn't be needed, since it will get pulled in due to linking against `libnvinfer.so`.

madisongh (changes requested):
Just a couple more minor things to check

recipes-devtools/gie/tensorrt-samples_8.0.1-1.bb
madisongh: I looked at the code for the samples, and while trtexec does make `dlopen` calls, it only does so for libraries you tell it to on the command line - nothing implicit in the code.

ichergui: both `RDEPENDS` were removed

external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins_8.0.1-1.bb
ichergui: Agree

recipes-devtools/gie/tensorrt-plugins-prebuilt_8.0.1-1.bb
ichergui: Agree

madisongh (approved):

PR #804, opened 2021-09-25
u-boot-bup-payload: add dtc-native into dependency list

ichergui:


ichergui:
The goal is to fix the issue detected in this [build](https://builder.madison.systems/#/builders/3/builds/138)

PR #865, opened 2021-12-09
bootlogo: use linear gamma to match nvdisp window settings in cboot

kekiefer:
The SOR settings in cboot apply gamma correction to the window
containing the bootlogo buffer. As such, either the window needs to
degamma the logo image, or the logo image must have a linear gamma.

This change does the latter, to keep limit the amount of patching
we're adding to cboot.

---

Note that this change only affects customizations to the bootlogo-custom recipe. The stock oe4t logo already cooks the bitmaps in a way that is compatible with cboot.


kekiefer:
FYI @Lexmark-chad

madisongh:
LGTM.  If you rebase on top of master latest, the pre-commit check should pass.

PR #866, opened 2021-12-11
Fixup for tensorrt python wrapper

timongentzsch:
In addition to Matt Madison patches I have modified the template function `getOverload` in `/python/include/utils.h` to accommodate the change in nvidia's tensorrt libraries.


madisongh:
Thanks for the submission.  Does this fix a build issue of some kind?  It would help to know more about what this is for.

Also, if this is a new patch, it would be easier to review if you just added it separately, instead of modifying the original.

Finally, please format the commit message and include a signed-off-by line, as described in our [contributing guidelines](https://github.com/OE4T/meta-tegra/blob/master/CONTRIBUTING.md).

Thanks!

timongentzsch:
With the old code I got a build error during compiling. After a bit of research I found out that Nvidia recently changed the name of the `get_overwrite` function to `get_overload` (compare commit: [python/include/utils.h](https://github.com/NVIDIA/TensorRT/commit/2d517d270e3697a5775e7861873dc21d5fba6bae#)).
At this point it probably makes sense to base this package on the commit mentioned. However, your patchfile is no longer compatible with this version as the layout of the CMakeLists file has changed too (which is why I came up with this messy solution for now 😉)

madisongh:
Hmm.  I'm not sure we can upgrade past the 8.0.1 tag, can we?  TensorRT in JetPack is still at 8.0.1.

timongentzsch:
Yeah i think youre right. As a kind of hotfix while we are waiting for Tensorrt release 8.2 on Tegra, I reverted the `.patch` to your original state and stitched in my modifications with a signed-off line commit.

madisongh:
I did some digging on this, since I don't see any issues on master with the existing recipe and patch.

I happened to stumble across a build error trying to build python3-tensorrt on the dunfell branch, however.  The root cause appears to be due to a breaking API change made in pybind11 2.6.  meta-python's dunfell branch provides pybind11 2.4.3 that provides `get_overload`; version 2.6 and later replaces that with `get_override`.

Since meta-python provides pybind11 2.6 in their honister branch and 2.7 in their master branch, this change isn't appropriate for our master branch.  If you could resubmit this PR as an added patch targeting dunfell instead of master, I'll merge it.

timongentzsch:
Yes this makes sense. I only tested it on `dunfell` branch and missed up by creating this PR

PR #867, opened 2021-12-13
Provide co-resident gcc 8 toolchain for CUDA

madisongh:
OE-Core now includes Rust support, and that toolchain requires gcc 11 in order to build its standard library. Because of that change, we need to change how we support building CUDA libraries and programs, since CUDA 10.2 is stuck at gcc 8.

This PR proposes that we bite the bullet and provide the gcc 8 toolchain directly in the layer.  The recipes are modified to allow this toolchain to co-reside with the default toolchain from the core, and are intended to be used only when CUDA compatibility is required. A new `cuda-gcc.bbclass` is provided; recipes that need to be built with the older toolchain can inherit that class (and `cuda.bbclass` is updated to do that).

This allows us to build everything else with the latest toolchain available.  We no longer need to force users to switch to the older toolchain at the distro level.


madisongh:
After further testing and examining how Debian handles this, it looks like we can use the main toolchain for everything in most cases, and just pass the old compiler to `nvcc`.  So `CXX` and `CC` will invoke the main compiler by default; if a particular package's build files make it difficult to mix the toolchains (as for the CUDA sample programs), you can just set `CC` to `${CC_FOR_CUDA}` and/or `CXX` to `${CXX_FOR_CUDA}` in the recipe.

I've also updated the gcc-8 recipes to version 8.5.0 to pick up all of the available fixes, including some CVE patches that had been back-ported to 8.3.0 in the original recipes.

ichergui:
> After further testing and examining how Debian handles this, it looks like we can use the main toolchain for everything in most cases, and just pass the old compiler to `nvcc`. So `CXX` and `CC` will invoke the main compiler by default; if a particular package's build files make it difficult to mix the toolchains (as for the CUDA sample programs), you can just set `CC` to `${CC_FOR_CUDA}` and/or `CXX` to `${CXX_FOR_CUDA}` in the recipe.
> 
> I've also updated the gcc-8 recipes to version 8.5.0 to pick up all of the available fixes, including some CVE patches that had been back-ported to 8.3.0 in the original recipes.

Thanks @madisongh for doing this.
Please let me know if I can help with the testing

madisongh:
Everything looks good with build, SDK, and runtime testing I've done.

PR #871, opened 2021-12-20
Hotfix to account for naming mismatch of get_override function

timongentzsch:
Signed-off-by: Timon Gentzsch <timongentzsch@gmail.com>


madisongh:
Thanks!

PR #887, opened 2022-01-25
[tegra-libraries-multimedia-v4l / cuda-cudart] Add missing .so to nvidia docker pass-thorugh list.

schneith:
The following libraries were not mounted into the docker containers:
- /usr/local/cuda-10.2/lib/libcudart.so.10.2.300
- /usr/lib/libv4l/plugins-wrapped/libv4l2_nvvidconv.so

which caused gstreamer to fail when loading plugins that depend on the libraries above.


madisongh:
This looks good overall, although I'm not sure the change to the cuda-shared-binaries include file is really needed.

Could you split this into two commits, to separate the tegra-libraries change from the cuda change?

ichergui:
Thanks @schneith for this PR.
Looks good to me.
Agree with @madisongh

schneith:
@madisongh @ichergui  thanks for the fast feedback. i have addressed your comments. could you please take another look?

PR #908, opened 2022-02-16
Fix wireless regulatory database for tegra platforms

ichergui:
* This PR was tested on `Jetson Xavier NX`

* Example (setting the regulatory domain to "Great Britain")
```
# iw reg set GB
```
* Result

```
root@jetson-xavier-nx-devkit-emmc:~# iw reg get 
global
country US: DFS-FCC
        (2400 - 2472 @ 40), (N/A, 30), (N/A)
        (5150 - 5250 @ 80), (N/A, 23), (N/A), AUTO-BW
        (5250 - 5350 @ 80), (N/A, 23), (0 ms), DFS, AUTO-BW
        (5470 - 5730 @ 160), (N/A, 23), (0 ms), DFS
        (5730 - 5850 @ 80), (N/A, 30), (N/A), AUTO-BW
        (5850 - 5895 @ 40), (N/A, 27), (N/A), NO-OUTDOOR, AUTO-BW, NO-IR
        (57240 - 71000 @ 2160), (N/A, 40), (N/A)

root@jetson-xavier-nx-devkit-emmc:~# iw reg set GB
root@jetson-xavier-nx-devkit-emmc:~# iw reg get 
global
country GB: DFS-ETSI
        (2400 - 2483 @ 40), (N/A, 20), (N/A)
        (5150 - 5250 @ 80), (N/A, 23), (N/A), NO-OUTDOOR, AUTO-BW
        (5250 - 5350 @ 80), (N/A, 20), (0 ms), NO-OUTDOOR, DFS, AUTO-BW
        (5470 - 5730 @ 160), (N/A, 26), (0 ms), DFS
        (5725 - 5850 @ 80), (N/A, 23), (N/A), NO-OUTDOOR
        (57000 - 71000 @ 2160), (N/A, 40), (N/A)

root@jetson-xavier-nx-devkit-emmc:~#
```


madisongh:
The regdb requires using crda, which in turn requires using meta-networking, and I'd rather not add that requirement.

I don't mind dropping `wireless-regdb-static`, but I'd rather substitute putting the database in the kernel with some changes to the kernel recipe.  The addition should be optional, with the default value being based on `MACHINE_FEATURES` containing `wifi`.  If on, we can add a config fragment to enable the internal regdb and then copy the `db.txt` file into the right place in the kernel tree (similar to what `kernel_wireless_regdb.bbclass` in meta-networking does).

This would give us a solution that doesn't require an extra layer by default, but allows folks to switch to using `crda` if they prefer.

ichergui:
> The regdb requires using crda, which in turn requires using meta-networking, and I'd rather not add that requirement.
> 
> I don't mind dropping `wireless-regdb-static`, but I'd rather substitute putting the database in the kernel with some changes to the kernel recipe. The addition should be optional, with the default value being based on `MACHINE_FEATURES` containing `wifi`. If on, we can add a config fragment to enable the internal regdb and then copy the `db.txt` file into the right place in the kernel tree (similar to what `kernel_wireless_regdb.bbclass` in meta-networking does).
> 
> This would give us a solution that doesn't require an extra layer by default, but allows folks to switch to using `crda` if they prefer.

Thanks @madisongh for your feedback.

There is two reasons that make me push this PR:
* Linux kernel version 
* `CONFIG_CFG80211_INTERNAL_REGDB` recommendation

Please take a look [here](http://linuxwireless.sipsolutions.net/en/developers/Regulatory/CRDA/__v45.html#CONFIG_CFG80211_INTERNAL_REGDB) 

I believe that the usage of `crda` makes users  life easy instead of dealing with Linux kernel (updates of  the regulatory domain).
Please let me know what do you think ?

ichergui:
* Tests result with `Jetson Xavier NX` WiFi `80211AC` `2.4GHZ`

```
root@jetson-xavier-nx-devkit-emmc:~# hostapd ./hostapd.conf 
wlan0: interface state UNINITIALIZED->COUNTRY_UPDATE
[  284.870155] IPv6: ADDRCONF(NETDEV_CHANGE): wlan0: link becomes ready
wlan0: interface state COUNTRY_UPDATE->ENABLED
wlan0: AP-ENABLED 
wlan0: STA 94:d0:0d:8f:4d:e8 IEEE 802.11: associated
wlan0: AP-STA-CONNECTED 94:d0:0d:8f:4d:e8
wlan0: STA 94:d0:0d:8f:4d:e8 RADIUS: starting accounting session 79F52E272099AEF9
wlan0: STA 94:d0:0d:8f:4d:e8 WPA: pairwise key handshake completed (RSN)
wlan0: EAPOL-4WAY-HS-COMPLETED 94:d0:0d:8f:4d:e8

```

recipes-kernel/linux/linux-tegra_4.9.bb
```diff
@@ -9,6 +9,7 @@ require recipes-kernel/linux/linux-yocto.inc
 
 DEPENDS:remove = "kern-tools-native"
 DEPENDS:append = " kern-tools-tegra-native"
+DEPENDS:append = " ${@bb.utils.contains('DISTRO_FEATURES', 'wifi', 'wireless-regdb-native', '', d)}"
```
madisongh, line 12: To make this easier for users to override, could you add a new variable to control this, with the default being based on `wifi` being in DISTRO_FEATURES? Something like
```
KERNEL_INTERNAL_WIRELESS_REGDB ?= "${@bb.utils.contains('DISTRO_FEATURES', 'wifi', '1', '0', '', d)}"
```
and then base all of the conditionals on that variable, as in
```
DEPENDS:append = "@' wireless-regdb-native' if bb.utils.to_boolean(d.getVar('KERNEL_INTERNAL_WIRELESS_REGDB')) else ''"
```
That will make it easier for a user to switch to using `crda` if they prefer.

recipes-kernel/linux/linux-tegra_4.9.bb
```diff
@@ -187,3 +189,12 @@ python () {
...
 }
+
+SRCTREECOVEREDTASKS += "do_kernel_add_regdb"
+do_kernel_add_regdb() {
+    if ${@bb.utils.contains('DISTRO_FEATURES', 'wifi', 'true', 'false', d)}; then
+        cp ${STAGING_LIBDIR_NATIVE}/crda/db.txt ${S}/net/wireless/db.txt
+    fi
+}
+do_kernel_add_regdb[dirs] = "${S}"
+addtask kernel_add_regdb before do_build after do_configure
```
madisongh, line 200: That `before do_build` should be `before do_compile`, since the file is needed during kernel compilation (the bbclass in meta-networking has this wrong, too).

madisongh (changes requested):
A couple of changes needed, otherwise looks good.

If you rebase your branch on top of master latest, the automated checks should pass.

recipes-kernel/linux/linux-tegra_4.9.bb
ichergui: Agree 100%

ichergui: Will push the changes

recipes-kernel/linux/linux-tegra_4.9.bb
ichergui: Will update the recipe

madisongh (approved):

PR #909, opened 2022-02-16
tegra-configs-alsa: install ALSA config files for tegra210

brgl:
HDMI audio doesn't work with pulseaudio on the Jetson Nano devkit. This adds the missing ALSA configs.


madisongh:
Thanks!

PR #930, opened 2022-03-12
Add support for Weston 10

### LINKED ISSUE #897: Weston is broken on master
**Describe the bug**
OE-Core master has upgraded to Weston 10.0.0, and integration on the Jetson platforms has not yet been completed.
The EGLStreams patches we had been using with Weston 9.0.0 have not been ported to the new version.

**To Reproduce**
Build `demo-image-weston` on the demo distro master branch.
Build succeeds, but `weston.service` fails to start.

**Additional context**
Possible options:
1. Re-port the EGLStreams patches. This may or may not be possible.
2. Try converting to using the GBM back-end and adapt the [EGL backend](https://github.com/NVIDIA/egl-gbm) for `libgbm`.  This may or may not work.
3. Pull the Weston 9.0.0 recipe into the layer and have the tegra platforms prefer that version, instead of upgrading.

kekiefer:
This pull request completes basic support for Weston 10 in OE4T.

It adds a new couple new packages, including a new `tegra-udrm-gbm` library, which is pulled in by Mesa by relatively recent (NVIDIA-contributed) GBM backend loader. This library wraps the libnvgbm.so in a way that can be used by mesa and other components that expect the Mesa ABI for libgbm.

The other key piece is a port of the Weston 9 patches to handle eglstreams surfaces from clients.

More eyes on this will be appreciated, so I figured it's better to get this in sooner than later. The core Weston EGL and SHM clients have been tested and are fully functional. EGL-DRM clients do not work, but they did not before either.

Fixes #897


recipes-bsp/tegra-binaries/tegra-libraries-gbm_32.6.1.bb
```diff
@@ -17,6 +17,13 @@ TEGRA_LIBRARIES_TO_INSTALL = "\
     tegra/libnvgbm.so \
 "
 
+do_install:append() {
+    patchelf --set-soname libnvgbm.so.1 ${D}${libdir}/libnvgbm.so
+}
+
+EXCLUDE_FROM_SHLIBS = "1"
+INSANE_SKIP:${PN} += "file-rdeps"
```
madisongh, line 25: Since you're setting the SONAME, you shouldn't need these settings - with the unique SONAME, the shlibs processing should "just work".  Worked OK for me in a test buid.

recipes-graphics/mesa/mesa_%.bbappend
```diff
@@ -2,6 +2,8 @@ EXTRA_OEMESON:append:tegra = " -Dglvnd=true"
 DEPENDS:append:tegra = " libglvnd"
 PROVIDES:tegra = "virtual/mesa virtual/libgbm"
 
+PACKAGECONFIG:append:tegra = "gbm"
+
```
madisongh, line 6: `gbm` is in the PACKAGECONFIG by default, so this shouldn't be needed.  If it _is_ needed in some cases, make sure you prefix it with a blank

recipes-graphics/mesa/tegra-udrm-gbm_1.0.0.bb
```diff
@@ -0,0 +1,18 @@
+DESCRIPTION = "Tegra GBM backend for mesa"
+HOMEPAGE = "https://github.com/kekiefer/tegra-udrm-gbm"
```
madisongh, line 2: Maybe update this now that you've transferred the repo over to OE4T?

recipes-graphics/wayland/weston_%.bbappend
```diff
@@ -0,0 +1,6 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+SRC_URI += "file://0001-gl-renderer-Add-EGL-client-support-for-EGLStream-fra.patch \
+            file://0002-compositor-Process-stream-attach-requests-with-wl_eg.patch \
+            "
+
```
madisongh, line 6: Could you move these patch additions into the `weston_10.0.0.bbappend`?  I expect this got created by devtool, which doesn't handle modifying existing bbappends when they've got version numbers in them.

recipes-graphics/mesa/egl-gbm_1.1.0.bb
```diff
@@ -0,0 +1,30 @@
...
+
+inherit meson pkgconfig features_check
+
+do_install:append() {
+    install -d ${D}${datadir}/egl/egl_external_platform.d
+    install -m 0644 ${WORKDIR}/nvidia_gbm.json ${D}${datadir}/egl/egl_external_platform.d/15_nvidia_gbm.json
+}
+
+FILES:${PN} += "${datadir}/egl"
+INSANE_SKIP:${PN} = "dev-so"
```
madisongh, line 30: Do we need an `RDEPENDS` on `tegra-udrm-gbm` in this recipe, to make sure your wrapper gets pulled in?  This might not be the right place for it, but we do need something to pull in that package so users don't have to do it themselves.

madisongh (changes requested):
This is really great!  I've got some updates to make to master to deal with the latest OE-Core upstream changes, which I'll push shortly.  Those updates combined with this did get weston working again.

I'd appreciate it if you could squash the commits together so that there's one per recipe, and I have a few inline comments, but otherwise this looks good.

I did run into an issue with weston dying with a SIGABRT (and no explanatory message) the first time I booted after flashing on my Xavier NX.  Worked OK after rebooting, though.  Not sure what that was.

From my tests, basic clients (e.g., wayland-simple-egl and friends) look like they're working, and gstreamer 1.20's `waylandsink` also works (not 1.14's though).  The `nveglglessink winsys=wayland` fails to find the compositor and shell - maybe a protocol version mismatch there?

I didn't test xwayland, but we probably should, as well as testing non-weston (i.e., sato) builds to make sure nothing's broken there.

Thanks!

madisongh:
I've pushed the changes to master so it passes the layer check again, so if you rebase the automated check should pass.

kekiefer:
All concerns addressed, thanks for the feedback!

Regarding the broken `nveglglessink winsys=wayland`, Weston 10.0.0 deprecated wlshell, and it's disabled by default. Given that it is used by our gstreamer plugins still, it probably makes sense to continue pulling it in. I'll add this to the change set.

recipes-graphics/mesa/egl-gbm_1.1.0.bb
kekiefer: I've added this as an RDEPENDS on mesa's libgbm, because that's the component that really needs the backend support.

madisongh (approved):

PR #938, opened 2022-03-16
weston: fix dmabuf synchronization for tegra

kekiefer:
Tegra's egl fence fds are compatible only with the kernel's older
android sync fence api. This patch update's weston's implementation
of zwp_linux_explicit_synchronization_v1 to use the android api
on tegra devices.

Signed-off-by: Kurt Kiefer <kekiefer@gmail.com>


madisongh:
Yikes, this is a bit scary.  We have both sync implementations in our kernel, mutually incompatible, and no way for userland code to tell which implementation it should be using.  I wonder if maybe we should patch the kernel to disable the `drivers/dma-buf/sync_file.c` ioctl implementation, since it looks like all the NV code still uses the Android one?

PR #940, opened 2022-03-18
Add support for Clara AGX Developer Kit

ibstewart:
This adds support for the [Clara AGX Developer Kit](https://www.nvidia.com/en-us/clara/developer-kits/)

Note that this only enables the integrated Tegra GPU; it does not enable the RTX 6000 that is included with the Clara AGX Devkit.


ichergui (changes requested):
Hey @ibstewart 
Could you please change the name of the machine  to keep it aligned with `Jetpack` / `L4T` ?
Please use `clara-agx-xavier-devkit.conf` instead of `clara-agx-devkit.conf`

ichergui (approved):

ichergui:
Thanks @ibstewart

ibstewart:
Thanks for the reviews @ichergui!

PR #941, opened 2022-03-18
conf/machine: Use soft defaults for some machine variables

ibstewart:
This allows new meta layers to specify their own values for these variables without having them overridden by the meta-tegra inclusion.

Specifically, I am working on a new meta layer that enables the RTX 6000 discrete GPU (dGPU) that is included with the Clara AGX Devkit, but using the dGPU requires the integrated Tegra GPU (iGPU) be disabled.

I tried using a combination of `:remove` and `:append` lines to remove the iGPU components and then add the dGPU components, which works for most variables since this override style is resolved at variable expansion rather than being applied immediately, but it did not work for the xserver PREFERRED_PROVIDER variables since the remove/append always left trailing whitespace in the variable which is not handled properly.


ichergui (approved):

ichergui:
Thanks @ibstewart for this PR

PR #959, opened 2022-04-14
external/qt5-layer: fix egldevice backend for L4T 32.7.1

### LINKED ISSUE #956: Blank screen with Qt5 (EGLFS + KMS/DRM) since 32.7.1
Since L4T32.7.1 (dunfell branch) my Qt application does not render any more. 
The application is starting and interaction with mouse/keyboard/touch is working but the screen is black. No issues on previous L4T32.6.1 branch. I do not use X11 or wayland, both are disabled.

Qt installation from [meta-qt5](https://github.com/meta-qt5/meta-qt5) (tested both, dunfell & master)
`PACKAGECONFIG_append_qtbase = " eglfs kms gles2 "`

runtime environment variables:
`QT_QPA_PLATFORM=eglfs`
`QT_QPA_EGLFS_INTEGRATION=eglfs_kms_egldevice`

Qt sample application:
[https://github.com/alpqr/quickmwtest](https://github.com/alpqr/quickmwtest)

kekiefer:
This change adds two patches to support tegra devices in qt5 with
the L4T release 32.7.1. The newer version of L4T appears to unify
the capabilities of tegra-udrm and drm-nvdc drivers in the NVIDIA's
libdrm, and this enables us to more easily support Qt using either
method, but does require some fixups.

The first patch is a backport from newer versions of Qt which will
allow the egldevice backend to function with the tegra-udrm. It is
not strictly necessary, but is nice because then users don't have
to worry about removing this driver if they want to e.g. switch
between Weston and Qt.

The second patch adds a default framebuffer to the initial calls
to libdrm to set modes and planes. The default value of -1 that was
used by Qt before wouldn't have been valid for any other platform,
and now it is not valid for tegra either.

Fixes #956

### DIFF
--- a/external/qt5-layer/recipes-qt/qt5/qtbase/0001-eglfs-Newer-Nvidia-libdrm-provide-device-instead-dri.patch
+++ b/external/qt5-layer/recipes-qt/qt5/qtbase/0001-eglfs-Newer-Nvidia-libdrm-provide-device-instead-dri.patch
@@ -1,0 +1,37 @@
+From e28a134e847eff8bc6a27ca7b0bdde424716497f Mon Sep 17 00:00:00 2001
+From: =?UTF-8?q?Pasi=20Pet=C3=A4j=C3=A4j=C3=A4rvi?= <pasi.petajajarvi@qt.io>
+Date: Tue, 30 Mar 2021 17:40:03 +0300
+Subject: [PATCH] eglfs: Newer Nvidia libdrm provide device instead driver
+ module name
+
+In newer Nvidia proprietary libdrm binaries display device name is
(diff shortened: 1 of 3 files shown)


kekiefer:
It's probably worth mentioning that I experimented with enabling the gbm integration for Qt as well. With the work we did for Weston, this was very close to working. Surfaces were sent to the display, but with the wrong tiling modifiers, as Qt wasn't using the version of addfb2 that includes passing along the modifiers the way Weston did. If we want to, this could probably be addressed (maybe with a Qt patch, maybe with a patch to our libgbm backend), but I chose to go down the path of fixing egldevice, since this fix wants backporting to dunfell as well.

PR #962, opened 2022-04-16
libnvidia-container-tools: bump version to 0.11.0

ichergui:


madisongh:
Backing off the SRCREV wasn't quite the right approach.  They've redesigned the jetson-specific library to be co-resident with the latest (1.9.0) library and get `dlopen`ed by the CLI tool.

I've manually reworked what you had here for 0.11, did a bit of cleanup, added the recipe for 1.9.0, and also updated the rest of the nvidia-container recipes (nvidia-docker and nvidia-container-toolkit, dropping the now-obsolete nvidia-container-runtime separate recipe) to latest, to match what's in JetPack 5.

The only missing piece is to figure out how to deal with the change to not passthrough-mounting all of the Jetson-specific libraries. That's on by default now, and none of the Jetson-specific libraries are getting passed through because we don't list them in the `l4t.csv` file.  I'm not sure yet what the right way is to fix that.

PR #970, opened 2022-04-27
external/virtualization-layer: fix libnvidia-container-tools git issue

selyunin:
The latest git release adds ownership checks on the git tree when
a git command is invoked from pseudo context, so when the
makefiles are processed during the do_install task, an error is
reported.

Fix this by adding a REVISION variable setting to EXTRA_OEMAKE, and
also include a COMPILER setting to override shell callout to
extract the compiler path.

The following changes were already done to dunfell, and other
dunfell-* branches.

The original fix created by Matt Madison <matt@madison.systems>

Signed-off-by: Konstantin Selyunin <selyunin.k.v.@gmail.com>

### DIFF
--- a/external/virtualization-layer/recipes-containers/libnvidia-container-tools/libnvidia-container-tools_0.9.0.bb
+++ b/external/virtualization-layer/recipes-containers/libnvidia-container-tools/libnvidia-container-tools_0.9.0.bb
@@ -62,7 +62,7 @@ def build_date(d):
 
 # We need to link with libelf, otherwise we need to
 # include bmake-native which does not exist at the moment.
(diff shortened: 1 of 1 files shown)


selyunin:
We use the `meta-tegra` repo, the `dunfell-l4t-r32.4.3` branch, and recently encountered problem with `git` -- unsafe repository:

`fatal: unsafe repository ('../build/tmp/work/aarch64-poky-linux/libnvidia-container-tools/0.9.0-r1/git' is owned by someone else)`

@madisongh already created a fixed for this, which is available in other branches, but not in `dunfell-l4t-r32.4.3`. 
The change is: https://github.com/OE4T/meta-tegra/commit/c611c7c0b74594201c3061b69be1c7e339dd3260

The `libnvidia-container-tools_0.10.0.bb` is of different version in `dunfell` and in `dunfell-l4t-r32.4.3`, `git cherry-pick` did not work.

PR #978, opened 2022-05-12
change: use initramfs link name for tegra_kernel_image

kvberge:
This change leverages the INITRAMFS_LINK_NAME to 'dynamically' build
the image name returned from tegra_kernel_image method during tegra_flash
building.

This helps avoid build failures if a custom INITRAMFS_LINK_NAME is used
to generate a bundled kernel and ramdisk image.

By default, INITRAMFS_LINK_NAME expands to
```
INITRAMFS_LINK_NAME ?= "initramfs-${KERNEL_ARTIFACT_LINK_NAME}"
```
which would yield:
```
KERNEL_ARTIFACT_LINK_NAME ?= "${MACHINE}"
```
resulting in no change to default behavior.

Signed-off-by: Kevin Vanden Berge <kvandenberge@gmail.com>


madisongh:
Thanks!  I've found a couple of other similar references elsewhere and will change them as well.

madisongh:
Minor changes like this are fine to back-port.  I've done that for this and pushed an updated dunfell branch.

PR #984, opened 2022-05-18
Fix APT upgrade conflicts with the tegra-libraries

MikeLoder:
The tegra-libraries we all split in to their own recipes however they
conflict with the old tegra-libraries packages when upgrading. This
marks the previous package dependencies as conflicts so they are removed
on an upgrade.

Signed-off-by: Mike Loder <mloder@miovision.com>

_NOTE: I only have a PR for the Dunfell branch as I don't have a setup for the new master yocto with the updated syntax._

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-libraries-camera_32.7.1.bb
+++ b/recipes-bsp/tegra-binaries/tegra-libraries-camera_32.7.1.bb
@@ -71,3 +71,8 @@ FILES_${PN}-nvtunerd = "${sbindir}/nvtunerd"
 RDEPENDS_${PN} = "tegra-argus-daemon"
 
 CONTAINER_CSV_FILES_append = "${@bb.utils.contains('PACKAGECONFIG', 'x11', ' ${libdir}/libv4l/plugins/*', '', d)}"
+
+# This package includes files that were moved from tegra-libraries, tegra-libraries-argus, tegra-libraries-libv4l-plugins
+#  so we need to mark them as conflicts to have a successful apt upgrade.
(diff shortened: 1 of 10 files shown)


ichergui (approved):
LGTM

Maybe you could use `master` branch instead of `dunfell` 

Let's see what @madisongh thinks.

madisongh (commented):
I don't see `RBREAKS_${PN}` being used anywhere in the packaging code.  Are those lines really needed?

MikeLoder:
Good catch, we have RBREAKS_ in a number of our recipes but it does not appear to be used by anything. Not sure where that came from, i will remove it.

On a side note I only submitted a Dunfell PR as master is a different syntax and I don't have a setup to test it.

MikeLoder:
I have removed the RBREAKS_ fields from the recipes.

madisongh:
I've had to revert this, as the combination of the RREPLACES  and RCONFLICTS in this change is causing ordinary builds to fail when using RPM packaging (possibly others as well).

In OE-Core's `package_rpm.bbclass`, RREPLACES is translated into a combination of `Obsoletes:` and `Provides:` in the RPM package headers.  So the packages replacing the old `tegra-libraries` package are also now providing that package, so all of the packages that use RCONFLICTS to conflict with `tegra-libraries` are refusing to install into an RPM-based rootfs.

I don't know if what `package_rpm.bbclass` is doing is wrong, or if there is some other combination of settings that would allow for runtime deb package feed handling to correctly upgrade from the old (pre-32.6) `tegra-libraries` package to the new split packages without breaking normal rootfs builds. So for now, I'm just reverting this change.

MikeLoder:
Hum that's disappointing. Thanks for letting me know. I can keep these in our appends layer.

PR #986, opened 2022-05-20
Remove tegra libraries recipe - Dunfell

MikeLoder:
This recipe is no longer needed in the 32.7.1 setup with all the tegra-libaries having split-out recipes.


madisongh (changes requested):
Thanks for updating the siggen exclusions with this, too. 

Just add a signed-off-by to the commit message and this should be good to go.

MikeLoder:
Added a signed off by, should be good to go

madisongh (approved):

PR #994, opened 2022-06-09
Fix QA issue

ichergui:


madisongh (commented):
I'm not sure about this.  Let me think about it.

madisongh:
> This fixes my issues with deepstream-6.1

I understand, but the problem is really with how those DS libraries are linked.  Adding `file-rdeps` to INSANE_SKIP should also work.

Have you tried using the other intermediate packaging formats (`deb`, `ipk`)?  Does this RPROVIDES hackery work with them as well?

ichergui:
> RPROVIDES

Adding the `file-rdeps` doesn't resolve the issue.
I didn't try other intermediate packaging formats.

ichergui:
Close this PR because we have another solution

PR #996, opened 2022-06-10
Remove "blacklist mttcan" from linux-tegra.bb

homalozoa:
The module "mttcan" is disabled by default in linux-tegra.bb right now. I don't think it's a good idea for other kernels that derived from this bb file. If I want to let my kernel auto load `mttcan`. I must create something to remove it from blacklist. So I make this PR to enable mttcan by default. It will make this project more compatible.

### DIFF
--- a/recipes-kernel/linux/linux-tegra_5.10.bb
+++ b/recipes-kernel/linux/linux-tegra_5.10.bb
@@ -176,12 +176,6 @@ COMPATIBLE_MACHINE = "(tegra)"
 
 RDEPENDS:${KERNEL_PACKAGE_NAME}-base = ""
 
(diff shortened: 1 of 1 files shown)


ichergui:
Hey @homalozoa

madisongh:
Hmm. It looks like NVIDIA has disable autoloading of `mttcan`  by default in most earlier L4T releases, too, but I never noticed it before, so we don't have these settings in the earlier kernel recipes. I don't have a problem with dropping the settings here, too, as long as it doesn't break anything. (And from the dev forum posts I'm seeing the L4T/JetPack default has caused confusion for folks wanting to use CANbus, anyway.)

PR #1000, opened 2022-06-15
linux-tegra: update packaging RDEPENDS->RRECOMMENDS

### LINKED ISSUE #999: kernel-image-* erroneously included in rootfs (kirkstone and later)
As of commit https://github.com/openembedded/openembedded-core/commit/1c90b27d2c65cfb4f9debf0272820b6a95942f76 the kernel images are included in the rootfs, as the meta-tegra layer needs to be updated to either:

a) replace the linux-tegra recipe's existing fixup to `RDEPENDS` with a fixup to `RRECOMMENDS` instead
b) remove the fixup and add a `PACKAGE_EXCLUDE:append` to the tegra-common.inc

I've tested both approaches, and both will work. So this issue is to solicit feedback regarding which path makes more sense.

kekiefer:
Upstream kernel.bbclass changed the inclusion of kernel image
artifacts from RDEPENDS to RRECOMMENDS.

Update this recipe accordingly.

Fixes #999

### DIFF
--- a/recipes-kernel/linux/linux-tegra_4.9.bb
+++ b/recipes-kernel/linux/linux-tegra_4.9.bb
@@ -174,7 +174,7 @@ do_deploy[depends] += "${EXTRADEPLOYDEPS}"
 
 COMPATIBLE_MACHINE = "(tegra)"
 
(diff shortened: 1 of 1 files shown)


madisongh:
I've cherry-picked this over to kirkstone as well.

PR #1005, opened 2022-06-24
tegra-configs-alsa: enable default ALSA config overrides

EasyIP2023:
The default asound.conf in L4T is only configured on the
Jetson Xavier NX for (card: tegra-hda-xnx, device id: 7).
Which is HDMI Audio Out. One can't hear audio over DP
unless the alsa config is updated to
(card: tegra-hda-xnx, device id: 3).

Commit:

-  Allows for default L4T provided alsa config to
   be overridden
- Adds in a custom alsa config that configures
  audio over both DP and HDMI. Only for machines
  jetson-xavier-nx-devkit-{emmc}

Signed-off-by: Vincent Davis Jr <vince@underview.tech>

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-configs-alsa_32.7.2.bb
+++ b/recipes-bsp/tegra-binaries/tegra-configs-alsa_32.7.2.bb
@@ -9,13 +9,14 @@ require tegra-debian-libraries-common.inc
 MAINSUM = "44fa43d9646d66b5c1bb28def9736c3d8cf7a3e21082fbf9017c63246e2b0ae1"
 MAINSUM:tegra210 = "a21c44851b287f0fde245f753fa7f591a2d84f125b111bbf54ae34d7c0f3b255"
 
-do_install() {
-    if [ -n "${TEGRA_AUDIO_DEVICE}" ]; then
-        install -d ${D}${sysconfdir}
-        install -m 0644 ${S}/etc/asound.conf.${TEGRA_AUDIO_DEVICE} ${D}${sysconfdir}/asound.conf
-    fi
+TEGRA_AUDIO_CONFIG ?= "${S}/etc/asound.conf.${TEGRA_AUDIO_DEVICE}"
 
-    install -d ${D}${datadir}/alsa/cards
+do_install() {
+    install -d ${D}${sysconfdir} ${D}${datadir}/alsa/cards
+
+    if [ -n "${TEGRA_AUDIO_CONFIG}" ]; then
+        install -m 0644 ${TEGRA_AUDIO_CONFIG} ${D}${sysconfdir}/asound.conf
+    fi
 }
 
 do_install:append:tegra186() {


recipes-bsp/tegra-binaries/tegra-configs-alsa_32.7.2.bb
```diff
@@ -9,12 +9,24 @@ require tegra-debian-libraries-common.inc
 MAINSUM = "44fa43d9646d66b5c1bb28def9736c3d8cf7a3e21082fbf9017c63246e2b0ae1"
 MAINSUM:tegra210 = "a21c44851b287f0fde245f753fa7f591a2d84f125b111bbf54ae34d7c0f3b255"
 
+FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
```
ichergui, line 12: I think that you don't need this line

ichergui:
@EasyIP2023 Thanks for this PR
I will try this later and will keep you updated :slightly_smiling_face:

EasyIP2023:
@ichergui Thanks! Changes made.

ichergui:
@EasyIP2023 @madisongh I tested this PR with my `Jetson Xavier NX` and it works fine:
Here the commands line I used for my test:
* Test with HDMI 
```
$ speaker-test -c 2
```
* Test with DP
```
$ speaker-test -D hdmi:CARD=tegrahdaxnx,DEV=0 -c 2 -r 48000 -F S16_LE
```

ichergui (approved):
LGTM

ichergui:
> Having two variables `TEGRA_AUDIO_DEVICE` and `TEGRA_AUDIO_CONFIG` that enable the same thing but don't really play together is confusing here. I do like that the changes enable appending layers to more easily add asound.conf files. What if this were done using only `TEGRA_AUDIO_DEVICE` variable, by adding a step around `do_patch` to stuff any `asound.conf.*` files found in `FILES` to `${S}/etc/asound.conf.*`, letting `TEGRA_AUDIO_DEVICE` continue to do its thing as before. Then appending layers could still append `FILES` and just update `TEGRA_AUDIO_DEVICE`.

Thanks @kekiefer for this 
I agree with you. Do you think we should move this PR to `meta-tegra-community` and rework the logic. ?

kekiefer:
`TEGRA_AUDIO_DEVICE` could stay working as it does today. The recipe should just add an accommodation for installing additional `asound.conf.*` files from the SRC_URI collection into `${S}/etc/`.

madisongh:
I'm OK with having two variables... the TEGRA_AUDIO_DEVICE variable was just there for choosing one of the configs out of the L4T kit, so it's insufficient for generalizing to custom config files provided elsewhere.

Maybe have something like this in the recipe:
```
TEGRA_AUDIO_CONFIG ?= "${S}/etc/asound.conf.${TEGRA_AUDIO_DEVICE}"
```
and change the `install` command to reference that variable.

Then a bbappend to customize it for a particular configuration would just need something like:
```
FILESEXTRAPATHS:prepend := "${THISDIR}/files:"
SRC_URI += "file://my-custom-asound.conf"
TEGRA_AUDIO_CONFIG = "${WORKDIR}/my-custom-asound.conf"
```

I'm less sure about actually including the HDMI+DP config file directly in the layer, though.  It's specific to just one of the Jetsons, and I'm worried about it not getting regular testing, and thus breaking after an L4T update at some point.

EasyIP2023:
@madisongh @ichergui @kekiefer Okay made change, will add custom asound config to meta-tegra-community

kekiefer (approved):
LGTM

PR #1008, opened 2022-06-30
Getting Weston working on both  tegra234 and tegra194

ichergui:
* tests were done on `Jetson Xavier AGX` and `Jetson Orin AGX` successfully 
* running wayland client `weston-simple-egl` and `weston-simple-shm` successfully


recipes-graphics/mesa/tegra-udrm-gbm_1.0.2.bb
```diff
@@ -14,6 +14,8 @@ SRC_URI = "git://${SRC_REPO};branch=${SRCBRANCH}"
 # v1.0.2 tag
 SRCREV = "103d33d07e6c38106bf952403b336dc3ca809a08"
 
+SRC_URI:append:tegra234 = " file://0001-update-backend-and-library-name.patch"
```
kekiefer, line 17: I would like to know if there's a clean way to update the original library to support both tegra-udrm and nvidia-drm, rather than patching this

recipes-graphics/wayland/weston_10.0.0.bbappend
```diff
@@ -9,7 +9,9 @@ SRC_URI:append:tegra = " file://0001-Drop-DRM-version-check-in-meson.build.patch
 DEPENDS:append:tegra = " egl-wayland"
 PACKAGE_ARCH:tegra = "${TEGRA_PKGARCH}"
 
-RDEPENDS:${PN}:append:tegra = " egl-wayland egl-gbm tegra-udrm-probeconf"
+RDEPENDS:${PN}:append:tegra = " egl-wayland egl-gbm"
+RDEPENDS:${PN}:append:tegra194 = " tegra-udrm-probeconf"
+RDEPENDS:${PN}:append:tegra234 = " nvidia-drm-probeconf"
```
kekiefer, line 14: Is there a reason to keep the pattern of using a recipe to implement this? I think the tegra-udrm-probeconf might just be historical, because [`KERNEL_MODULE_AUTOLOAD`](https://docs.yoctoproject.org/dev/ref-manual/variables.html?#term-KERNEL_MODULE_AUTOLOAD) and [`KERNEL_MODULE_PROBECONF`](https://docs.yoctoproject.org/dev/ref-manual/variables.html?#term-KERNEL_MODULE_PROBECONF) are the current standard ways to accomplish this. Using the variables, we should just think if there is any case where we shouldn't be autoloading tegra-udrm/nvidia-drm, and overriding the assignment based on distro features? Certainly, if it's included in the image and loaded, it's worthless without the probeconf (`modeset=1`).

conf/machine/include/tegra194.inc
```diff
@@ -39,3 +39,5 @@ L4T_DEB_SOCNAME = "t194"
 NVPMODEL ?= "nvpmodel_t194"
 NVPOWER ?= "jetsonpower_t194"
 NVFANCONTROL ?= "nvfancontrol_p2888"
+
+MACHINE_EXTRA_RRECOMMENDS += "kernel-module-tegra-udrm"
```
kekiefer, line 43: should tegra234 get a `kernel-module-nvidia-drm` like this, or is that what is not possible because of the `KERNEL_SPLIT_MODULES=0` on the `nvidia-display-driver`?

kekiefer (changes requested):

conf/machine/include/tegra194.inc
ichergui: Yes, That's true
For now, we have `MACHINE_EXTRA_RRECOMMENDS += "nvidia-display-driver"` in the `tegra194.inc`
I will rework it at some point to have the kernel modules separate something like: 
```
kernel-module-nvidia
kernel-module-nvidia-drm
kernel-module-nvidia-modeset
```

recipes-graphics/wayland/weston_10.0.0.bbappend
ichergui: I'm not sure about the history, I agree with you, we should use `KERNEL_MODULE_AUTOLOAD` and `KERNEL_MODULE_PROBECONF`

recipes-graphics/mesa/tegra-udrm-gbm_1.0.2.bb
ichergui: I'm trying another alternative. I will push updates once I'm done with the testing

recipes-graphics/wayland/weston_10.0.0.bbappend
madisongh: I had originally enabled tegra-udrm that way in f860c21c902617cf09e67269acb10f22d55de370, but there were conflicts between it and non-Wayland/Weston DRM usage (for Qt, maybe?), so I switched to using the separate recipe that was only pulled in for W/W builds.

Assuming that conflict is gone now, we could switch back again and eliminate the extra recipe.

ichergui: Super @madisongh thanks for the explanation.
I will rework that and remove the additional recipes

kekiefer: At this point Qt is fine with the tegra-udrm.

recipes-graphics/mesa/tegra-udrm-gbm_1.0.2.bb
ichergui: I tried with a symlink creation and it works fine. We don't need the patch

conf/machine/include/tegra194.inc
```diff
@@ -39,3 +39,8 @@ L4T_DEB_SOCNAME = "t194"
...
+
+KERNEL_MODULE_PROBECONF += "nvidia-drm"
+module_conf_nvidia-drm = "options tegra-udrm modeset=1"
```
kekiefer, line 46: Should these be `tegra-udrm` for t194?

kekiefer (commented):
Just the above, for consistency. Otherwise LGTM

conf/machine/include/tegra194.inc
ichergui: Ah you are right ? 
I will fix it now 
Thanks  @kekiefer

kekiefer (approved):

recipes-kernel/nvidia-display-driver/nvidia-display-driver_34.1.1.bb
```diff
@@ -16,7 +16,7 @@ S = "${WORKDIR}/NVIDIA-kernel-module-source-TempVersion"
 B = "${S}"
 
 MODULES_MODULE_SYMVERS_LOCATION = "kernel-open"
-KERNEL_SPLIT_MODULES = "0"
+INHIBIT_PACKAGE_DEBUG_SPLIT = "1"
```
madisongh, line 19: It wouldn't hurt to add a comment here to explain why this is being set.

recipes-graphics/mesa/tegra-udrm-gbm_1.0.2.bb
```diff
@@ -18,4 +18,9 @@ S = "${WORKDIR}/git"
 
 inherit meson pkgconfig
 
+do_install:append:tegra234() {
```
madisongh, line 21: If you make this SoC-specific, we'd have to set the PACKAGE_ARCH to make the package SoC-specific, too.

I don't think it will hurt to add this symlink unconditionally, so just drop the `:tegra234` here.

recipes-graphics/wayland/weston_10.0.0.bbappend
```diff
@@ -9,7 +9,10 @@ SRC_URI:append:tegra = " file://0001-Drop-DRM-version-check-in-meson.build.patch
...
 PACKAGE_ARCH:tegra = "${TEGRA_PKGARCH}"
 
-RDEPENDS:${PN}:append:tegra = " egl-wayland egl-gbm tegra-udrm-probeconf"
+RDEPENDS:${PN}:append:tegra = " egl-wayland egl-gbm"
 
 # wlshell is used by nvidia gstreamer plugins
 EXTRA_OEMESON:append:tegra = " -Ddeprecated-wl-shell=true"
+
+RRECOMMENDS:${PN}:append:tegra194 = " kernel-module-tegra-udrm"
+RRECOMMENDS:${PN}:append:tegra234 = " kernel-module-nvidia-drm"
```
madisongh, line 18: These SoC-specific appends would mean having to set PACKAGE_ARCH to make this package SoC-specific, too.

I *think* it's safe to just use `:tegra` here and mention both of the kernel modules, since RRECOMMENDS is a weak dependency.

madisongh (changes requested):
Generally looks fine.  See in-line for some small changes.

recipes-kernel/nvidia-display-driver/nvidia-display-driver_34.1.1.bb
ichergui: Agree, will add a comment there.

recipes-graphics/mesa/tegra-udrm-gbm_1.0.2.bb
ichergui: I will add `PACKAGE_ARCH`.
FYI, `nvidia-drm` kernel module exists only for `tegra234`

recipes-graphics/wayland/weston_10.0.0.bbappend
ichergui: I will add `PACKAGE_ARCH` 
regarding the `tegra194` vs `tegra234`, I wanted to be explicit on the `RRECOMMENDS` 
I will change it to tegra and add both kernel modules in the same list

recipes-kernel/nvidia-display-driver/nvidia-display-driver_34.1.1.bb
ichergui: @madisongh 
Just to double check with you. Should I add `PACKAGE_ARCH` in this recipe ?

recipes-graphics/wayland/weston_10.0.0.bbappend
madisongh: If you do what I suggest, you **do not** need to set PACKAGE_ARCH.  It's one or the other.

recipes-graphics/mesa/tegra-udrm-gbm_1.0.2.bb
madisongh: I'd prefer it the other way.  Don't set PACKAGE_ARCH, and just create the symlink for all tegras.  The fact that only one of the two kernel modules would be active on a particular machine doesn't matter.

recipes-kernel/nvidia-display-driver/nvidia-display-driver_34.1.1.bb
madisongh: Oh, right, I missed that.  It's a kernel driver, so you should set `PACKAGE_ARCH = "${MACHINE_ARCH}"` in this recipe.

ichergui: I added it

madisongh (approved):

PR #1014, opened 2022-07-18
weston: rename bbappend to apply on new version from oe-core

sung-gon-kim:
This patch fixes the build error caused by the recent weston version change from oe-core.

https://git.yoctoproject.org/poky/commit/?h=kirkstone&id=742176bf5d26ff9587437dd04cb11570271fa23a


madisongh:
Thanks!

PR #1017, opened 2022-07-26
tegra-firmware: Create a symlink for the acr_ucode.bin firmware

pticon:
Using the Jetson Xavier NX, the firmware used for the gv11b is not loaded because it tries to load the file acr_ucode.bin and the only two files available are acr_ucode_dbg.bin and acr_ucode_prod.bin.

Here is the log of the error:
`[   23.011377] gk20a 17000000.gv11b: Direct firmware load for gv11b/acr_ucode.bin failed with error -2`

This patch creates a symlink from acr_ucode.bin to acr_ucode_prod.bin.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-firmware_32.7.2.bb
+++ b/recipes-bsp/tegra-binaries/tegra-firmware_32.7.2.bb
@@ -29,6 +29,11 @@ do_install:append:tegra210() {
     done
 }
 
+do_install::append:tegra194() {
+    cd ${D}/lib/firmware/gv11b
+    ln -s acr_ucode_prod.bin acr_ucode.bin
+}
+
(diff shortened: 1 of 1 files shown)


madisongh:
I'm not seeing that error on my kirkstone (L4T R32.7.2) builds, and the driver explicitly selects either the prod or dbg firmware [here](https://github.com/OE4T/linux-tegra-4.9/blob/ad2fc41d9ae8fa00515d4ee253fe7fad952dc363/nvidia/nvgpu/drivers/gpu/nvgpu/gv11b/acr_gv11b.c#L151).

Are you, by any chance, using a kernel that's based off an older L4T release?  The switch to using the prod/dbg firmware appears to be new to R32.7.x.

pticon:
Hi @madisongh 
Thank you for your quick reply. I was indeed using an older version (L4T R32.6.x).
After updating my tree with the R32.7.2, the issue is not present anymore.
We can disregard this PR because there is no issue.

PR #1029, opened 2022-08-18
Jetpack 5 GA: add nvidia gbm backend

kekiefer:
This change adds nvidia's binary gbm backend (`libnvidia-allocator.so`) from the Jetpack 5 GA release. It's enabled with a `PREFERRED_PROVIDER` entry, so users can switch to the open source backend if desired.

Also dropping a patch to egl-gbm that worked around not fully supported `EGL_NV_stream_consumer_eglimage`, which is now compliant and listed as a supported extension.


ichergui (approved):
LGTM

madisongh (changes requested):
Generally looks good, but the runtime (RPROVIDES/RDEPENDS) references should not use a `virtual/` prefix.  If you could drop that throughout, should be good to go.

PR #1030, opened 2022-08-18
jetpack-5-ga: drop gstreamer-1.14.conf

quaresmajose:


madisongh:
Right, I meant to do that when I delete the recipes, but forgot.  Thanks!

Could you add a Signed-off-by to the commit, and prefix the first line with `contrib: `?  If that's too much bother, I'll take care of the deletion with my next batch of updates.

quaresmajose:
Done.

The contrib layer only have a single recipe now, maybe is better to move it to meta-tegra-community and remove the layer as well

madisongh:
Thanks... and yes, I was thinking the same thing.

ichergui:
> Done.
> 
> 
> 
> The contrib layer only have a single recipe now, maybe is better to move it to meta-tegra-community and remove the layer as well

Thanks @quaresmajose for your PR

PR #1034, opened 2022-08-19
libnvvpi2: add pkg_postinst to create a symlink of vpi2.conf file

ichergui:
This is needed for letting the dynamic linker able to find the Nvidia
VPI library

### DIFF
--- a/recipes-devtools/vpi/libnvvpi2_2.1.6.bb
+++ b/recipes-devtools/vpi/libnvvpi2_2.1.6.bb
@@ -36,10 +36,14 @@ do_compile() {
 
 do_install() {
     install -d ${D}/opt/nvidia/vpi2
-    cp -R --preserve=mode,timestamps ${B}/opt/nvidia/vpi2/etc ${D}/opt/nvidia/vpi2/
     cp -R --preserve=mode,timestamps ${B}/opt/nvidia/vpi2/include ${D}/opt/nvidia/vpi2/
     cp -R --preserve=mode,timestamps ${B}/opt/nvidia/vpi2/lib ${D}/opt/nvidia/vpi2/
     ln -snf lib/aarch64-linux-gnu ${D}/opt/nvidia/vpi2/lib64
+    install -d ${D}${sysconfdir}/ld.so.conf.d
+    install -m 0644 ${B}/opt/nvidia/vpi2/etc/ld.so.conf.d/vpi2.conf ${D}${sysconfdir}/ld.so.conf.d/
+    install -d ${D}${nonarch_base_libdir}/firmware
+    install -m 0644 ${B}/opt/nvidia/vpi2/lib/aarch64-linux-gnu/priv/pva_auth_allowlist ${D}${nonarch_base_libdir}/firmware/
+    rm -f ${D}/opt/nvidia/vpi2/lib/aarch64-linux-gnu/priv/pva_auth_allowlist
 }
 
 INHIBIT_PACKAGE_STRIP = "1"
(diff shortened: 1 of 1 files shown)


recipes-devtools/vpi/libnvvpi2_2.1.6.bb
```diff
@@ -42,12 +42,18 @@ do_install() {
     ln -snf lib/aarch64-linux-gnu ${D}/opt/nvidia/vpi2/lib64
 }
 
+pkg_postinst:${PN}() {
+    if [ -e "$D/opt/nvidia/vpi2/etc/ld.so.conf.d/vpi2.conf" ]; then
+        ln -sf /opt/nvidia/vpi2/etc/ld.so.conf.d/vpi2.conf $D${sysconfdir}/ld.so.conf.d/vpi2.conf
```
madisongh, line 47: You should probably also check if `$D${sysconfdir}/ld.so.conf.d` exists, and create it if needed.

Looking at the postinst scriptlet in the deb package from NVIDIA, it looks like it also copies the `pva_auth_allowlist` file to `${nonarch_base_libdir}/firmware`, and writes some values under `/sys/kernel/debug` to enable VPU app authentication... although checking those entries on my Xavier NX, they were already set to the correct value, so we can probably ignore that.

madisongh (changes requested):

recipes-devtools/vpi/libnvvpi2_2.1.6.bb
ichergui: Thanks @madisongh 
I will introduce new changes

recipes-devtools/vpi/libnvvpi2_2.1.6.bb
```diff
@@ -42,6 +42,22 @@ do_install() {
...
 }
 
+pkg_postinst:${PN}() {
+    if [ -e "$D/opt/nvidia/vpi2/etc/ld.so.conf.d/vpi2.conf" ]; then
+        if [ ! -d "$D${sysconfdir}/ld.so.conf.d" ]; then
+            install -d $D${sysconfdir}/ld.so.conf.d
+        fi
+        ln -sf /opt/nvidia/vpi2/etc/ld.so.conf.d/vpi2.conf $D${sysconfdir}/ld.so.conf.d/
+    fi
+    if [ ! -d "$D/lib/firmware" ]; then
```
madisongh, line 52: `/lib` should be `${nonarch_base_libdir}` here.

recipes-devtools/vpi/libnvvpi2_2.1.6.bb
```diff
@@ -42,6 +42,22 @@ do_install() {
...
+        install -d $D/lib/firmware
+    fi
+    if [ -f "$D/lib/firmware/pva_auth_allowlist" ]; then
```
madisongh, line 55: Since this postinstall will only run at image creation time, I don't think you have to do this rename.  There's no way the old one could ever be put back.

madisongh (changes requested):
Thinking about it, maybe we should just install these files where the system needs to be in the first place, rather than doing this in a postinstall.  Unlike the NV deb package, this isn't going to be installed/removed at runtime, nor dynamically installed into a container.

I'm fine with leaving it the way you have it for now, though.  Just a couple of minor things mentioned in-line.

recipes-devtools/vpi/libnvvpi2_2.1.6.bb
ichergui: I'm doing this  already

recipes-devtools/vpi/libnvvpi2_2.1.6.bb
ichergui: It works, I tested it.
But I agree with you, I will remove it

ichergui:
> Thinking about it, maybe we should just install these files where the system needs to be in the first place, rather than doing this in a postinstall. Unlike the NV deb package, this isn't going to be installed/removed at runtime, nor dynamically installed into a container.
> 
> I'm fine with leaving it the way you have it for now, though. Just a couple of minor things mentioned in-line.

We can do that right now since we have the PR open 
I don't have any objection to do that :wink:

recipes-devtools/vpi/libnvvpi2_2.1.6.bb
```diff
@@ -36,18 +36,22 @@ do_compile() {
 
 do_install() {
     install -d ${D}/opt/nvidia/vpi2
-    cp -R --preserve=mode,timestamps ${B}/opt/nvidia/vpi2/etc ${D}/opt/nvidia/vpi2/
     cp -R --preserve=mode,timestamps ${B}/opt/nvidia/vpi2/include ${D}/opt/nvidia/vpi2/
     cp -R --preserve=mode,timestamps ${B}/opt/nvidia/vpi2/lib ${D}/opt/nvidia/vpi2/
     ln -snf lib/aarch64-linux-gnu ${D}/opt/nvidia/vpi2/lib64
+    install -d ${D}${sysconfdir}/ld.so.conf.d
+    install -m 6444 ${B}/opt/nvidia/vpi2/etc/ld.so.conf.d/vpi2.conf ${D}${sysconfdir}/ld.so.conf.d/
```
madisongh, line 43: `6444` here should be `0644`.

madisongh (changes requested):
One minor typo, otherwise looks good, thanks

recipes-devtools/vpi/libnvvpi2_2.1.6.bb
ichergui: Done

madisongh (approved):

PR #1038, opened 2022-08-23
Jetpack 5 GA: update container support

atharvanan1:
Tested with https://github.com/atharvanan1/jetson-containers/tree/a35149d52425843f6791c657044bc54dd9e910ac


atharvanan1:
Please find the test report here - https://drive.google.com/file/d/146FqosVXubAYwaZyKpq811JNqeBAveaC/view?usp=sharing

ichergui (approved):
Overall LGTM. Please see my comment regarding the `tegra234` platform 
Thanks @atharvanan1

recipes-bsp/tegra-binaries/tegra-configs_35.1.0.bb
```diff
@@ -20,6 +20,7 @@ SRC_URI += "\
     file://0001-Patch-udev-rules-for-OE-use.patch \
     file://0002-Patch-nv.sh-script-for-OE-use.patch \
     file://nv-l4t-bootloader-config.sh \
+    file://l4t.csv \
```
ichergui, line 23: I think we need a specific `l4t.csv` for `tegra234` (option 1) because of the `nvidia-display-driver` 

Please check [here](https://github.com/OE4T/meta-tegra/blob/wip-jetpack-5-ga/conf/machine/include/tegra234.inc#L42)
we need to have something like:
```
dev, /dev/nvidiactl
dev, /dev/nvidia0
dev, /dev/nvidia-modeset
```
Or (option 2), we have to provide the `GPU` devices at the `docker` runtime
Example:

```
$ docker run --runtime=nvidia \
    --device /dev/nvidia0:/dev/nvidia0 \
   --device /dev/nvidiactl:/dev/nvidiactl \
   --device /dev/nvidia-modeset:/dev/nvidia-modeset \
  ... 
```

Any thoughts ?

madisongh: I don't see `/dev/nvida0` or `/dev/nvidia-modeset` in the CSV in the stock L4T kit, either.  So how does it work on stock JetPack?

(`/dev/nvidiactl` is already there in the diff.)

madisongh (approved):
LGTM.  Thanks for taking care of this!

recipes-bsp/tegra-binaries/tegra-configs_35.1.0.bb
ichergui: > I don't see `/dev/nvida0` or `/dev/nvidia-modeset` in the CSV in the stock L4T kit, either. So how does it work on Orin in stock JetPack?
> 
> (`/dev/nvidiactl` is already there in the diff.)

As I mentioned in my option2, you have to provide the `GPU` devices for GPU rendering or displaying something.

I will try to play with different docker containers and share the results

madisongh: Let's resolve this question with a later PR, if needed, since this change set at least aligns with stock L4T.

PR #1039, opened 2022-08-23
gstreamer1.0-plugins-nveglgles: fix builds without x11

kekiefer:
Wayland only builds won't have the x11 header files, so only pick
up this header if we expect to have it.


ichergui (approved):
LGTM
Thanks @kekiefer

PR #1040, opened 2022-08-24
Minor updates in meta-tegra

ichergui:


external/openembedded-layer/recipes-support/opencv/opencv_4.6.%.bbappend
```diff
@@ -26,3 +26,8 @@ SRC_URI[opticalflow.sha256sum] = "e300c02e4900741700b2b857965d2589f803390849e1e2
 
 # No stable URI is available for NVIDIAOpticalFlowSDK
 INSANE_SKIP:append:cuda = " src-uri-bad"
+
+# Build paths are currently embedded
+INSANE_SKIP:${PN}-dev += "buildpaths"
+INSANE_SKIP:${PN}-src += "buildpaths"
+INSANE_SKIP:libopencv-core += "buildpaths"
```
kekiefer, line 33: Are these caused by our append?

ichergui: No, maybe I should remove that commit

ichergui: Tbh, I was not sure about pushing that 😞

kekiefer: Yeah I don't think we need to be in the business of fixing up opencv in this layer

madisongh (changes requested):
Thanks for fixing the helper script.  Let's not put the opencv package QA workaround here, though.

external/openembedded-layer/recipes-support/opencv/opencv_4.6.%.bbappend
madisongh: Yes, please drop this change.

ichergui:
> Thanks for fixing the helper script. Let's not put the opencv package QA workaround here, though.

I will remove that commit. Should we create `.bbappend` in `meta-tegra-community` ?

madisongh (approved):

madisongh:
> I will remove that commit. Should we create `.bbappend` in `meta-tegra-community` ?

No.  It should be fixed upstream.

PR #1042, opened 2022-08-25
tegra-common: always add tegraflash as it is the most used image

quaresmajose:
Using a weak default for that it is very sensitive and not take effect
in distros that uses other types.

Setting the IMAGE_FSTYPES with a weak default can also introduce circular
dependencies because the tegraflash is lost when something is set in
IMAGE_FSTYPES and adding this again using an append in disto config
add circular dependencies for example in tegra-espimage.

Signed-off-by: Jose Quaresma <jose.quaresma@foundries.io>

### DIFF
--- a/conf/machine/include/tegra-common.inc
+++ b/conf/machine/include/tegra-common.inc
@@ -38,7 +38,7 @@ IMAGE_ROOTFS_ALIGNMENT ?= "4"
 TEGRA_BLBLOCKSIZE ?= "${@int(d.getVar('IMAGE_ROOTFS_ALIGNMENT')) * 1024}"
 EXTRA_IMAGECMD:ext4 ?= "-i 4096 -b 4096"
 IMAGE_CLASSES += "image_types_tegra"
-IMAGE_FSTYPES ?= "tegraflash"
+IMAGE_FSTYPES += "tegraflash"
 
 TEGRA_ESSENTIAL_EXTRA_RDEPENDS ??= ""
 MACHINE_FEATURES = "alsa usbhost pci rtc cuda ext2"


madisongh:
I guess I'm OK with this.  I was a bit concerned about doing a hard setting here when I originally put it in, but looking around at various BSP layers, I see some that do it, while some use the weak setting.

We should probably fix the `tegra-espimage` recipe to use `IMAGE_FSTYPES:forcevariable` to prevent other possible conflicts, too.

quaresmajose:
> I guess I'm OK with this. I was a bit concerned about doing a hard setting here when I originally put it in, but looking around at various BSP layers, I see some that do it, while some use the weak setting.

I thing it will be very hard to build a funcional tegra image without using the brilliant `image_types_tegra.bbclass` or some fork of it. However for these exceptional cases it is more useful to use `IMAGE_FSTYPES:remove = "tegraflash"` in my opinion.

> We should probably fix the tegra-espimage recipe to use IMAGE_FSTYPES:forcevariable to prevent other possible conflicts, too.

Using `IMAGE_FSTYPES:forcevariable` in `tegra-espimage` seams a good ideia as `esp` is the only supported type of the recipe.
Maybe `esp` is a bit of a generic name and would be perhaps more appropriate `tegra-esp` but this is clearly not a problem.

quaresmajose:
@madisongh can this patch be ported to master and kirkstone? there is no problem of circular dependencies on these but in my opinion it would be positive have this branches aligned

madisongh:
This will make it into master when the WIP branch gets merged, hopefully in the next few days.  Post-merge, we'll be backporting the JetPack 5 content to kirkstone, and current kirkstone will become kirkstone-l4t-r32.7.x.  I can cherry-pick it there at that point.

quaresmajose:
right, in the meantime it will be there.

PR #1045, opened 2022-08-26
image_types_tegra: replace LNXFILE_b

atharvanan1:
- rootfs_ab.xml file for xavier-nx uses LNXFILE_b as filename
  placeholder

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -89,6 +89,7 @@ tegraflash_post_sign_pkg() {
 tegraflash_post_sign_pkg:tegra194() {
     mksparse -b ${TEGRA_BLBLOCKSIZE} --fillpattern=0 "${IMAGE_TEGRAFLASH_ROOTFS}" ${IMAGE_BASENAME}.img
     [ "${TEGRA_SPIFLASH_BOOT}" = "1" -o "${TEGRA_ROOTFS_AND_KERNEL_ON_SDCARD}" = "1" ] || rm -f ./${IMAGE_BASENAME}.${IMAGE_TEGRAFLASH_FS_TYPE}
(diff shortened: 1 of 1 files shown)


kekiefer:
Doing this would increase the time to flash, and the benefit is unclear, as the B slots are meant to be written during update.

atharvanan1:
Using ab flash layout doesn't really work, if these settings are not set.

kekiefer:
Ok, I think I see what you mean -- since it would try to find those files that don't exist? How about replacing them with empty values?

atharvanan1:
I'll try that and keep this thread posted with results.

madisongh:
The other thing that NVIDIA's `flash.sh` script does is to make a second copy of `boot.img` with a modified kernel command line that instructs the kernel to boot off the B rootfs.  Rookie mistake - relying on that won't work for any production use of A/B OTA updates unless we force the OTA payload to include two copies of the kernel image and the OTA client has a mechanism to decide which one to use based on whether it's installing into the A or B slot.  SIgh.

Our initramfs script shouldn't need that, as long as there's some way for it to know which boot chain is the current one.  If UEFI isn't passing that as a kernel command line addition, maybe we could extract that info from an EFI variable?

madisongh (changes requested):
Could you make the corresponding changes in the `tegra234` functions as well?

ichergui:
@atharvanan1 
could you please add `Signed-off-by` to you commit  ?
Thanks

madisongh:
You'll also need to fix the other places where `APPFILE_b` needs similar treatment as `APPFILE`, both in the bbclass and the flash-helper scripts.

I have some changes to the `tegra-minimal-init` scripts that use the `BootChainFwCurrent` EFI variable to determine which rootfs to mount.

madisongh:
Thanks for doing the initial groundwork on this!

PR #1058, opened 2022-09-04
tegra234.inc: no need to add nvidia-drm and nvidia-modeset to the

ichergui:
KERNEL_MODULE_AUTOLOAD list.

This fixes the following issue

[    44.378] (II) NVIDIA(0): NVIDIA GPU Orin (nvgpu) (GA10B) at SoC (GPU-0)
[    44.378] (--) NVIDIA(0): Memory: 31269432 kBytes
[    44.378] (--) NVIDIA(0): VideoBIOS:
[    44.378] (EE) NVIDIA(GPU-0): Failed to acquire modesetting permission.
[    44.378] (EE) NVIDIA(0): Failing initialization of X screen

### DIFF
--- a/conf/machine/include/tegra234.inc
+++ b/conf/machine/include/tegra234.inc
@@ -12,7 +12,7 @@ INITRAMFS_FSTYPES:append = "${@'' if d.getVar('INITRAMFS_IMAGE_BUNDLE') == '1' e
 
 SERIAL_CONSOLES ?= "115200;ttyTCU0"
 
-KERNEL_MODULE_AUTOLOAD ?= "nvgpu nvidia nvidia-drm nvidia-modeset"
+KERNEL_MODULE_AUTOLOAD ?= "nvgpu nvidia nvidia-drm"
 
 NVIDIA_CHIP = "0x23"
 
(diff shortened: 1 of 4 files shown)


ichergui:
* This patch was tested on `Jetson Orin AGX` with the following image:
* `demo-image-sato` 
* `demo-image-full`
Both work fine

IMG-1791

ichergui:
Can anyone please try `demo-image-weston` image ?

kekiefer (changes requested):
I think we should leave the `KERNEL_MODULE_AUTOLOAD` alone, regardless, as this doesn't hurt x11.

The question is about adding the `modeset=1` to `nvidia-drm`. Perhaps we could only choose to remove this from `KERNEL_MODULE_PROBECONF` if we have a distro features setup with `x11` and not `wayland`?

ichergui:
> I think we should leave the `KERNEL_MODULE_AUTOLOAD` alone, regardless, as this doesn't hurt x11.
> 
> The question is about adding the `modeset=1` to `nvidia-drm`. Perhaps we could only choose to remove this from `KERNEL_MODULE_PROBECONF` if we have a distro features setup with `x11` and not `wayland`?

Our distro features have both `X11` and `Wayland`

madisongh:
> > I think we should leave the `KERNEL_MODULE_AUTOLOAD` alone, regardless, as this doesn't hurt x11.
> > The question is about adding the `modeset=1` to `nvidia-drm`. Perhaps we could only choose to remove this from `KERNEL_MODULE_PROBECONF` if we have a distro features setup with `x11` and not `wayland`?
> 
> Our distro features have both `X11` and `Wayland`

But @kekiefer 's comment still stands.  Can you leave the modules in `KERNEL_MODULE_AUTOLOAD` and just use  the recipe to do the `nvidia-drm modeset=1` probeconf file?

madisongh:
I've merged manually with the changes suggested above.

PR #1063, opened 2022-09-05
haveged: reintroduce systemd service

moto-timo:
Upstream meta-oe dropped the systemd service since it is no longer
needed in Linux kernel > 5.6. But we are still on 4.9.y so we need it.

Based on https://raw.githubusercontent.com/jirka-h/haveged/v1.9.14/init.d/service.redhat

Signed-off-by: Tim Orling <tim.orling@konsulko.com>

### DIFF
--- a/external/openembedded-layer/recipes-extended/haveged/haveged_%.bbappend
+++ b/external/openembedded-layer/recipes-extended/haveged/haveged_%.bbappend
@@ -1,0 +1,19 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
+inherit ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'systemd', '', d)}
+
+SYSTEMD_PACKAGES = "${PN}"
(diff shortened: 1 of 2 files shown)


external/openembedded-layer/recipes-extended/haveged/haveged_%.bbappend
```diff
@@ -0,0 +1,17 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${BPN}:"
+inherit ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'systemd', '', d)}
+
+SYSTEMD_PACKAGES = "${PN}"
+SYSTEMD_SERVICE:${PN} = "haveged.service"
+
```
madisongh, line 6: We'll also need to set
```
```
so our additions are tegra-specific.

madisongh (changes requested):
Mostly there, see in-line

madisongh (approved):

PR #1075, opened 2022-09-15
Support virtual/dtb provider for out-of-tree dtb/dtbo

kekiefer:
This set of changes accommodate out-of-tree device tree binaries generated by a provider of `virtual/dtb`. Overall, this enables consumers of meta-tegra to generate the main kernel dtbs and dtbos, as well as the dtbos that can be used by EFI, without patching the linux-tegra or EDK2 sources.

The approach was derived from conventions set forth in the meta `devicetree` bbclass and its handling by the `kernel-fitimage` class. The assumption is that these device tree binaries will be placed under the `${DEPLOYDIR}/devicetree` location during deploy, or in `/boot/devicetree` on install. These paths are not variable in the `devicetree` bbclass, and as such are not variable here, and their contents simply globbed.

To enable building a dtb that referenced nvidia's overlayed kernel device tree sources, I required a minor fix to the kernel sources so that the upstream includes would not be used. This PR stands alone, and we can cross that bridge subsequently.

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -16,6 +16,8 @@ TEGRA_SIGNING_ARGS ??= ""
 TEGRA_SIGNING_ENV ??= ""
 TEGRA_SIGNING_EXCLUDE_TOOLS ??= ""
 TEGRA_SIGNING_EXTRA_DEPS ??= ""
+DTB_EXTRA_DEPS ??= "${@'virtual/dtb:do_populate_sysroot' if d.getVar('PREFERRED_PROVIDER_virtual/dtb') else ''}"
+EXTERNAL_KERNEL_DEVICETREE ??= "${@'${RECIPE_SYSROOT}/boot/devicetree' if d.getVar('PREFERRED_PROVIDER_virtual/dtb') else ''}"
 
 TEGRA_BUPGEN_SPECS ??= "boardid=${TEGRA_BOARDID};fab=${TEGRA_FAB};boardrev=${TEGRA_BOARDREV};chiprev=${TEGRA_CHIPREV}"
 TEGRA_BUPGEN_STRIP_IMG_NAMES ??= ""
@@ -275,6 +277,31 @@ BOOTFILES:tegra234 = "\
     camera-rtcpu-sce.img \
 "
 
+copy_dtbs() {
+    local destination=$1
+    local dtb dtbf
+    for dtb in ${KERNEL_DEVICETREE}; do
+        dtbf=`basename $dtb`
+        if [ -e $destination/$dtbf ]; then
+            bbnote "Overwriting $destination/$dtbf with KERNEL_DEVICETREE content"
+            rm -f $destination/$dtbf
+        fi
+        bbnote "Copying KERNEL_DEVICETREE entry $dtbf to $destination"
+        cp -L "${DEPLOY_DIR_IMAGE}/$dtbf" $destination/$dtbf
+    done
+    if [ -n "${EXTERNAL_KERNEL_DEVICETREE}" ]; then
+        for dtb in $(find "${EXTERNAL_KERNEL_DEVICETREE}" \( -name '*.dtb' -o -name '*.dtbo' \) -printf '%P\n' | sort); do
+            dtbf=`basename $dtb`
+            if [ -e $destination/$dtbf ]; then
+                bbnote "Overwriting $destination/$dtbf with EXTERNAL_KERNEL_DEVICETREE content"
+                rm -f $destination/$dtbf
+            fi
+            bbnote "Copying EXTERNAL_KERNEL_DEVICETREE entry $dtbf to $destination"
+            cp -L "${EXTERNAL_KERNEL_DEVICETREE}/$dtbf" $destination/$dtbf
+        done
+    fi
+}
+
 create_tegraflash_pkg() {
     :
 }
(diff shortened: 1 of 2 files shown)


classes/image_types_tegra.bbclass
```diff
@@ -295,7 +321,7 @@ create_tegraflash_pkg:tegra194() {
         cp "${IMAGE_TEGRAFLASH_DATA}" ./${DATAFILE}
         DATAARGS="--datafile ${DATAFILE}"
     fi
-    cp -L "${DEPLOY_DIR_IMAGE}/${DTBFILE}" ./${DTBFILE}
+    copy_dtbs "${WORKDIR}/tegraflash"
```
kekiefer, line 324: Note that the resultant functionality in a non-default setup could differ slightly from the original. Normally, `DTBFILE` is generated as the first entry in `KERNEL_DEVICETREE`, and only one entry is in `KERNEL_DEVICETREE`, so there is no functional change. Since it may be useful to copy over multiple device tree binaries (overlay support for EFI) multiple entries in `KERNEL_DEVICETREE` or from a `virtual/dtb` provided will all be copied over.

madisongh: This will change the default setup, too, but I'm cool with that, since we actually have multiple DTBs listed for Orin.

However, we currently have separate `cp` commands for copying over all `.dtbo` files from `${DEPLOY_DIR_IMAGE}`, too. If you want this other-DTB-provider to cover the overlays as well, then you should probably change those commands, too.  It's a little tricky, since (by default) some of the overlays originate from the UEFI build, while some originate from the kernel build.

recipes-bsp/uefi/l4t-launcher-extlinux.bb
```diff
@@ -20,10 +20,14 @@ def compute_dependencies(d):
...
     if initramfs_image != '' and (d.getVar('INITRAMFS_IMAGE_BUNDLE') or '') != '1':
         deps += " %s:do_image_complete" % initramfs_image
+    if d.getVar("PREFERRED_PROVIDER_virtual/dtb"):
+        deps += " virtual/dtb:do_deploy"
     return deps
 
 
 PATH =. "${STAGING_BINDIR_NATIVE}/tegra-flash:"
+DEPLOY_DIR_IMAGE_DTB = "${DEPLOY_DIR_IMAGE}"
+DEPLOY_DIR_IMAGE_DTB:append = "${@'/devicetree' if d.getVar('PREFERRED_PROVIDER_virtual/dtb') else ''}"
```
madisongh, line 30: Rather than using `:append` here, can you just put this directly into the setting above?  Otherwise, if someone uses a bbappend to modify `DEPLOY_DIR_IMAGE_DTB`, the append will be applied to their modified setting, which isn't really what's intended here.

madisongh (changes requested):
Thanks... I'm generally OK with this, but see in-line.

classes/image_types_tegra.bbclass
```diff
@@ -72,6 +72,14 @@ python() {
...
     if pt not in valid_types:
         bb.fatal("TEGRAFLASH_PACKAGE_FORMAT %s not recognized, supported types are: %s" % (pt, ', '.join(valid_types)))
+
+    # handle external provider for dtb (adapted from kernel-fitimage.bbclass)
+    providerdtb = d.getVar("PREFERRED_PROVIDER_virtual/dtb")
+    if providerdtb:
+        d.appendVar('CONVERSION_DEPENDS_bup-payload', ' virtual/dtb:do_populate_sysroot')
+        d.appendVarFlag('do_image_tegraflash', 'depends', ' virtual/dtb:do_populate_sysroot')
+        # d.appendVarFlag('create_bup_payload_image', 'depends', ' virtual/dtb:do_populate_sysroot')
+        d.setVar('EXTERNAL_KERNEL_DEVICETREE', "${RECIPE_SYSROOT}/boot/devicetree")
```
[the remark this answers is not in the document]
kekiefer: Yes, I can do it this way. I chose the python function since it was this way in the kernel/fitimage class, but I see your points, and furthermore it is more consistent with other schemes used here.

recipes-bsp/uefi/l4t-launcher-extlinux.bb
kekiefer: Yeah, this is awkward. I'll clean this up.

classes/image_types_tegra.bbclass
madisongh: > Is the concern that this is not doing enough to ship all the overlay content from the virtual/dtb provider?

No, I was concerned that if the intent is to have the `virtual/dtb` provider replace DTB/DTBO files that would normally come from either the kernel build or the UEFI build (or the NVIDIA prebuilts), to make sure that the other copy command  won't wipe out the ones being provided here.

kekiefer: Ok, understood. What I just move the copy command down below? I can emit an info message when replacing existing content. It seems that anything generated in a custom recipe should be what the user intends to use.

madisongh: I wasn't sure whether just moving the `copy_dtbs` to after was the right way to deal with it, or skipping the default `*.dtbo` copy altogether if `EXTERNAL_KERNEL_DEVICETREE` is defined, I figured you'd know. :)

kekiefer: It appears to be. I'll take the time to finish evaluating the switch from the plugin-manager to bootloader and kernel overlays for my custom hardware, just to be sure nothing is missed, and then make the call.

kekiefer: I've moved the `copy_dtbs` below the other dtb copies, so that any user-generated content will get priority.

kekiefer:
This should be good now. There may be some more cleanup of dtb copies from various other locations that we could factor into `copy_dtbs`, but it isn't critical, and I'd sooner get this in with a small footprint.

Two more PRs are forthcoming to make building DTBs against kernel sources possible, and a second to the UEFI to fix customer board part number handling.

ichergui (approved):
LGTM

madisongh (approved):

PR #1079, opened 2022-09-18
edk2-firmware-tegra: fix eeprom customer part number handling

kekiefer:
The logic in edk2-firmware-tegra's eeprom handling logic suffers a few oversights regarding customer part numbers addressed here.

First, the eeprom parsing didn't match the old cboot in terms of not requiring initialization of the length field in the eeprom. Existing hardware may not have this field set properly, so we check if it is zero and bypass the length check for this case.

Second, customer part numbers with the leading 0xcc were not being handled properly. This change fixes the addition of these parts to the /chosen/ids node, as well as fixes the handling of these parts for for comparison in the overlay logic.

### DIFF
--- a/recipes-bsp/uefi/edk2-firmware-tegra/0001-Fix-eeprom-customer-part-numbers.patch
+++ b/recipes-bsp/uefi/edk2-firmware-tegra/0001-Fix-eeprom-customer-part-numbers.patch
@@ -1,0 +1,83 @@
+diff -ru edk2-tegra_35.1.orig/edk2-nvidia/Silicon/NVIDIA/Drivers/EepromDxe/Eeprom.c edk2-tegra_35.1/edk2-nvidia/Silicon/NVIDIA/Drivers/EepromDxe/Eeprom.c
+--- edk2-tegra_35.1.orig/edk2-nvidia/Silicon/NVIDIA/Drivers/EepromDxe/Eeprom.c	2022-08-24 13:08:10.015078516 -0700
++++ edk2-tegra_35.1/edk2-nvidia/Silicon/NVIDIA/Drivers/EepromDxe/Eeprom.c	2022-09-16 16:46:38.880756491 -0700
+@@ -49,6 +49,9 @@
+   if (ChipID == T194_CHIP_ID) {
+     T194EepromData = (T194_EEPROM_DATA *)EepromData;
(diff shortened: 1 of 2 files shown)


kekiefer:
Which brings up the question, do we want to start initializing the length field in `tegra-eeprom-tool`, even though we're working around it here? The existing logic doesn't use this length for anything other than a check, the CRC is always computed on the first 254 bytes.

recipes-bsp/uefi/edk2-firmware-tegra/0001-Fix-eeprom-customer-part-numbers.patch
```diff
@@ -0,0 +1,83 @@
+diff -ru edk2-tegra_35.1.orig/edk2-nvidia/Silicon/NVIDIA/Drivers/EepromDxe/Eeprom.c edk2-tegra_35.1/edk2-nvidia/Silicon/NVIDIA/Drivers/EepromDxe/Eeprom.c
+--- edk2-tegra_35.1.orig/edk2-nvidia/Silicon/NVIDIA/Drivers/EepromDxe/Eeprom.c	2022-08-24 13:08:10.015078516 -0700
++++ edk2-tegra_35.1/edk2-nvidia/Silicon/NVIDIA/Drivers/EepromDxe/Eeprom.c	2022-09-16 16:46:38.880756491 -0700
+@@ -49,6 +49,9 @@
+   if (ChipID == T194_CHIP_ID) {
+     T194EepromData = (T194_EEPROM_DATA *)EepromData;
+     EepromBoardInfo = (TEGRA_EEPROM_BOARD_INFO *) BoardInfo;
++    if (T194EepromData->PartNumber.Leading[0] == 0xcc)
```
ichergui, line 8: We are using `0xcc` in different places in the code. Maybe we should create a `define` with a significant name

ichergui (changes requested):

madisongh:
> Which brings up the question, do we want to start initializing the length field in tegra-eeprom-tool, even though we're working around it here?

It wouldn't hurt.

madisongh:
BTW, you could try submitting this upstream to NVIDIA, if you haven't already.  It looks like they take issues and PRs on their repo, but if not, via the dev forum.

PR #1049, opened 2022-08-27
OPTEE for Tegra

ichergui:
* Tests were done only by using `Jetson Orin AGX` 
* When using recipes that build from source then doing tests,  system hangs `Investigation is needed`

PR #1101, opened 2022-10-23
OP-TEE source builds take 2

madisongh:
Replaces #1049.  Differences:

* Patch added to set `-fcommon` in the optee_os core, since the NVIDIA downstream changes to the memory management code requires the pre-gcc-10 common-block behavior.
* Added in missing TAs that are built with the core `optee-os` recipe but were not getting installed.
* Version numbering is based on OP-TEE upstream version with the L4T version appended.
* Recipes are better aligned with those in meta-arm, using the same naming and mostly the same structure.
* The prebuilt recipes have been replaced with a single recipe and moved over to `recipes-bsp/tegra-binaries`.
* The TF-A recipe's version number has been corrected - the NVIDIA sources are based off 2.6 (approximately) now.  We still call it `arm-trusted-firmware` rather than `trusted-firmware-a`, though.
* NVIDIA sample apps are packaged in a way that should be usable for LUKS setup in an initrd.
* Defaulting to building from source, rather than using prebuilts.

With this change set, the full `optee-test` suite passes on both Xavier and Orin platforms - no hangs.


ichergui (approved):
LGTM

madisongh:
I'll back-port this to the kirkstone branch once I've seen a passing build of the demo distro on master with these changes in place.

ichergui:
> I'll back-port this to the kirkstone branch once I've seen a passing build of the demo distro on master with these changes in place.

Superb 
Thanks @madisongh

PR #1109, opened 2022-11-01
Source builds of nvcompositor

kekiefer:
nvcompositor emits a spurious warning message for every frame that goes through it. This typically goes unnoticed until GST_DEBUG is at least at the warn level, but besides being annoying it is doing unnecessary work.

This PR patches the nvcompositor to skip the extraneous mapping call, and sets up the gstreamer1.0-plugins-tegra collection to use the source build of the nvcompositor.


ichergui:
@kekiefer 
Could you please update your branch ? it looks like `out-of-date` 
Thanks

kekiefer:
Updated

PR #1111, opened 2022-11-08
Fixes for sysvinit-based startup issues

### LINKED ISSUE #1110: Default orin machine (jetson-agx-orin-devkit) hangs on S20nv_update_verifier
Have just tested jetson-agx-orin-devkit in combination with latest kirkstone of meta-tegra and poky.

It hangs on S20nv_update_verifier init script, from tegra-redundant-boot

The problem is that it waits on a partition label not available on the default orin flash image.

From the script

```
root@jetson-agx-orin-devkit:/etc/rc5.d# cat /etc/init.d/nv_update_verifier 
#! /bin/sh
NAME=nv_update_verifier
DESC="OTA bootloader update verifier"

case "$1" in
  start|restart)
      while [ ! -e /dev/disk/by-partlabel/kernel ]; do sleep 1; done
      sleep 2
      /usr/sbin/nvbootctrl verify < /dev/null
      ;;
...
```

And the default partition labels are

```
root@jetson-agx-orin-devkit:/dev/disk/by-partlabel# ls -la
drwxr-xr-x    2 root     root           260 Jan  1  1970 .
drwxr-xr-x    7 root     root           140 Jan  1  1970 ..
lrwxrwxrwx    1 root     root            15 Jan  1  1970 APP -> ../../mmcblk0p1
lrwxrwxrwx    1 root     root            15 Jan  1  1970 A_kernel -> ../../mmcblk0p2
lrwxrwxrwx    1 root     root            15 Jan  1  1970 A_kernel-dtb -> ../../mmcblk0p3
lrwxrwxrwx    1 root     root            15 Jan  1  1970 A_reserved_on_user -> ../../mmcblk0p4
lrwxrwxrwx    1 root     root            15 Jan  1  1970 B_kernel -> ../../mmcblk0p5
lrwxrwxrwx    1 root     root            15 Jan  1  1970 B_kernel-dtb -> ../../mmcblk0p6
lrwxrwxrwx    1 root     root            15 Jan  1  1970 B_reserved_on_user -> ../../mmcblk0p7
lrwxrwxrwx    1 root     root            16 Jan  1  1970 esp -> ../../mmcblk0p10
lrwxrwxrwx    1 root     root            15 Jan  1  1970 recovery -> ../../mmcblk0p8
lrwxrwxrwx    1 root     root            15 Jan  1  1970 recovery-dtb -> ../../mmcblk0p9
lrwxrwxrwx    1 root     root            16 Jan  1  1970 reserved -> ../../mmcblk0p11
```

So it will hang on this init script forever.

madisongh:
* Fix the installation of the optee supplicant init script to match the INITSCRIPT_NAME
* Drop the wait on the kernel partition for `nv_update_verifier`

Fixes #1110

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-redundant-boot/nv_update_verifier.init
+++ b/recipes-bsp/tegra-binaries/tegra-redundant-boot/nv_update_verifier.init
@@ -4,9 +4,9 @@ DESC="OTA bootloader update verifier"
 
 case "$1" in
   start|restart)
-      while [ ! -e /dev/disk/by-partlabel/kernel ]; do sleep 1; done
-      sleep 2
+      echo -n "Running $DESC: "
       /usr/sbin/nvbootctrl verify < /dev/null
+      echo "[OK]"
       ;;
   stop)
       ;;
--- a/recipes-bsp/tegra-binaries/tegra-redundant-boot/nv_update_verifier.service
+++ b/recipes-bsp/tegra-binaries/tegra-redundant-boot/nv_update_verifier.service
@@ -1,6 +1,6 @@
 [Unit]
 Description=OTA bootloader update verifier
-After=nvstartup.service dev-disk-by\x2dpartlabel-kernel.device
+After=nvstartup.service
 
 [Service]
 Type=simple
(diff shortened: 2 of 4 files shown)


madisongh:
Tested this using demo-image-sato builds out of the demo distro, reconfigured to use sysvinit instead of systemd as the INIT_MANAGER.

PR #1130, opened 2022-12-27
Support for initrd-based flashing

madisongh:
This is the cleaned-up series of commits for implementing initrd-based flashing support.   The content is identical to the original work-in-progress branch, with the exception of the UEFI update and addition of BootOrder DTB overlays, neither of which are needed for the final implementation.

This is needed for flashing to a Jetson where you want the APP partition (rootfs) to reside on an external storage device, but can also be used instead of the normal `doflash.sh` script for flashing to internal storage.


madisongh:
See [this wiki page](https://github.com/OE4T/meta-tegra/wiki/initrd-flashing-support) for more information.

dwalkes:
LGTM, thanks Matt!  I had one issue nobody else has reproduced when using `jetson-xavier-nx-devkit` with SDCard (default) boot, details in https://gitter.im/OE4T/initrd-flashing?at=63ab9470fb195421bd594918.  `do-flash.sh` for this scenario was an acceptable workaround.

PR #1132, opened 2022-12-27
libglvnd: add libxcb-glx runtime dependency

ibstewart:
This fixes Vulkan support with X11 (tested with `vulkaninfo` tool in addition to other Vulkan-enabled apps).


madisongh:
I don't see a direct dependency between libglvnd itself and libxcb-glx.  It looks like the dependency is from a dlopen() call inside libGLX_nvidia.so on possibly several of the libxcb-XXX libraries... `strings` turns up mentions of libxcb-dri3, libxcb-present, and libxcb-sync, in addition to libxcb-glx.  An RDEPENDS in the `tegra-libraries-glxcore` recipe is probably the right fix for this.

ibstewart:
You're right, thanks for looking into that Matt. I updated/tested the change and changed the PR to master.

PR #1133, opened 2023-01-06
tegra-configs: Fix RDEPENDS remove of bash from tegra-configs-bootloader

Kleist:
Was using the old _ syntax instead of :.

Signed-off-by: Andreas Helbech Kleist <akle@ambu.com>


ichergui (approved):
@Kleist Thanks

PR #1135, opened 2023-01-09
Adding COMPATIBLE_SPEC support for BUP payloads

madisongh:
This is already present in the JetPack 5/R35.1.0-based branches, but was more involved for R32.7.x because:
* More platforms are involved
* tegra-boot-tools also needed updating

When using the stock NVIDIA bootloader update tools, the `setup-nv-boot-control` service will add a COMPATIBLE_SPEC line to the configuration file so that BUP payload entries either matching the actual TNSPEC or the compatible spec will be considered valid for an update.  The new version of the `tegra-bootloader-update` tool implements similar logic for its compatibility checks.

With the above changes, the `TEGRA_BUP_SPECS` settings for all machines needed revising.  In most cases, fewer specs are required, making BUP packages smaller and taking less time to build/sign.

For reviewing, I'd appreciate it if someone could cross check the `setup-nv-boot-control.sh` changes here (and the translation of the logic in that script to the C code [here](https://github.com/OE4T/tegra-boot-tools/blob/86d9f18a7eec7570ffa69e12a3481bb10e6c0764/bup.c#L340-L480) in tegra-boot-tools) to derive the compatible spec string against NVIDIA's implementation in their `nv-l4t-bootloader-config.sh` script.  To get a copy of the NVIDIA script, extract it from the `nvidia-l4t-init` package, which you can download [here](https://repo.download.nvidia.com/jetson/t186/pool/main/n/nvidia-l4t-init/nvidia-l4t-init_32.7.2-20220417024839_arm64.deb), with:

       $ dpkg-deb --fsys-tarfile nvidia-l4t-init_32.7.2-20220417024839_arm64.deb | tar -x -f- --strip-components=4 ./opt/nvidia/l4t-bootloader-config/nv-l4t-bootloader-config.sh

The relevant function in that script is called `generate_compatible_spec_info`.

### DIFF
--- a/conf/machine/include/tegra210.inc
+++ b/conf/machine/include/tegra210.inc
@@ -43,8 +43,6 @@ TEGRA_FAB ?= "100"
 TEGRA_BOARDSKU ?= ""
 TEGRA_BOARDREV ?= ""
 TEGRA_CHIPREV ?= "0"
-# Extracted from l4t_generate_soc_bup.sh for BOARDID=2180 and board=jetson-tx1-devkit
-TEGRA_BUPGEN_SPECS ?= "fab=100 fab=410"
 
 PARTITION_LAYOUT_EXTERNAL ?= "external_layout.xml"
 BOOT_PARTITIONS_ON_EMMC = "1"
--- a/conf/machine/jetson-agx-xavier-devkit.conf
+++ b/conf/machine/jetson-agx-xavier-devkit.conf
@@ -5,9 +5,8 @@
 require conf/machine/include/tegra194.inc
 
 # Extracted from l4t_generate_soc_bup.sh for BOARDID=2888 and board=jetson-agx-xavier-devkit
-TEGRA_BUPGEN_SPECS ?= "fab=400;boardsku=0001;boardrev=H.0 \
-                       fab=400;boardsku=0001;boardrev=J.0 \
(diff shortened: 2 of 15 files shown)


dwalkes:
Thanks @madisongh.  I needed to use ` nvidia-l4t-init_32.7.2-20220420143418_arm64.deb` with the dpkg-deb line above, so the full `dpkg-deb` command was:
```
dpkg-deb --fsys-tarfile nvidia-l4t-init_32.7.2-20220420143418_arm64.deb |  tar -x -f- --strip-components=4 ./opt/nvidia/l4t-bootloader-config/nv-l4t-bootloader-config.sh
```
I see the logic in `update_jetson_xxx` functions within  `nv-l4t-bootloader-config.sh` which you changed from bash string checks to `sh` and `grep`.  The logic looks correct to me.

I'll try testing this branch with `tegrademo-mender` next.

madisongh:
Thanks @dwalkes.  Looks like I pasted in the link to the t210 version of that deb package, rather than the t186/t194 one.  The script is the same in either one, though. I'll fix the link.

dwalkes:
> I'll try testing this branch with tegrademo-mender next.

Verified with jetson-agx-xavier-devkit.

LGTM, thanks @madisongh !

PR #1139, opened 2023-01-17
image_types_tegra: use a weak to assign the IMAGE_TEGRAFLASH_INITRD_FLASHER

quaresmajose:
There are other distros that don't use the cboot initrd images and only uses the default cpio.gz so add an options to change the image.

Signed-off-by: Jose Quaresma <jose.quaresma@foundries.io>


ichergui (approved):
LGTM

madisongh:
I don't object to this change, but I'm curious about the reason stated in the commit message. The initrd flasher has to be in a form that's compatible with NVIDIA's bootloaders, regardless of what bootloader you might be using for loading the OS you're building.  Do you have some other mechanism in place to produce an Android kernel image (which is all the `cboot` image type is)?

quaresmajose:
I didn't get to test if this new flash method works without the `cboot` image but as we use the `cpio.gz` and not build with the `cpio.gz.cboot` and `cpio.gz.cboot.bup-payload` our build is failing.
Maybe another approach can be made this new flash method configurable in a way we disable it if needed.

madisongh:
> Maybe another approach can be made this new flash method configurable in a way we disable it if needed.

Yeah, I was thinking about providing a way to disable it altogether, for folks that don't need/want it.  If the change here will get you by until I figure that out, that's fine.  If you could rebase on top of master latest, I'll go ahead and pull this in.

quaresmajose:
```
meta-tegra/conf/machine/include/tegra234.inc:INITRAMFS_FSTYPES:append = "${@'' if d.getVar('INITRAMFS_IMAGE_BUNDLE') == '1' else ' cpio.gz.cboot cpio.gz.cboot.bup-payload'}"
meta-tegra/conf/machine/include/tegra194.inc:INITRAMFS_FSTYPES:append = "${@'' if d.getVar('INITRAMFS_IMAGE_BUNDLE') == '1' else ' cpio.gz.cboot cpio.gz.cboot.bup-payload'}"
```
The issue is because of this snip that introduces this type of images in every recipe that uses INITRAMFS_FSTYPES.
If we change it to INITRAMFS_FSTYPES_TEGRA we can customize it be recipe instead of globally. Do you think it is a good approach?

quaresmajose:
> > Maybe another approach can be made this new flash method configurable in a way we disable it if needed.
> 
> Yeah, I was thinking about providing a way to disable it altogether, for folks that don't need/want it. If the change here will get you by until I figure that out, that's fine. If you could rebase on top of master latest, I'll go ahead and pull this in.

You don't need to merge this as it can be changed with a machine override `IMAGE_TEGRAFLASH_INITRD_FLASHER:tegra` and is what we do in our distro

madisongh:
> The issue is because of this snip that introduces this type of images in every recipe that uses INITRAMFS_FSTYPES.
If we change it to INITRAMFS_FSTYPES_TEGRA we can customize it be recipe instead of globally. Do you think it is a good approach?

Something like that should work.  I'll try and work that in with the changes for making initrd flashing optional.

quaresmajose:
> > The issue is because of this snip that introduces this type of images in every recipe that uses INITRAMFS_FSTYPES.
> > If we change it to INITRAMFS_FSTYPES_TEGRA we can customize it be recipe instead of globally. Do you think it is a good approach?
> 
> Something like that should work. I'll try and work that in with the changes for making initrd flashing optional.

I think that with the introduction of the INITRAMFS_FSTYPES_TEGRA I can disable the cboot image only on the recipes that I need so in this way I can build the kernel image with the default bsp layer types and this PR is not need any more.

I will prepare a new PR with this INITRAMFS_FSTYPES_TEGRA and adapting our distro to use this approach.

Thanks for your help

madisongh:
@quaresmajose Take a look at #1140 to see if that covers what you're asking for here.

quaresmajose:
https://github.com/OE4T/meta-tegra/pull/1140 look perfect and it should address all the issues I mention on this PR, it made my job a lot easier.

Many thanks for you quick response.

PR #1140, opened 2023-01-17
Make initrd-flash support more configurable

madisongh:
Clean up some global settings so that distros/users that want to use an alternative initrd-flash image, or exclude that support altogether to cut down on build times, can do so.

`TEGRAFLASH_INITRD_FLASH_IMAGE` can be set to the null string to exclude support, or to the name of another recipe.

Also refactor some more common settings to `tegra-common.inc`, and  avoid direct appends to `INITRAMFS_IMAGE_FSTYPES`, allowing users to specify alternative types for their initramfs builds by setting `TEGRA_INITRAMFS_FSTYPES` (and `TEGRA_INITRD_FLASH_INITRAMFS_FSTYPES` for the initrd flasher).


quaresmajose:
This should solve a couple of issues and with it in place we can chose the initrd types for other images that may exist and use the one specified in this layer for flashing

ichergui:
Thanks @madisongh. 
LGTM

ichergui (approved):

PR #1144, opened 2023-01-26
WIP branch for Jetpack 5.1/L4T R35.2.1

ichergui:


kekiefer:
I guess that libnvgbm was dropped from the bsp for this release? The tegra-udrm-gbm does need it, so it may require an update to get what it needs if there won't be a libnvgbm. It does seem that tegra-libraries-gbm-backend does not need it.

ichergui:
> I guess that libnvgbm was dropped from the bsp for this release? The tegra-udrm-gbm does need it, so it may require an update to get what it needs if there won't be a libnvgbm. It does seem that tegra-libraries-gbm-backend does not need it.

That's correct, I noticed that as well. 
I started with `demo-image-full` and tried to fix the issues at build time.
I will check definitely the requirement of some recipes regarding `libnvgbm` and will keep you updated.

PR #1155, opened 2023-01-30
tegra-udrm-gbm: update to version 1.1.0

kekiefer:
This update is required due to the removal of nvgbm in 35.2.1. It replaces the nvgbm wrapper with its own buffer allocation using nvbuffersurface.


ichergui:
Thanks @kekiefer for the updates.

ichergui (approved):
LGTM

PR #1156, opened 2023-01-30
tegra-minimal-initramfs: Bind DTB/DTBO extra dependencies via DTB_EXT…

dchvs:
…RA_DEPS

When installing new DTBOs with OVERLAY_DTB_FILE:append = ",some-overlay.dtbo", you will find that do_image_tegraflash sorts intertasks dependencies with do_image_tegraflash[depends] += "${DTB_EXTRA_DEPS}". This is not the case with tegra-minimal-initramfs, and it actually collects the DTBOs into OVERLAY_DTB_FILE the same way as do_image_tegraflash does. With this dependency definition we tell it to wait for a given dependency's do_deploy definition. User will do DTB_EXTRA_DEPS:append = " some-overlay-recipe:do_deploy" so it works sucessfully.

Signed-off-by: Daniel Chaves <dchvs11@gmail.com>

### DIFF
--- a/recipes-core/images/tegra-minimal-initramfs.bb
+++ b/recipes-core/images/tegra-minimal-initramfs.bb
@@ -13,6 +13,9 @@ PACKAGE_INSTALL = "\
     ${TEGRA_INITRD_INSTALL} \
 "
 
+# Bind DTB/DTBO extra dependencies
+do_rootfs[depends] += " ${DTB_EXTRA_DEPS}"
+
 IMAGE_FEATURES = ""
 IMAGE_LINGUAS = ""
 


madisongh:
Can you describe the problem you're seeing a bit more?  For the initramfs recipe, the overlays in OVERLAY_DTB_FILE should only be getting applied during BUP generation, and the DTB_EXTRA_DEPS variable is already present in CONVERSION_DEPENDS_bup_payload to handle that case.

dchvs:
The tegra-minimal-initramfs' do_image_cpio task failed to find a DTBO that comes from a different provider other than the Kernel (another recipe). Added the mentioned definitions to resolve:

```bash
@ .conf
DTB_EXTRA_DEPS:append = " some-overlay-recipe:do_deploy"

@ tegra-minimal-initramfs.bbappend
do_rootfs[depends] += " ${DTB_EXTRA_DEPS}"
```

But testing again it seems that `DTB_EXTRA_DEPS:append = " some-overlay-recipe:do_deploy"` it's actually enough.

madisongh:
@dchvs That means the addition you proposed here is not needed, right?

dchvs:
Yes correct, please close the PR.

PR #1161, opened 2023-02-04
[wip-l4t-r35.2.1] Add support for UEFI secure boot

madisongh:
R35.2.1 adds support for signature validation of the kernel, initrd, and kernel DTB using UEFI secure boot.  Changes in this PR add support for enablement of UEFI secure boot and signing the binaries UEFI validates.

To enable secure boot:
* Generate the PK, KEK, and db keys as described in [the L4T documentation](https://docs.nvidia.com/jetson/archives/r35.2.1/DeveloperGuide/text/SD/Security/SecureBoot.html#prepare-the-pk-kek-db-keys), and run the `gen_uefi_default_keys_dts.sh` described in that section to generate the `UefiDefaultSecurityKeys.dts` file used for enrolling those keys during boot. (Only 1 db key is required.)
* Add a bbappend to the `tegra-uefi-keys-dtb` recipe to supply the `UefiDefaultSecurityKeys.dts` file during the build.

Note that you can enable UEFI secure boot on Jetson targets that do not have secure boot fuses burned, to test it out.

To test signing UEFI-verified artifacts during the build, add these lines to your `local.conf`:
```
TEGRA_UEFI_DB_KEY = "/path/to/your/db.key"
TEGRA_UEFI_DB_CERT = "/path/to/your/db.crt"
```
When both of these variables are set, signing functions will get called to generate the signatures in the way that NVIDIA's UEFI code (or L4TLauncher app) expects.  Both partition-resident kernel/DTB and L4TLauncher's `extlinux.conf` setup are supported.

(Note that you don't need to make the PK or KEK key files or cert files available during the build for signing.)

**Not** supported is postponing UEFI signing to flashing time.  That might be possible to add later, at least for a subset of use cases (e.g., when not using `extlinux.conf`).

Other changes in this PR include:
* Some cleanup for L4T DTB overlay handling, to better separate the overlays needed for bootloader control from the plugin-manager overlays used to modify the kernel DTB based on boot-time hardware discovery.  You can still use the `OVERLAY_DTB_FILE` variable to add more overlays, as before, but I'd like to deprecate that at some point.
* Some `extlinux.conf` cleanup, based on the implementation in the L4TLauncher app that comes with NVIDIA's UEFI.


recipes-kernel/linux/linux-tegra_5.10.bb
```diff
@@ -97,6 +97,35 @@ do_apply_devicetree_overlays() {
 do_apply_devicetree_overlays[dirs] = "${B}"
 do_apply_devicetree_overlays[depends] += "dtc-native:do_populate_sysroot"
 
+addtask apply_devicetree_overlays after do_compile_devicetree_overlays before do_install
+
+# Override this function in a bbappend to
+# implement other signing mechanisms
+sign_kernel_image_and_dtb_files() {
+    if [ -n "${TEGRA_UEFI_DB_KEY}" -a "${TEGRA_UEFI_DB_CERT}" ]; then
```
ichergui, line 105: I think `-n` is missing here

recipes-kernel/linux/linux-tegra_5.10.bb
```diff
@@ -111,9 +140,30 @@ do_deploy:append() {
 			install -m 0644 ${KERNEL_OUTPUT_DIR}/dts/$dtbo $deployDir
 		fi
 	done
+	if [ -n "${TEGRA_UEFI_DB_KEY}" -a "${TEGRA_UEFI_DB_CERT}" ]; then
```
ichergui, line 143: I think `-n` is missing here as well

recipes-bsp/uefi/edk2-firmware-tegra_35.2.1.bb
```diff
@@ -43,6 +43,22 @@ do_compile:append() {
 }
 do_compile[depends] += "${NVDISPLAY_INIT_DEPS}"
 
+# Override this function in a bbappend to
+# implement other signing mechanisms
+sign_efi_app() {
+    if [ -n "${TEGRA_UEFI_DB_KEY}" -a "${TEGRA_UEFI_DB_CERT}" ]; then
```
ichergui, line 49: I think `-n` is missing here

recipes-bsp/tegra-binaries/tegra-uefi-prebuilt_35.2.1.bb
```diff
@@ -9,26 +9,44 @@ PROVIDES = "virtual/bootloader standalone-mm-optee-tegra"
...
 }
 
 do_compile:append:tegra234() {
     cp ${S}/bootloader/standalonemm_optee_t234.bin ${B}/standalone_mm_optee.bin
 }
 
+# Override this function in a bbappend to
+# implement other signing mechanisms
+sign_efi_app() {
+    if [ -n "${TEGRA_UEFI_DB_KEY}" -a "${TEGRA_UEFI_DB_CERT}" ]; then
```
ichergui, line 34: I think `-n` is missing here

recipes-kernel/linux/linux-tegra_5.10.bb
madisongh: Yep, thanks, will fix

ichergui (approved):
LGTM

madisongh:
Wiki page is [here](https://github.com/OE4T/meta-tegra/wiki/Secure-Boot-Support-in-L4T-R35.2.1).

PR #1174, opened 2023-02-14
edk2-firmware-tegra: fix patch for customer eeprom handling

kekiefer:
In the port of this patch, the customer prefix was erroneously dropped from the population of customer board ids in the chosen devicetree node:

    # cat /sys/firmware/devicetree/base/chosen/ids
    2888-0004-400 0123-123

Should read as:

    # cat /sys/firmware/devicetree/base/chosen/ids
    2888-0004-400 ABCD-0123-123

The master branch of edk2-firmware contains a rewritten version that works correctly, so update to this version. A PR for update to 35.2.1 happened upstream around the time of the release:

https://github.com/NVIDIA/edk2-nvidia/pull/34


ichergui (approved):

ichergui:
Thanks @kekiefer

PR #1186, opened 2023-02-24
python3-pybind11: fix bug with casts to void *

kekiefer:
Building python3-tensorrt fails with:

error: void value not ignored as it ought to be
|   219 |             PYBIND11_OVERLOAD_PURE_NAME(
|       |             ^~~~~~~~~~~~~~~~~~~~~~~~~~~

This bug is fixed upstream in subsequent versions, patch is applied by this commit.

### DIFF
--- a/external/openembedded-layer/recipes-devtools/python/python3-pybind11_2.10.0.bbappend
+++ b/external/openembedded-layer/recipes-devtools/python/python3-pybind11_2.10.0.bbappend
@@ -1,0 +1,4 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+SRC_URI:append:tegra = " file://0001-Fix-casts-to-void-4275.patch"
+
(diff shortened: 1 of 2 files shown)


ichergui (approved):
LGTM. Thanks @kekiefer

madisongh:
Even though this passed the layer check, we can't be making changes to how recipes from other layers build for everyone, even if this is a good change for everyone.  Making the patch and package arch to be tegra-specific for tegra builds would be OK (see the haveged bbappend for an example).  Even better would be to submit a patch to openembedded-devel to update the meta-python recipe.

kekiefer:
Thanks, that's right. I updated it to just apply to tegra.

I feel like this is gray area since the problem is exposed by the bindings we require in this layer, at least it won't need to be carried forward.

external/openembedded-layer/recipes-devtools/python/python3-pybind11_2.10.0.bbappend
```diff
@@ -0,0 +1,4 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+SRC_URI:append:tegra = " file://0001-Fix-casts-to-void-4275.patch"
+
```
madisongh, line 4: Add
```
```
and that should do it.

madisongh (changes requested):

kekiefer:
In retrospect since this is actually for the native version, it's too awkward to do this here.

PR #1200, opened 2023-03-09
Update find-jetson-usb.sh to support usb hubs

akhilveeraghanta:
Currently, the `find-jetson-usb` script only looks for one level of usb devices, which doesn't let you use a usb hub

all the devices are symlinked automatically to the top level directory under `/sys/bus/usb/devices`, so we can just look at those to support direct connections and usb hubs

Not sure if this the right branch to target


madisongh:
NVIDIA recommends against putting a USB hub between the host and the Jetson for flashing, but I suspect that's more of an issue with direct flashing than it is for initrd flashing, where we use this script, so this should be OK.  Thanks!

PR #1202, opened 2023-03-14
edk2-firmware-tegra: use bfd linker

sapiippo:
Build fails if using gold linker with DISTRO_FEATURE ld-is-gold. Add compiler flags to always use bfd linker.


ichergui:
Hey @sapiippo 
Thanks for the PR

ichergui (approved):
LGTM

PR #1206, opened 2023-03-17
nvidia-docker: update branch name

ichergui:


madisongh:
This change doesn't look like it's needed until we update to v2.12.0, or unless upstream actually deletes the old branch from their repo.

ichergui:
I will close this PR since it is not needed for now

PR #1220, opened 2023-04-05
update initrd script for auto mounting

ssuesens:
removed patch (initrd-flash: removed space before tab in indent)

initrd-flash: fix patch for partition mounting and unmounting
On some systems auto mount is faster than udisksctl.
Therefore udisksctl mount command fails with the error message,
that the partition can not be mounted because partition is already mounted.
Is the partition auto mounted, udisksctl can not unmount the partition.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra194-flash-helper.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra194-flash-helper.sh
@@ -593,7 +593,7 @@ if [ $bup_blob -ne 0 ]; then
     TBCFILE="uefi_jetson.bin"
     TOSFILE="tos-optee_t194.img"
     . "$here/l4t_bup_gen.func"
-    spec="${BOARDID}-${FAB}-${BOARDSKU}-${BOARDREV}-1-${CHIPREV}-${MACHINE}-${BOOTDEV}"
+    spec="${BOARDID}-${FAB}-${BOARDSKU}-${BOARDREV}-1-${CHIPREV}-${MACHINE}-"
     if [ $(expr length "$spec") -ge 128 ]; then
 	echo "ERR: TNSPEC must be shorter than 128 characters: $spec" >&2
 	exit 1
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra234-flash-helper.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra234-flash-helper.sh
@@ -672,7 +672,7 @@ if [ $bup_blob -ne 0 ]; then
     bpfdtbfilename="$BPFDTB_FILE"
     localbootfile="boot.img"
     . "$here/l4t_bup_gen.func"
-    spec="${BOARDID}-${FAB}-${BOARDSKU}-${BOARDREV}-1-${CHIPREV}-${MACHINE}-${BOOTDEV}"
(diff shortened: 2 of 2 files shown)


ichergui:
Thanks @ssuesens for this PR 
I think the first commit is not needed (removing space before tab) also please rebase your branch to avoid the Merge branch commit.
Only one commit needed for the review

ssuesens:
initrd-flash: removed space before tab in indent

initrd-flash: fix patch for partition mounting and unmounting
On some systems automount is faster than udisksctl.
Therefore udisksctl mount command fails with the error message,
that the partition can not be mounted because partition is already mounted.
Is the partition automounted, udisksctl can not unmount the partition.

ssuesens:
The initrd-flash.sh mixed tab and space. I think we should use only tab or space, but do not mixed it.

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -220,21 +220,40 @@ run_rcm_boot() {
...
 
 unmount_and_release() {
-    local mnt="$1"
-    local dev="$2"
+    local dev="$1"
     local remain=3
-    if [ -n "$mnt" ]; then
-	udisksctl unmount --force -b "$dev"
-    fi
+    while true; do
```
ichergui, line 244: Could you please avoid using `while true` ?

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -220,21 +220,40 @@ run_rcm_boot() {
...
+        local mnt=$(cat /proc/mounts | grep "^$dev" | cut -d' ' -f2)
+        local mnt=$(echo ${mnt} | cut -d' ' -f1)
+        if [ -n "${mnt}" ]; then
+            umount "${mnt}" > /dev/null
+            if [ "$?" -eq 1 ]; then
```
ichergui, line 249: I prefer to use `if [ ! "$?" -eq 0 ]` instead 
Please refer to the documentation: https://man7.org/linux/man-pages/man2/umount.2.html

ichergui (changes requested):

ichergui:
> The initrd-flash.sh mixed tab and space. I think we should use only tab or space, but do not mixed it.

I understand. My attention is make the review easy and track the changes quickly

madisongh:
One of the prerequisites mentioned [on the wiki page](https://github.com/OE4T/meta-tegra/wiki/initrd-flashing-support#prerequisites) is to disable automounting of removable media.  If you do those steps, is this patch still necessary?  If so, can you provide more info on the distro, version, etc., you're using?

I'll add some specific comments about the patch, but I'd like to understand the problem a bit better before going into too much detail on the specific solution.

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
madisongh: In keeping with usage in the rest of the script, using `if ! umount "${mnt}" > /dev/null 2>&1; then` would be even better.

madisongh (changes requested):
A few specific comments here, but as I mention elsewhere, I'd like to understand better what the problem is that you're trying to solve.

Also, please add a Signed-off-by line in the commit message, per our [contribution guidelines](https://github.com/OE4T/meta-tegra/blob/master/CONTRIBUTING.md).

Thanks!

PR #1225, opened 2023-04-09
JetPack 5.1.1/L4T R35.3.1 for kirkstone

madisongh:


ibstewart:
Thanks Matt. I tested this with my IGX Orin devkit today with both iGPU (i.e. the L4T drivers) and dGPU and didn't encounter any issues (besides a build error with the dGPU kernel module due to the disabled module signing, but I think that's specific to the dGPU driver build and it was easy enough to fix by enabling the module signing again).

PR #1232, opened 2023-04-14
tegra-flashtools-native: apply secureboot overlay

madisongh:
This applies the secureboot_overlay_35.2.1.tbz2 update to the signing tools that enables secure boot on Orin platforms.


ichergui (approved):
LGTM. thanks @madisongh

PR #1233, opened 2023-04-14
tegra-flashtools-native: install tegrasign_v3* files

ibstewart:
This fixes an issue where the tegrasign_v3_oemkey.yaml file was not being installed by the existing 'tegrasign_v3*py' install rule.


ibstewart:
@ichergui That patch copies the file from the patch into ${S} but it doesn't actually get installed into ${D}${BINDIR}.

ichergui (approved):
Thanks @ibstewart 
I missed that from the previous PR.

PR #1236, opened 2023-04-18
initrd-flash: fix auto mount trouble

ssuesens:
Disable automount feature on OS is not longer required.


madisongh:
Thanks... I'll still recommend that folks disable automount, but this will help in situations where they can't for some reason.

PR #1240, opened 2023-04-26
Add machine for jetson-orin-nano-devkit-with-nx-module

ssuesens:
Added machine configuration and flashvar settings for jetson orin nano devkit with Orin NX module.


ssuesens:
Hello,

is it possible to integrate this patch into the kirkstone branch?

madisongh:
Thanks for this... I've manually merged this into master, with some edits, mainly to use a shorter MACHINE name.  We have to keep the name length shorter than 32 characters (for NVIDIA's bootloader update payloads), so I renamed the machine to `p3768-0000-p3767-0000`.  It's less user-friendly, but more compact.

After some testing, I've also back-ported the change to mickledore and kirkstone.

PR #1244, opened 2023-04-30
Refactor tegra-libraries-multimedia

### LINKED ISSUE #1243: Enable NVMM support for `gstreamer1.0-plugins-base >= 1.20`
**Describe the bug**
In `gstreamer1.0-plugins-base >= 1.20`, the GStreamer elements `glupload` and `gldownload` has direct support for the tegra `NVMM` type buffers.
This means that one can convert to and from `OpenGL` buffers and `NVMM` without doing a full buffer copy, which is critical for not tanking performance in high-resolution live video applications.

Currently, there are some challenges to getting this support enabled though.

Going through the GStreamer build system, one needs to get the `HAVE_NVMM` define set up.
Reference: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-base/ext/gl/gstgldownloadelement.c#L37

To get this define set up, one needs OpenGL and EGL support, which is already supported through the PACKAGECONFIG system :tada:
_But_, one also needs Nvidia-specific proprietary headers available during the build phase of `gstreamer1.0-plugins-base`, namely `nvbuf_utils.h`.
Reference: https://gitlab.freedesktop.org/gstreamer/gstreamer/-/blob/main/subprojects/gst-plugins-base/gst-libs/gst/gl/meson.build#L958

This header is supplied by `tegra-mmapi`.
I tried letting `gstreamer1.0-plugins-base` DEPEND on `tegra-mmapi`, but that led to a dependency cycle.
The cycle seems to be the following:
`tegra-mmapi` `DEPENS` on `tegra-libraries-multimedia` which `DEPENDS` on `gstreamer1.0-plugins-base`.

I believe this cycle is not really needed, I think `tegra-mmapi` should be possible to build without any reference to `gstreamer1.0-plugins-base`. But I do not know the best way to break this cycle.

I would be very thankful for any insight that could lead to fixing this.

I am basing my conclusions on this revision of meta-tegra: `9d58efcf`

Please tell me if there is any more information I can supply to help with this issue.

madisongh:
Creates a new recipe to install the `libnvdsbufferpool.so` library so that the main `tegra-libraries-multimedia` recipe doesn't have to depend on `gstreamer1.0-plugins-base`.

Fixes #1243

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-libraries-multimedia-ds_35.3.1.bb
+++ b/recipes-bsp/tegra-binaries/tegra-libraries-multimedia-ds_35.3.1.bb
@@ -1,0 +1,29 @@
+L4T_DEB_COPYRIGHT_MD5 = "3d9212d4d5911fa3200298cd55ed6621"
+DEPENDS = "tegra-libraries-multimedia glib-2.0 gstreamer1.0-plugins-base"
+
+L4T_DEB_TRANSLATED_BPN = "nvidia-l4t-multimedia"
+
+require tegra-debian-libraries-common.inc
+
+LICENSE += "& MIT & BSD-3-Clause"
+
+MAINSUM = "83dd039e6e790fd0aae4a1e68101ba1ac75b661038142e6ded826a35ff27376e"
+
+inherit features_check
+
+REQUIRED_DISTRO_FEATURES = "opengl"
+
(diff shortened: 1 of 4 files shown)


ichergui (approved):
LGTM 
Thanks @madisongh

PR #1250, opened 2023-05-07
tegra-helper-scripts: fix partition type setting in make-sdcard script

madisongh:
Fix the make_partitions function in make-sdcard so that it honors the partition type GUID value coming from the flash layout for any partition, not just the final one.

This fixes an issue with the ESP partition not getting recognized by UEFI as the boot partition.


madisongh:
See discussion #1239

PR #1265, opened 2023-06-05
initrd-flash: Add doflash dry run to retrieve board specs

dchvs:
Ensure that initrd-flash meets the board's specs parameters during flashing.


dchvs:
There is an issue with the flashing process in the script since it does not contain the necessary board specification variables. An example of this issue can be observed in this specific line of the script: https://github.com/OE4T/meta-tegra/blob/master/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh#L159.

ichergui:
Hello @dchvs 
I tried flashing my devices via `initrd-flash` script and it was working fine to me.
Could you please create an issue and share the logs ?
Thanks.

PR #1268, opened 2023-06-08
Add support for Orin Nano 4GB Variant

joekale-pp:
Set to draft while initial testing is being completed so people know it's being worked on.


joekale-pp:
`nvfancontrol` seems to behave appropriately when running s process to heat up the processor. Values look backwards in the deployed conf, compared to examples in the developer documentation, but after messing with them some they seem to be correct. 

`nvpmodel` works to change from the default mode of 10W (0) to a lower mode 7W_AI or 7W_CPU, but there seems to be an issue changing back where the command throws an error.
```
NVPM ERROR: Error writing 2 to /sys/devices/gpu.0/fbp_pg_mask: 19
NVPM ERROR: failed to write PARAM FBP_POWER_GATING: ARG FBP_PG_MASK: PATH: /sys/devices/gpu.0/fbp_pg_mask VAL: 2
NVPM ERROR: failed to set power mode!
NVPM ERROR: optMask is 1, no request for power mode
```

If these issues seem unrelated to basic device support then I can mark this ready unless there are other items you would want me to check on.

madisongh (changes requested):
Would you mind changing the name of this to `p3768-0000-p3767-0004.conf`?  We do have to deviate a bit from the naming that NVIDIA uses (because we need to know the specific module SKU at build time), but using the carrier-module combination is less apt to cause problems, should NVIDIA release an actual Nano 4G dev kit at some point in the future.

Otherwise, LGTM

madisongh:
> If these issues seem unrelated to basic device support then I can mark this ready unless there are other items you would want me to check on.

It's minor enough; there's a minor issue with `nvpmodel` with the Orin NX module, too.  Let's go ahead and get this in and we can address it later.

madisongh (approved):

PR #1273, opened 2023-06-11
tegra-binaries: inherit dos2unix

liuming50:
There are some source files in tegra-binaries using CRLF as line terminator, this is hard for developer to make their own patches on it.

Fix it by inheritting dos2unix.


madisongh:
I've looked at that bbclass, and I'm a bit worried that using this might damage some of the binary files, since it applies `dos2unix` to everything in the tree.  I'd be OK with adding a new task to the `tegra-binaries` recipe that is more selective - it looks like it's mainly the `.cfg` and `.dtsi` files that have the DOS-style line termination.

liuming50:
@madisongh good suggestion!

Changes in patch set V2:

1 Only convert "*.cfg", "*dts*" files with dos2unix as suggested by @madisongh

recipes-bsp/tegra-binaries/tegra-binaries-35.3.1.inc
```diff
@@ -10,7 +10,12 @@ SRC_URI = "\
 SRC_URI[sha256sum] = "80a55504a2ce9cdc0a328edb6fd0690615c4ffde9c2b32f4e64e0a1a3432a2e2"
 SRC_URI[xusb.sha256sum] = "89ce71eb7d2b0df84346635d72c805e2b6832aa0e2b00360bec46d0aa0a6f40c"
 
-inherit l4t_bsp
+inherit l4t_bsp dos2unix
+
+# Override convert_crlf_to_lf task to only convert "*.cfg", "*.dts*" files
+do_convert_crlf_to_lf () {
+	find ${S} -type f -name "*.cfg" -o -name "*.dts*" -exec dos2unix {} \;
```
madisongh, line 17: The way `find` works, the ORed predicates need to be parenthesized, or else the `-exec` will only apply to the second one:
```
find ${S} -type f "-(" -name "*.cfg" -o -name "*.dts*" "-)" -exec dos2unix {} \;
```

liuming50: @madisongh thanks for the review!

liuming50:
Changes in patch set 3:

1 Adjusted find command as suggested by @madisongh

madisongh (approved):

PR #1285, opened 2023-06-30
Add support for building capsules

Lexmark-chad:
Consider this just a starting point to spur some discussion on how best to implement.

1.  Should the capsules be included in the rootfs by default or simply be deployed to `${DEPLOYDIR}`?  I initially went ahead and included them in the rootfs.  That follows the pattern of how `swupdate` works (which is what we use) in `tegra-test-distro` whereby `tegra-bootloader-update` and `bl_update_payload` are packaged into the rootfs and then swupdate mounts the just programmed inactive partition to drive the bootloader upddate.  Having said that, I believe when we implemented `dm-verity` to create a full chain of trust from the bootloaders to the rootfs it created a cycle of dependencies in the build.  To break the cycle we pulled the bootloader payload out of the rootfs and packaged it directly into the .swu.  For uefi the capsules eventually have to end up in `/opt/nvidia/esp/EFI/UpdateCapsule/`.  Maybe that means that the capsules should simply be deployed to the `${DEPLOYDIR}` and it is up to the update mechanism to pull them out of the update specific packaging and copy them to the appropriate directory.

2.  Is a new recipe really required?  I created a separate recipe instead of piling on top of `tegra-bup-payload`.  By doing so I needed to move some logic to a  .bbclass to avoid having to repeat it in the new recipe and creating a new point of maintenance.

3.  Requesting a capsule update is obscure so I created and a helper script to set `OSIndications`.

I like choosing good names for things so if there are any suggestions to improved naming on anything I am open to suggestions.

### DIFF
--- a/recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
+++ b/recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
@@ -1,0 +1,73 @@
+DESCRIPTION = "Generate UEFI capsules for bup paylods"
+LICENSE = "MIT"
+LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
+
+inherit l4t_bsp tegra-bup
+
+COMPATIBLE_MACHINE = "(tegra)"
+
+DEPENDS += "edk2-basetools-tegra-native tegra-bup-payload"
+
+GUID:tegra194 ?= "be3f5d68-7654-4ed2-838c-2a2faf901a78"
+GUID:tegra234 ?= "bf0d4599-20d4-414e-b2c5-3595b1cda402"
+
+def get_hex_bsp_version(bsp_version):
+    verparts = bsp_version.split('.')
+    return hex(int(verparts[0])<<16 | int(verparts[1])<<8 | int(verparts[2]))
+
+BSP_VERSION32 = "${@get_hex_bsp_version(d.getVar('L4T_VERSION'))}"
+
+PYTHON_BASETOOLS = "${RECIPE_SYSROOT_NATIVE}/usr/bin/edk2-BaseTools/Source/Python"
+
+UEFI_CAPSULE_SIGNER_PRIVATE_CERT ?= "${PYTHON_BASETOOLS}/Pkcs7Sign/TestCert.pem"
+UEFI_CAPSULE_OTHER_PUBLIC_CERT ?= "${PYTHON_BASETOOLS}/Pkcs7Sign/TestSub.pub.pem"
+UEFI_CAPSULE_TRUSTED_PUBLIC_CERT ?= "${PYTHON_BASETOOLS}/Pkcs7Sign/TestRoot.pub.pem"
+
+# Override this function to have a secure signing server
+# perform the capsule signing.
+sign_uefi_capsules() {
+    export PYTHONPATH="${PYTHONPATH}:${PYTHON_BASETOOLS}"
+    if [ -e ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.bl_only.bup-payload ]; then
+        python3 ${PYTHON_BASETOOLS}/Capsule/GenerateCapsule.py \
+            -v --encode --monotonic-count 1 \
+            --fw-version "${BSP_VERSION32}" \
+            --lsv "${BSP_VERSION32}" \
+            --guid "${GUID}" \
+            --signer-private-cert "${UEFI_CAPSULE_SIGNER_PRIVATE_CERT}" \
+            --other-public-cert "${UEFI_CAPSULE_OTHER_PUBLIC_CERT}" \
+            --trusted-public-cert "${UEFI_CAPSULE_TRUSTED_PUBLIC_CERT}" \
+            -o ./tegra-bl.cap \
+            ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.bl_only.bup-payload
+    fi
+    if [ -e ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.kernel_only.bup-payload ]; then
+        python3 ${PYTHON_BASETOOLS}/Capsule/GenerateCapsule.py \
+            -v --encode --monotonic-count 1 \
+            --fw-version "${BSP_VERSION32}" \
+            --lsv "${BSP_VERSION32}" \
+            --guid "${GUID}" \
+            --signer-private-cert "${UEFI_CAPSULE_SIGNER_PRIVATE_CERT}" \
+            --other-public-cert "${UEFI_CAPSULE_OTHER_PUBLIC_CERT}" \
+            --trusted-public-cert "${UEFI_CAPSULE_TRUSTED_PUBLIC_CERT}" \
+            -o ./tegra-kernel.cap \
+            ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.kernel_only.bup-payload
+    fi
+}
+
+do_compile() {
+    sign_uefi_capsules
+}
+
+CAPSULE_DIR = "/opt/nvidia/UpdateCapsule"
+
+do_install() {
+    install -d ${D}${CAPSULE_DIR}
+    if [ -e ${B}/tegra-bl.cap ]; then
+        install -m 0644 ${B}/tegra-bl.cap ${D}${CAPSULE_DIR}
+    fi
+    if [ -e ${B}/tegra-kernel.cap ]; then
+        install -m 0644 ${B}/tegra-kernel.cap ${D}${CAPSULE_DIR}
+    fi
+}
+
+FILES:${PN} += "${CAPSULE_DIR}"
+PACKAGE_ARCH = "${MACHINE_ARCH}"
--- a/classes/tegra-bup.bbclass
+++ b/classes/tegra-bup.bbclass
@@ -1,0 +1,13 @@
+inherit kernel-artifact-names
+
+def bupfile_basename(d):
+    if bb.utils.to_boolean(d.getVar('INITRAMFS_IMAGE_BUNDLE')):
+        return "${KERNEL_IMAGETYPE}-${INITRAMFS_LINK_NAME}"
+    return "${INITRAMFS_IMAGE}-${MACHINE}"
+
+def bup_dependency(d):
+    if bb.utils.to_boolean(d.getVar('INITRAMFS_IMAGE_BUNDLE')):
+        return "kernel-bup-payload:do_deploy"
+    return "${INITRAMFS_IMAGE}:do_image_complete"
+
+BUPFILENAME = "${@bupfile_basename(d)}"
--- a/conf/machine/jetson-xavier-nx-devkit-emmc.conf
+++ b/conf/machine/jetson-xavier-nx-devkit-emmc.conf
@@ -5,10 +5,7 @@
 
 TEGRA_BOARDSKU ?= "0001"
 TEGRA_BUPGEN_SPECS ?= "fab=100;boardsku=0001;boardrev= \
-		       fab=200;boardsku=0001;boardrev= \
-		       fab=300;boardsku=0001;boardrev= \
-		       fab=301;boardsku=0001;boardrev= \
-		       fab=301;boardsku=0003;boardrev="
+		       fab=301;boardsku=0001;boardrev="
 IMAGE_ROOTFS_ALIGNMENT ?= "4"
 
 require conf/machine/include/xavier-nx.inc
--- a/recipes-bsp/tegra-bup-payload/tegra-bup-payload_1.0.bb
+++ b/recipes-bsp/tegra-bup-payload/tegra-bup-payload_1.0.bb
@@ -4,19 +4,7 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 
 COMPATIBLE_MACHINE = "(tegra)"
 
-inherit kernel-artifact-names
-
-def bupfile_basename(d):
-    if bb.utils.to_boolean(d.getVar('INITRAMFS_IMAGE_BUNDLE')):
-        return "${KERNEL_IMAGETYPE}-${INITRAMFS_LINK_NAME}"
-    return "${INITRAMFS_IMAGE}-${MACHINE}"
-
-def bup_dependency(d):
-    if bb.utils.to_boolean(d.getVar('INITRAMFS_IMAGE_BUNDLE')):
-        return "kernel-bup-payload:do_deploy"
-    return "${INITRAMFS_IMAGE}:do_image_complete"
-
-BUPFILENAME = "${@bupfile_basename(d)}"
+inherit tegra-bup
 
 do_install() {
     install -d ${D}/opt/ota_package/
--- a/recipes-bsp/tools/setup-nv-boot-control/oe4t-set-uefi-OSIndications
+++ b/recipes-bsp/tools/setup-nv-boot-control/oe4t-set-uefi-OSIndications
@@ -1,0 +1,8 @@
+#!/bin/sh
+
+source /usr/bin/uefi_common.func
+
+rc=0
+value="\x04\x00\x00\x00\x00\x00\x00\x00"
+set_efi_var "OsIndications" "8be4df61-93ca-11d2-aa0d-00e098032b8c" "$value" || rc=1
+exit $rc
--- a/recipes-bsp/tools/setup-nv-boot-control/setup-nv-boot-control.sh.in
+++ b/recipes-bsp/tools/setup-nv-boot-control/setup-nv-boot-control.sh.in
@@ -1,132 +1,10 @@
 #!/bin/sh
 TARGET="@TARGET@"
 BOOTDEV="@BOOTDEV@"
-ESPMOUNT="@ESPMOUNT@"
-NVESPMOUNT="@NVIDIA_ESPMOUNT@"
-ESPVARDIR="@ESPVARDIR@"
 sysconfdir="@sysconfdir@"
 controlfile="${sysconfdir}/nv_boot_control.conf"
-nv_guid="781e084c-a330-417c-b678-38e696380cb9"
 
-fab=
-boardsku=
-boardrev=
-chiprev=
-mmc_only_hack=
-
-set_efi_var() {
-    local varname="$1"
-    local value="$2"
-
-    # Only set if it doesn't exist already
-    if [ "$mmc_only_hack" = "yes" ]; then
-	if ! mountpoint -q "$ESPMOUNT"; then
-	    echo "ERR: cannot store EFI variable - ESP partition not mounted" >&2
-	    return 1
-	fi
-	if [ ! -e "$ESPVARDIR/${varname}-${nv_guid}" ]; then
-	    local datatmp=$(TMPDIR=$RUNTIME_DIRECTORY mktemp --tmpdir nvcvar.XXXXXX)
-	    printf "\x07\x00\x00\x00%s" "$value" > "$datatmp"
-	    mkdir -p -m 0700 "$ESPVARDIR" || return 1
-	    cp "$datatmp" "$ESPVARDIR/${varname}-${nv_guid}" || return 1
-	fi
-    else
-	if efivar -n "${nv_guid}-$varname" >/dev/null 2>&1; then
-	    return 0
-	fi
-	local datatmp=$(TMPDIR=$RUNTIME_DIRECTORY mktemp --tmpdir nvcvar.XXXXXX)
-	printf "%s" "$value" > "$datatmp"
-	efivar -w -f "$datatmp" -n "${nv_guid}-$varname"
-    fi
-}
-
-gen_compat_2888() {
-    if [ "$fab" = "400" ]; then
-	if [ "$boardsku" = "0004" ]; then
-	    boardrev=
-	else
-	    if echo "$boardrev" | grep -q "^[ABCD]\."; then
-		boardrev="D.0"
-	    else
-		boardrev="E.0"
-	    fi
-	    boardsku="0001"
-	fi
-    elif [ "$fab" = "600" -a "$boardsku" = "0008" ]; then
-	boardrev=
-    fi
-    return 0
-}
-
-gen_compat_3668() {
-    if [ "$fab" != "301" ]; then
-	fab="100"
-    fi
-    boardsku=
-    boardrev=
-    chiprev=
-    return 0
-}
-
-gen_compat_3701() {
-    boardrev=
-    if [ "$boardsku" = "0000" -o "$boardsku" = "0004" ]; then
-	if echo "$fab" | egrep -q "^([012][0-9][0-9]|TS|EB)"; then
-	    fab="000"
-	else
-	    fab="300"
-	fi
-    fi
-    if [ "$boardksu" = "0005" ]; then
-	fab=""
-    fi
-    boardrev=
-    chiprev=
-    return 0
-}
-
-gen_compat_3767() {
-    if [ "$boardsku" = "0000" -o "$boardsku" = "0002" ]; then
-	if ! echo "$fab" | egrep -q "^(TS|EB)"; then
-	    fab="000"
-	fi
-    else
-	fab=
-    fi
-    boardrev=
-    chiprev=
-    return 0
-}
-
-# boardspec should be piped into this function
-gen_compat_spec() {
-    IFS=- read boardid fab boardsku boardrev fuselevel chiprev
-    if gen_compat_$boardid 2>/dev/null; then
-	echo "$boardid-$fab-$boardsku-$boardrev-$fuselevel-$chiprev"
-	return 0
-    fi
-    echo ""
-}
-
-# boardspec should be piped into this function
-needs_mmc_hack() {
-    IFS=- read boardid fab boardsku boardrev fuselevel chiprev
-    if [ "$boardid" = "2888" -a "$fab" = "400" ]; then
-	echo "yes"
-    else
-	echo ""
-    fi
-    return 0
-}
-
-# NVIDIA tools will try to mount the ESP partition dynamically
-# if it's not already mounted at @NVIDIA_ESPMOUNT@, so bind-mount
-# it there if we've mounted that partition at a more typical location.
-if [ -n "$NVESPMOUNT" -a "$NVESPMOUNT" != "$ESPMOUNT" ]; then
-    if mountpoint -q "$ESPMOUNT" && ! mountpoint -q "$NVESPMOUNT"; then
-	mount --bind "$ESPMOUNT" "$NVESPMOUNT"
-    fi
-fi
+source /usr/bin/uefi_common.func
 
 [ ! -e "$controlfile" ] || exit 0
 
@@ -141,14 +19,13 @@ if [ -z "${boardspec}" ]; then
     exit 1
 fi
 rc=0
-mmc_only_hack=$(echo "$boardspec" | needs_mmc_hack)
-set_efi_var "TegraPlatformSpec" "${boardspec}-${TARGET}-" || rc=1
+set_efi_var "TegraPlatformSpec" "781e084c-a330-417c-b678-38e696380cb9" "${boardspec}-${TARGET}-" "write-once" || rc=1
 compatspec=$(echo "$boardspec" | gen_compat_spec)
 if [ -z "$compatspec" ]; then
     compatsed="-e/^COMPATIBLE_SPEC/d"
 else
     compatsed="-es,@COMPATIBLE_SPEC@,${compatspec}-${TARGET}-,"
-    set_efi_var "TegraPlatformCompatSpec" "${compatspec}-${TARGET}-" || rc=1
+    set_efi_var "TegraPlatformCompatSpec" "781e084c-a330-417c-b678-38e696380cb9" "${compatspec}-${TARGET}-" "write-once" || rc=1
 fi
 sed -e"s,@TNSPEC@,${boardspec}-${TARGET}-${BOOTDEV}," $compatsed \
     "${sysconfdir}/nv_boot_control.template" > "$controlfile" || rc=1
--- a/recipes-bsp/tools/setup-nv-boot-control/uefi_common.func.in
+++ b/recipes-bsp/tools/setup-nv-boot-control/uefi_common.func.in
@@ -1,0 +1,139 @@
+#!/bin/sh
+ESPMOUNT="@ESPMOUNT@"
+NVESPMOUNT="@NVIDIA_ESPMOUNT@"
+ESPVARDIR="@ESPVARDIR@"
+
+fab=
+boardsku=
+boardrev=
+chiprev=
+mmc_only_hack=
+
+set_efi_var() {
+    local varname="$1"
+    local guid="$2"
+    local value="$3"
+    local write_once="$4"
+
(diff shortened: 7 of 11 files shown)


recipes-bsp/uefi/tegra-uefi-capsules/oe4t-set-uefi-OSIndications
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+printf "\x07\x00\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00" > /tmp/var_tmp.bin
+dd if=/tmp/var_tmp.bin of=/sys/firmware/efi/efivars/OsIndications-8be4df61-93ca-11d2-aa0d-00e098032b8c bs=12;sync
```
madisongh, line 4: The destination path has to be different for the AGX Xavier, which doesn't support writing variables in the efivars sysfs.

It might be better to move this script to the `setup-nv-boot-control` recipe and have it leverage the logic in the existing script there (maybe with a bit of refactoring) to determine how to modify the EFI variable.

recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
```diff
@@ -0,0 +1,31 @@
...
+
+SRC_URI += "file://oe4t-set-uefi-OSIndications"
+
+inherit bup
+
+COMPATIBLE_MACHINE = "(tegra)"
+
+DEPENDS += "openssl-native tegra-flashtools-native tegra-bup-payload"
+
+L4T_DIR = "${TMPDIR}/work-shared/L4T-native-${PV}-${PR}/Linux_for_Tegra"
```
madisongh, line 13: Rather than patching the script and running it out of the shared work directory, I'd prefer moving the capsule generation logic directly into the recipe and using the `GenerateCapsule.py` tool from the EDK2 sources, as mentioned in the comments in the script.  Last I looked, the copy of that tool in the L4T kit was unmodified from the original.

We'll also need a way for the signing arguments to get filled in based on some variable settings, and a hook for calling out to a signing server.

madisongh (changes requested):

madisongh:
Thanks @Lexmark-chad, this is a good start.

> Should the capsules be included in the rootfs by default or simply be deployed to ${DEPLOYDIR}?

`${DEPLOY_DIR_IMAGE}` might be a better place when creating the capsules.  Something in the OTA updater is going to have to write the capsule into the `/boot/efi/EFI/UpdateCapsule` directory.  For Mender updates, that might be a post-install state script that copies the capsule file out of the rootfs, so we'd also need a recipe to install the capsule(s) into the rootfs for that case.  For swupdate, the capsule(s) could just be another payload in the `.swu` file, keeping it out of the rootfs, in which case the recipe could pull the capsule(s) directly from the deploy area, saving space in the rootfs.

> Is a new recipe really required? I created a separate recipe instead of piling on top of tegra-bup-payload

Probably new recipes/bbclass/whatever for now.  At some point we might be able to combine the BUP and capsule generation logic, together, eliminating the `.bup-payload` artifacts altogether and just outputting capsules.  That should simplify things a bit, since the capsules won't combine the bootloader and kernel into a single payload.  That should allow us to move some of the logic into a bbclass or include file and have cleaner, separated, paths for the bootloader and the kernel capsule generation.  @kekiefer - what do you think?   Could we drop the old BUPs and associated tools and switch over to the capsule method in one step?

recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
Lexmark-chad: That makes sense.  I'll try to get started on that this week.

Lexmark-chad: Should `GenerateCapsule.py` and dependent modules be packaged into a `tegra-uefi-capsule-native` package?  I started down a different path with the patch below, but it feels 'dirty' copying over all this into `${DEPLOY_DIR_IMAGE}`.

```
diff --git a/recipes-bsp/uefi/edk2-firmware-tegra_35.3.1.bb b/recipes-bsp/uefi/edk2-firmware-tegra_35.3.1.bb
index 41ac23d2..ae6ec280 100644
--- a/recipes-bsp/uefi/edk2-firmware-tegra_35.3.1.bb
+++ b/recipes-bsp/uefi/edk2-firmware-tegra_35.3.1.bb
@@ -75,6 +75,9 @@ do_deploy() {
        [ -e ${B}/images/$dtb ] || continue
        install -m 0644 ${B}/images/$dtb ${DEPLOYDIR}/
     done
+    install -d ${DEPLOYDIR}/Common
+    install -m 0644 ${S}/BaseTools/Source/Python/Capsule/GenerateCapsule.py ${DEPLOYDIR}/
+    cp -R --preserve=mode,timestamps ${S}/BaseTools/Source/Python/Common/* ${DEPLOYDIR}/Common
 }
```

This part is still a WIP w/r/t the signing certificates, but more context on where I was headed:

```
diff --git a/recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb b/recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
index c499143b..ab779f7d 100644
--- a/recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
+++ b/recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
@@ -4,17 +4,36 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 
 SRC_URI += "file://oe4t-set-uefi-OSIndications"
 
-inherit bup
+inherit bup l4t_bsp
 
-DEPENDS += "openssl-native tegra-flashtools-native tegra-bup-payload"
+DEPENDS += "openssl-native edk2-firmware-tegra tegra-bup-payload"
 
-L4T_DIR = "${TMPDIR}/work-shared/L4T-native-${PV}-${PR}/Linux_for_Tegra"
+GUID:tegra194 = "be3f5d68-7654-4ed2-838c-2a2faf901a78"
+GUID:tegra234 = "bf0d4599-20d4-414e-b2c5-3595b1cda402"
+
+def get_hex_bsp_version(bsp_version):
+    verparts = bsp_version.split('.')
+    return hex(int(verparts[0])<<16 | int(verparts[1])<<8 | int(verparts[2]))
+
+BSP_VERSION32 = "${@get_hex_bsp_version(d.getVar('L4T_VERSION'))}"
 
 do_compile() {
-    ${L4T_DIR}/generate_capsule/l4t_generate_soc_capsule.sh -i ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.bl_only.bup-payload -o ./tegra-bl.cap ${L4T_DEB_SOCNAME}
-    ${L4T_DIR}/generate_capsule/l4t_generate_soc_capsule.sh -i ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.kernel_only.bup-payload -o ./tegra-kernel.cap ${L4T_DEB_SOCNAME}
+    python3 ${DEPLOY_DIR_IMAGE}/GenerateCapsule.py \
+        -v --encode --monotonic-count 1 \
+        --fw-version "${BSP_VERSION32}" \
+        --lsv "${BSP_VERSION32}" \
+        --guid "${GUID}" \
+        -o ./tegra-bl.cap \
+        ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.bl_only.bup-payload
+    python3 ${DEPLOY_DIR_IMAGE}/GenerateCapsule.py \
+        -v --encode --monotonic-count 1 \
+        --fw-version "${BSP_VERSION32}" \
+        --lsv "${BSP_VERSION32}" \
+        --guid "${GUID}" \
+        -o ./tegra-kernel.cap \
+        ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.kernel_only.bup-payload
 }
```

madisongh: @Lexmark-chad As we discussed in the monthly meeting, I think it would make sense to create a recipes for extracting the necessary tools/scripts from EDK2.  Borrowing/adapting [this recipe](https://git.yoctoproject.org/meta-arm/tree/meta-arm/recipes-bsp/uefi/edk2-basetools-native_202302.bb?h=master) from the meta-arm layer might be enough.

madisongh (changes requested):

recipes-bsp/uefi/edk2-basetools-native_202302.bb
madisongh, line 1: This recipe should probably have `-tegra` in it, and use the L4T version, since we're using NVIDIA's sources here.

We could probably rework the `.inc` file a bit to make the SRC_URI needed here available as a common variable, too.

recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
```diff
@@ -0,0 +1,69 @@
...
+do_install() {
+    install -d ${D}${CAPSULE_DIR}
+    install -m 0644 ${B}/tegra-bl.cap ${D}${CAPSULE_DIR}
+    install -m 0644 ${B}/tegra-kernel.cap ${D}${CAPSULE_DIR}
+    install -d ${D}${bindir}
+    install -m 0755 ${WORKDIR}/oe4t-set-uefi-OSIndications ${D}${bindir}
+}
+
+FILES:${PN} += "${CAPSULE_DIR}"
+RDEPENDS:${PN} = "bash"
```
madisongh, line 69: This recipe should be marked machine-specific with `PACKAGE_ARCH = "${MACHINE_ARCH}"`

classes/bup.bbclass
madisongh, line 1: This should have `tegra` in the name, to avoid possible conflicts with bbclasses in other layers (although the likelihood of "bup" being used elsewhere is probably low).

recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
```diff
@@ -0,0 +1,69 @@
...
+        -v --encode --monotonic-count 1 \
+        --fw-version "${BSP_VERSION32}" \
+        --lsv "${BSP_VERSION32}" \
+        --guid "${GUID}" \
+        --signer-private-cert "${UEFI_CAPSULE_SIGNER_PRIVATE_CERT}" \
+        --other-public-cert "${UEFI_CAPSULE_OTHER_PUBLIC_CERT}" \
+        --trusted-public-cert "${UEFI_CAPSULE_TRUSTED_PUBLIC_CERT}" \
+        -o ./tegra-bl.cap \
+        ${DEPLOY_DIR_IMAGE}/${BUPFILENAME}.bl_only.bup-payload
+    python3 ${PYTHON_BASETOOLS}/Capsule/GenerateCapsule.py \
```
madisongh, line 42: This command should be conditional on the existence of the kernel-only BUP payload, which (legitimately) may not exist.

madisongh (changes requested):
The basic structure LGTM.  See in-line for some minor changes needed.

Lexmark-chad:
@madisongh , thanks for all the feedback.  I am close on the `OsIndications` rework and will hopefully get to that tomorrow.

recipes-bsp/uefi/edk2-basetools-native_202302.bb
Lexmark-chad: Recipe has been renamed.  See if you think I factored out the SRC_URI appropriately.

recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
Lexmark-chad: Done.

classes/bup.bbclass
Lexmark-chad: Done.

recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
Lexmark-chad: Is it appropriate to check both?  I followed the pattern in `tegra-bup-payload_1.0.bb` and checked for the existence of both kernel-only and bl-only BUP payloads.

recipes-bsp/uefi/tegra-uefi-capsules/oe4t-set-uefi-OSIndications
Lexmark-chad: See what you think of the refactoring.  I initially wanted to make a `.func` file for both compat spec logic and for setting the uefi var, but they were intertwined just enough that I left it together.

madisongh: At first glance it looks good - just the sort of thing I had in mind..  We should probably use the full path to the `.func` file in the `source` line, though.  I'll review in more detail when I get a chance.

Lexmark-chad: Full path has been taken care of.  Not sure why, but the auto-builder / checklayer broke after I forced push.  The commits all look correct so not sure what happened.  I am still chasing down an issue during my testing.  Invoking the script to set `OsIndications` works the first time through, but not the second.  I am tracking that down now.

Lexmark-chad: OK... I should have looked at the existing logic a little closer (and the comment).  I'll have to rework that logic as it looks like the design intent was to only write `TegraPlatformSpec` and `TegraPlatformCompatSpec` the first time -  presumably because those should be static values.

madisongh: > Not sure why, but the auto-builder / checklayer broke after I forced push. The commits all look correct so not sure what happened.

Force-pushes (and some other unusual events) can confuse the autobuilder's PR handler.  Fixing that is on my to-do list, but in the meantime, I re-queued the job.

Lexmark-chad: Logic should be fixed now.  Tested A->B and B->A on Orin AGX Devkit.  Will test Xavier AGX Devkit next.

Lexmark-chad: Tested A->B and B->A on Xavier AGX Devkit as well and everything looks OK.  I think this is gtg pending any additional feedback.

madisongh (approved):
LGTM

dwalkes (commented):
Thanks again @Lexmark-chad for taking this on.  I tried today with `jetson-xavier-nx-devkit-emmc`, adding to the build with:
```
IMAGE_INSTALL:append = " tegra-uefi-capsules"
```
Then after booting up with the resulting build, copied the tegra-bl.cap file to the expected location for UEFI:
```
cp /opt/nvidia/UpdateCapsule/tegra-bl.cap /opt/nvidia/esp/EFI/UpdateCapsule/TEGRA_BL.Cap
```
Then ran:
```
oe4t-set-uefi-OSIndications
```
Verified the value was set:
```
root@jetson-xavier-nx-devkit-emmc:~# efivar -n 8be4df61-93ca-11d2-aa0d-00e098032b8c-OsIndications
GUID: 8be4df61-93ca-11d2-aa0d-00e098032b8c
Name: "OsIndications"
Attributes:
        Non-Volatile
        Boot Service Access
        Runtime Service Access
Value:
00000000  04 00 00 00 00 00 00 00                           |........        |
root@jetson-xavier-nx-devkit-emmc:~# efivar -n 8be4df61-93ca-11d2-aa0d-00e098032b8c-OsIndications
GUID: 8be4df61-93ca-11d2-aa0d-00e098032b8c
Name: "OsIndications"
Attributes:
        Non-Volatile
        Boot Service Access
        Runtime Service Access
Value:
00000000  04 00 00 00 00 00 00 00                           |........        |
```
Then rebooted
After reboot I got 
```
root@jetson-xavier-nx-devkit-emmc:~# nvbootctrl dump-slots-info
Current version: 35.3.1
Capsule update status: 0
Current bootloader slot: A
Active bootloader slot: A
num_slots: 2
slot: 0,             status: normal
slot: 1,             status: normal
```
I didn't have a serial port connected yet, this would be the next step to debug this.  Wanted to make sure I'm not missing something.  @Lexmark-chad have you tried on `jetson-xavier-nx-devkit-emmc` yet or only on AGX Xavier?

Lexmark-chad:
> @Lexmark-chad have you tried on `jetson-xavier-nx-devkit-emmc` yet or only on AGX Xavier?

I've tried it on both Orin & Xavier AGX.  I'll take a peek at the Xavier NX Devkit.

Lexmark-chad:
@dwalkes, I got the same result.  I'll have to get someone here rig me up a serial debug cable so I can see what is going on during boot.

dwalkes:
This is exactly what happened before https://github.com/OE4T/meta-tegra/commit/64a3404aa2e7f2bb5735c99b6955fbb2c953441e but I see that is in the kirkstone branch.  I actually only tested it on master with https://matrix.to/#/!YBfWVpJwNVtkmqVCPS:gitter.im/$S8dzzJNfmhMzAt5cp4ikgNjJtez98R68O_39q3oAG0M?via=gitter.im&via=matrix.org&via=3dvisionlabs.com I believe.  I will try to re-test the NVIDIA built capsule update on kirkstone tonight.

dwalkes:
> I will try to re-test the NVIDIA built capsule update on kirkstone tonight.

I've verified I can successfully complete an update with this TEGRA_BL.Cap generated with the Jetpack 5.1.1 release and instructions [here](https://docs.nvidia.com/jetson/archives/r35.3.1/DeveloperGuide/text/SD/Bootloader/UpdateAndRedundancy.html#generating-the-capsule-update-payload) and the rest of the image matching the kirkstone build on this branch.  So it appears to be some difference in capsule content which causes the issue although I'm not sure what that would be.

Lexmark-chad:
When I used NVIDIA's tooling with the `*.bl_only.bup-payload` from the build it also didn't work.  Continuing to investigate...

Lexmark-chad:
I finally got a serial cable hooked up.  There wasn't anything that stood out as a hint of what is going on so I instrumented the tooling to output the exact command used to generate the bup payloads.  The main difference I saw was that L4T tooling uses `nvdisp-init.bin` for the `cpu-bootloader` partition while `meta-tegra` uses `uefi_jetson.bin`.  If I understand correctly that just means boot with or without a splash screen.  My next thought will be to try and enable debug in uefi so I get more serial trace output at boot.  @madisongh, can you think of anything else I should be looking at?

dwalkes:
>  My next thought will be to try and enable debug in uefi so I get more serial trace output at boot.

I suggest `EDK2_BUILD_MODE:pn-edk2-firmware-tegra = "DEBUG"` , as well as turning on debug prints [here](https://github.com/NVIDIA/edk2-nvidia/blob/456628b951de054f5fe19da2ebad304feeac62db/Platform/NVIDIA/NVIDIA.common.dsc.inc#L456)

Lexmark-chad:
Just changing `BUILD_MODE` to `DEBUG` produced useful information:

```
Jetson UEFI firmware (version v35.3.1 built on 2023-01-24T13:18:32+00:00)
ESC   to enter Setup.
F11   to enter Boot Manager Menu.
Enter to continue boot.
FwPackageValidateImageInfo WARNING: image info for 'BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_A' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_A' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_A' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_A' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_A' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_B' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_B' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_B' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_B' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT_B' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT-boot-chain_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT-boot-chain_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT-boot-chain_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT-boot-chain_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'BCT-boot-chain_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1_b' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1_b' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1_b' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1_b' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb1_b' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MB1_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MB1_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MB1_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MB1_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MB1_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MEM_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MEM_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MEM_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MEM_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'MEM_BCT' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'spe-fw' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mb2' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mts-preboot' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mts-mce' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'mts-proper' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'VER' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'VER' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'VER' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'VER' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'VER' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'cpu-bootloader' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bootloader-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bootloader-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bootloader-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bootloader-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bootloader-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secure-os' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'eks' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bpmp-fw' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bpmp-fw-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bpmp-fw-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bpmp-fw-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bpmp-fw-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'bpmp-fw-dtb' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'rce-fw' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'adsp-fw' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'sc7' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'xusb-fw' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt_backup' has version=13617, expected=0
FwPackageValidateImageInfo WARNING: image info for 'secondary_gpt_backup' has version=13617, expected=0
bootloader-dtb not found in package: Unsupported
FmpDxe(NVIDIA System Firmware): CheckTheImage() - FmpDeviceLib CheckImage failed. Status = Aborted
FmpDxe(NVIDIA System Firmware): SetTheImage() - Check The Image failed with Aborted.
HandleCapsules˙ˇ‚
```

Lexmark-chad:
I'm getting closer.  The tnspecs are different between meta-tegra builds and l4t tooling which causes the code in `FwPackageGetImageIndex` to flow through differently.  I've added debug all throughout that function and it is clearly following a different code path.  The tnspecs for `bootloader-dtb` for meta-tegra are:

```
3668-100-0001--1-2-jetson-xavier-nx-devkit-emmc-
3668-200-0001--1-2-jetson-xavier-nx-devkit-emmc-
3668-300-0001--1-2-jetson-xavier-nx-devkit-emmc-
3668-301-0001--1-2-jetson-xavier-nx-devkit-emmc- 
3668-301-0003--1-2-jetson-xavier-nx-devkit-emmc-
```

while l4t's is:

```
3668-100-0001--1-2-jetson-xavier-nx-devkit-emmc-
3668-301-0001--1-2-jetson-xavier-nx-devkit-emmc-
```

It's the `3668-200-0001--1-2-jetson-xavier-nx-devkit-emmc-` spec which causes the code to flow differently based on the call to `FwPackageGetImageIndex` with these parameters:

`FwPackageGetImageIndex:  Name: bootloader-dtb, FullSpec: 3668-200-0001-G.0-1-2-jetson-xavier-nx-devkit-emmc-, CompatSpec: 3668-100---1--jetson-xavier-nx-devkit-emmc-`

Looking at that code now to understand design intent.

Lexmark-chad:
While it was true that the `3668-200-0001--1-2-jetson-xavier-nx-devkit-emmc-` spec caused it to flow through a different code path and eventually return `EFI_UNSUPPORTED` the problem had already occurred when `3668-100-0001--1-2-jetson-xavier-nx-devkit-emmc-` didn't match the compat spec and kick it out of the loop.  I made a small change which appears to have fixed the problem, but I think it's better to actually refactor that code a bit as I found it confusing and hard to follow given what they were trying to accomplish.  I'll give that a shot and we can patch it in `meta-tegra` while we wait for it to be upstreamed.  Any opinions on whether the patch should be added to this PR and have it all go together?

madisongh:
Hmm.  The TEGRA_BUPGEN_SPECS setting for jetson-xavier-nx-devkit-emmc is incorrect.  I must have missed that when converting us over to using the compatibility specs.  We *should* be building just the -100 and -301 entries, just like L4T does.

>  the problem had already occurred when 3668-100-0001--1-2-jetson-xavier-nx-devkit-emmc- didn't match the compat spec and kick it out of the loop

Are we setting up the compat spec wrong for the Xavier NXes, too?

Lexmark-chad:
> Are we setting up the compat spec wrong for the Xavier NXes, too?

That looks correct:

https://github.com/OE4T/meta-tegra/blob/kirkstone/conf/machine/jetson-xavier-nx-devkit.conf#L7

madisongh:
I mean in this function: https://github.com/OE4T/meta-tegra/blob/3c429b60115600ddf3adaa7be28d6c2c0bfa1019/recipes-bsp/tools/setup-nv-boot-control/setup-nv-boot-control.sh.in#L61-L69

Lexmark-chad:
I modified `TEGRA_BUPGEN_SPECS` for the `jetson-xavier-nx-devkit-emmc` machine conf to match the l4t entries per your observation and it still failed.  I'll re-check everything tomorrow to see if I messed up something.  Also, per prior post:

```
root@jetson-xavier-nx-devkit-emmc:~# efivar --print --name 781e084c-a330-417c-b678-38e696380cb9-TegraPlatformCompatSpec
GUID: 781e084c-a330-417c-b678-38e696380cb9
Name: "TegraPlatformCompatSpec"
Attributes:
        Non-Volatile
        Boot Service Access
        Runtime Service Access
Value:
00000000  33 36 36 38 2d 31 30 30  2d 2d 2d 31 2d 2d 6a 65  |3668-100---1--je|
00000010  74 73 6f 6e 2d 78 61 76  69 65 72 2d 6e 78 2d 64  |tson-xavier-nx-d|
00000020  65 76 6b 69 74 2d 65 6d  6d 63 2d                 |evkit-emmc-     |
```

Lexmark-chad:
False alarm.  Not everything rebuilt after I changed the machine conf to make `TEGRA_BUPGEN_SPECS` match L4T's list.  Once rebuilt everything worked as expected.  I will soon add that change to the PR.  I'll log an issue against [edk2-nvidia](https://github.com/NVIDIA/edk2-nvidia/tree/main) as there is clearly a logic bug that will eventually get exposed.

dwalkes:
Thanks @Lexmark-chad , kicked off a build, will retry in the morning.

dwalkes:
Confirmed working on `jetson-xavier-nx-devkit-emmc`.  Thanks again @Lexmark-chad !

dwalkes (approved):

Lexmark-chad:
[Issue](https://github.com/NVIDIA/edk2-nvidia/issues/69) logged.

PR #1297, opened 2023-07-17
Tegra minimal init fix boot scripts

liuming50:


madisongh:
Thanks @liuming50 , looks like some good cleanup here.  I'll take a closer look at the changes when I have a bit more time.

recipes-core/initrdscripts/tegra-minimal-init_1.0.bb
```diff
@@ -19,6 +19,7 @@ do_install() {
     mknod -m 622 ${D}/dev/console c 5 1
     install -d ${D}${sysconfdir}
     install -m 0644 ${WORKDIR}/platform-preboot.sh ${D}${sysconfdir}/platform-preboot
+    sed -i -e "s#@@TNSPEC_BOOTDEV@@#${TNSPEC_BOOTDEV}#g" ${D}${sysconfdir}/platform-preboot
```
madisongh, line 22: With this change, the recipe is now machine-specific, so `PACKAGE_ARCH = "${MACHINE_ARCH}"` needs to be added at the end of the recipe.

recipes-core/initrdscripts/tegra-minimal-init/platform-preboot.sh
```diff
@@ -35,8 +35,22 @@ fi
 if [ -z "$rootdev" ]; then
     echo "[FAIL]"
     if [ -n "$mayberoot" ]; then
-	rootdev="$mayberoot"
+        if [ "`echo $mayberoot | cut -c1-5`" = "UUID=" ]; then
```
madisongh, line 39: I'm fine with trying to support these additional ways to specify the rootfs device, but will the `/dev/disk/by-xxxx` symlinks be present?  I thought it was udev (or systemd-udevd) that created those, and that wouldn't be running in a minimal initramfs.

Using `lsblk` to look up partitions by the specified attributes might work better, maybe?

madisongh (changes requested):

recipes-core/initrdscripts/tegra-minimal-init/platform-preboot.sh
liuming50: @madisongh thanks for the input, you are absolutely right, will change that in V2.

liuming50:
Changes in patch set V2:

1 Add "PACKAGE_ARCH = "${MACHINE_ARCH}"" to tegra-minimal-init_1.0.bb as suggested by @madisongh 

2 Drop the udev rules dependencies, change to use 'blkid' as suggested by @madisongh 

3 Add one more fix to rootwait condition checking in init-boot.sh

madisongh (approved):
Updated changes LGTM.  Thanks again!

PR #1313, opened 2023-08-07
tegra-mmapi-samples: fix build issues

ichergui:
- refresh patch
- update License
- add vulkan as dependency


madisongh:
This is fine for now.  We probably need to a bit more work to make the Vulkan dependency optional based on DISTRO_FEATURES, but that can wait.

PR #1319, opened 2023-08-10
edk2-firmware-tegra: adds ostree support

quaresmajose:
This group of patches is only applied with the sota override
 that is provided by the meta-updater [1] layer
    
 [1] https://github.com/uptane/meta-updater

### DIFF
--- a/external/sota/recipes-bsp/uefi/edk2-firmware-tegra/0001-L4TLauncher-boot-syslinux-instead-of-extlinux-for-os.patch
+++ b/external/sota/recipes-bsp/uefi/edk2-firmware-tegra/0001-L4TLauncher-boot-syslinux-instead-of-extlinux-for-os.patch
@@ -1,0 +1,81 @@
+From c5680ec1c81fd841d638968988757bd0abfb9d2f Mon Sep 17 00:00:00 2001
+From: Ricardo Salveti <ricardo@foundries.io>
+Date: Tue, 28 Feb 2023 10:59:20 +0000
+Subject: [PATCH 1/3] L4TLauncher: boot syslinux instead of extlinux for ostree
+ support
+
+Ostree uses syslinux.cfg instead of extlinux.conf for providing the
+required kernel/initrd/dtb files for the bootloader.
+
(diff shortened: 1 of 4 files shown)


madisongh:
Hmm.  I'm not sure it's a good idea to carry these patches here, since we don't normally run any builds with the meta-updater layer that would exercise them.

quaresmajose:
Sorry for the late replay but I've been off for the last few weeks.

Our main ideia is too show other folks that is possible to use [ostree](https://github.com/ostreedev/ostree) with meta-tegra adding another possible over the air update method to update the target image.

We have been using it on the `Linux microPlatform project` It's been a year now and it's been working pretty well.
https://github.com/foundriesio/meta-lmp/commits/main/meta-lmp-bsp/dynamic-layers/tegra/recipes-bsp/uefi
All the required patches we use is the ones on this PR and making them more visible would be useful for others to test and also improves them.

I understand that having it there can be tricky because they are not possible to test on the CI without using the meta-updater layer.

quaresmajose:
I think I will wait for the https://github.com/OE4T/meta-tegra/tree/wip-l4t-r35.4.1 merge and just submit the first patch which somewhat standardizes the way of applying the patches on the `edk2-firmware-tegra` which is constituted by a lot of git repos and that to improve the situation use the competitor's OS end of line.

quaresmajose:
First patch spited out and proposed in https://github.com/OE4T/meta-tegra/pull/1349

quaresmajose:
I build and test meta-tegra regularly (master branch every week and kirkstone daily) with these patches enabled so can can take care of it in order to keep them functional.

madisongh:
How about making these a bbappend in `external/sota-layer/recipes-bsp/uefi`, so they're only used when that layer is included in the build?

quaresmajose:
This makes perfect sense, I'll adapt it

quaresmajose:
Removed the `sota` override and move it to `external/sota`.

quaresmajose:
Many thanks!

PR #1323, opened 2023-08-14
image_types_tegra: added tegra234 donvme flash script

ssuesens:
This script can be used to flash a nvme connected to a host system.

I create the donvme.sh script to flash  a orin NX image on a nvme device connected with a x86 machine.
The script worked but I don't understand the parameter "./tegra234-flash-helper.sh --sdcard -B ${TEGRA_BLBLOCKSIZE} -s ${TEGRAFLASH_SDCARD_SIZE}" I think this parameters doesn't influence the viewing of the nvme partition. 

I would appreciate when you have a glimpse in my pull request.

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -613,6 +613,16 @@ END
         chmod +x burnfuses.sh
     fi
 
+    if [ -n "${IMAGE_TEGRAFLASH_INITRD_FLASHER}" ]; then
+        rm -f donvme.sh
+        cat > donvme.sh <<END
+#!/bin/sh
+./nvflashxmlparse --extract --type rootfs --change-device-type=nvme -o external-nvme-flash.xml.tmp external-flash.xml.in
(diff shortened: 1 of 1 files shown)


madisongh:
I like the idea, but I'd like to have one script/command that could be used to write any locally-mounted external storage (SDcard, NVMe, or USB), rather than having `doXXXX.sh` scripts for each type.  A few tweaks to the helper scripts would probably make that easier.   Let me think about it.

ssuesens:
I understand your idea, sounds good. 
I had further proposal. What do you think about a resize option in the script?

I flash the NVME for production on a host PC and to rapid the flash process I create the rootfs small as possible.
After NVME flash I resize e.g. the rootfs partition to the maximum NVME size.

ilies-chergui:
Hey @ssuesens 
I think you should wait for @madisongh to rebase the commits to `kirkstone` branch then update your PR

ichergui:
Hey @ssuesens 
I think you should wait until @madisongh update [kirkstone](https://github.com/OE4T/meta-tegra/tree/kirkstone) branch then you can rebase based on that.

madisongh:
With the JetPack 6/L4T R36.3.0 update, I added a `doexternal.sh` script to the tegraflash package when you've configured the build to put the rootfs on an external drive (see commit 667e13147aa671af3d6aa6769dfc84ddddb8e0ad).  This should be present in scarthgap and master.

PR #1331, opened 2023-08-18
Adds deploy task for capsules

joekale-pp:
deploy class flushes from DEPLOYDIR to DEPLOY_DIR_IMAGE as an sstate task.

### DIFF
--- a/recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
+++ b/recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
@@ -2,7 +2,7 @@ DESCRIPTION = "Generate UEFI capsules for bup paylods"
 LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
-inherit l4t_bsp tegra-bup
+inherit l4t_bsp tegra-bup deploy
 
 COMPATIBLE_MACHINE = "(tegra)"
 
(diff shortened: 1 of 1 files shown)


recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
```diff
@@ -71,3 +71,21 @@ do_install() {
...
+    if [ -e ${B}/tegra-kernel.cap ]; then
+	KERNEL_NAME=${TNSPEC_MACHINE}-tegra-kernel.cap
+	install -m 0644 ${B}/tegra-kernel.cap ${DEPLOYDIR}/$KERNEL_NAME
+	ln -s -r ${DEPLOYDIR}/$KERNEL_NAME ${DEPLOYDIR}/tegra-kernel.cap
+    fi
+}
+
+addtask deploy after do_install
+
+do_compile[depends] += "tegra-bup-payload:do_install"
```
madisongh, line 91: Good catch on the missing dependency, but the value here should really be `"${@bup_dependency(d)}"`, just like the `do_install` dependency in the `tegra-bup-payload` recipe.

madisongh (changes requested):
Thanks @joekale-pp ... mostly looks good, just one minor change needed.

recipes-bsp/uefi/tegra-uefi-capsules_35.3.1.bb
joekale-pp: should be corrected now.

madisongh (approved):

PR #1335, opened 2023-08-21
libdrm-nvdc: set DEBIAN_NOAUTONAME also for license package

sapiippo:
This is needed to avoid spdx from failing with:

ERROR: libdrm-nvdc-35.3.1-20230319081403-r0 do_create_spdx: The recipe libdrm-nvdc is trying to install files into a shared area when those files already exist. Those files and their manifest location are:
  /home/qt/work/build/build/tmp/deploy/spdx/armv8a_tegra/packages/libdrm-lic.spdx.json
    (matched in manifest-armv8a_tegra-libdrm.create_spdx)
Please verify which recipe should provide the above files.


madisongh:
Thanks!

PR #1336, opened 2023-08-21
jetson-orin-nano-devkit-nvme: set TNSPEC_BOOTDEV instead of TNSPEC_BO…

ichergui:
…OTDEV_DEFAULT

so that the external flash layout gets populated in the tegraflash package.


ichergui:
This PR is needed to address the following discussion:
https://github.com/OE4T/meta-tegra/discussions/1329

ichergui:
Closing this PR since Matt have it with other stuff.

Issue #1046, opened 2022-08-26
devtool modify edk2-firmware-tegra fails with wip-jetpack-5-ga branch.


atharvanan1:
**Describe the bug**
`devtool modify edk2-firmware-tegra` fails with wip-jetpack-5-ga branch.

**To Reproduce**
1. Build `meta-tegra` branch 'wip-jetpack-5-ga' with `jetson-xavier-nx-devkit-emmc`
2. Set `PREFERRED_PROVIDER_virtual/bootloader = "edk2-firmware-tegra"`
3. Run with devtool argument `modify edk2-firmware-tegra`
5. See error log below:
```
oe@pop-os:~/build/tegra-demo-distro/build (wip-jetpack-5-ga)$ devtool modify edk2-firmware-tegra
NOTE: Starting bitbake server...
INFO: Creating workspace layer in /home/oe/build/tegra-demo-distro/build/workspace
NOTE: Reconnecting to bitbake server...
NOTE: Retrying server connection (#1)...
Loading cache: 100% |                                                                                           | ETA:  --:--:--
Loaded 0 entries from dependency cache.
Parsing recipes: 100% |##########################################################################################| Time: 0:00:17
Parsing of 2811 .bb files complete (0 cached, 2811 parsed). 4342 targets, 116 skipped, 0 masked, 0 errors.
Removing 1 recipes from the jetson_xavier_nx_devkit_emmc sysroot: 100% |#########################################| Time: 0:00:00
NOTE: Resolving any missing task queue dependencies

Build Configuration:
BB_VERSION           = "2.0.1"
BUILD_SYS            = "x86_64-linux"
NATIVELSBSTRING      = "universal"
TARGET_SYS           = "aarch64-oe4t-linux"
MACHINE              = "jetson-xavier-nx-devkit-emmc"
DISTRO               = "tegrademo"
DISTRO_VERSION       = "4.1+snapshot-20220826"
TUNE_FEATURES        = "aarch64 armv8a crc"
TARGET_FPU           = ""
meta                 = "HEAD:21cfb871f7da604f1fab957d7747cb03b113b971"
meta-tegra           = "rootfs-ab-nx:ada84fa8932376178deb778ce83c745f16122a66"
meta-oe              
meta-python          
meta-networking      
meta-filesystems     = "HEAD:78fe37bb9c1456263df3a6ec450a486a2d8464db"
meta-virtualization  = "HEAD:884355eeedf0126c9cc7d5e704d4262feb753510"
[...]

PR #1349, opened 2023-08-29
edk2-firmware-tegra: refresh patches

quaresmajose:
Split the patches in groups and apply on the corresponding directory.

Fix https://github.com/OE4T/meta-tegra/issues/1046


ichergui (approved):
I tried with this updates and `$ devtool modify edk2-firmware-tegra` works properly 
Thanks @quaresmajose

PR #1352, opened 2023-09-02
[master] Update to L4T R35.4.1/JetPack 5.1.2

madisongh:


quaresmajose:
Patch https://github.com/OE4T/meta-tegra/pull/1352/commits/4025a5cd48ae926d2cc833b65c1b3d7b78244495 changes the author `Fom: ` and `Date: ` of all patches in `gstreamer1.0-plugins-nvvideo4linux2-1.14.0-r35.4.1` but in my opinion we should keep the original version.

madisongh:
Good point.  I have been pretty lax about things like that, as well as marking the patches with other metadata (such as `Upstream-Status`).  I would like to get that cleaned up and enable the patch checks that were added to `yocto-checklayer`.

quaresmajose:
To restore the patch headers https://github.com/OE4T/meta-tegra/pull/1355. I think the `Upstream-Status` is now mandatory in oe-core master because I have already send some fixes about that for another layers, anyway I have been testing meta-tegra master regularly every week and I have not seen any problems due to this.

madisongh:
Thanks.   So far, at least, the tools only error out on missing `Upstream-Status` in OE-Core recipes.  We do get warnings about them from the `yocto-checklayer` runs, but they aren't fatal.  I think it's a good idea to have that info available in the patches, though, and it's a reminder to try to upstream fixes when we can.

PR #1356, opened 2023-09-06
initrd-flash: Add argument to format internal mmc partitions

quaresmajose:
- Preparation for adding support to flash wic images on external devices

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
@@ -14,6 +14,7 @@ Options:
   -h|--help             Displays this usage information
   --skip-bootloader     Skip boot partition programming
   --usb-instance        USB instance of Jetson device
+  --format-mmc-part     Format (ext4) mmc partition
   --erase-nvme          Erase NVME drive during flashing
 
 Options passed through to flash helper:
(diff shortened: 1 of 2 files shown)


madisongh:
This seems like a too-specific addition to the script.  It might help if I could see more context around how this would enable use of wic for external devices (which might be relevant for #1323 as well).

Otherwise, I'd like to morph this into adding a generic mechanism for adding custom extensions to the initrd-flash process, maybe by installing extension scripts into a known location in the flashing initrd plus adding extension client-side script fragments into the tegraflash package that the initrd-flash script would automatically consume.  That way, if anyone has stuff they need to add to the flashing process for their specific needs, they can do that.

quaresmajose:
Look like this is very aligned with our use case in the Linux microPlatform (LmP) as it is to solve possible issues when we do an over the air update after moving the rootfile system from the internal mmc to the external nvme. I sent this ones separately because it was easy to decouple from all other patches on the serie I have.

I agree with you that the better approach for this is adding support for installing extension scripts as well as script fragments into the tegraflash to control them. I plan to come back to this in the next few weeks to add the pending patches so can try to implement such feature too.

Thank you very much for your input, which is always very valuable.

quaresmajose:
About the https://github.com/OE4T/meta-tegra/pull/1323: our implementation is to flash the wic on the nvme using the initrd flasher and using the tegra target to do that. We don't flash yet the nvme directly on the host machine for now.

madisongh:
@quaresmajose Can you take a look at #1535 and see if the generic callout implemented there would work for you?

quaresmajose:
> @quaresmajose Can you take a look at https://github.com/OE4T/meta-tegra/pull/1535 and see if the generic callout implemented there would work for you?

This solution sounds good, I will adapt my patch with this generic callout and will do some tests.

Issue #1364, opened 2023-09-13
Jetson Nano 2GB: kernel backtrace on dunfell and kirkstone-l4t-r32.7.x branches with L4T r32.7.4


housefm666:
**Describe the bug**
With the L4T r32.7.4 I got a kernel backtrace, see the attachment, so I had to revert to L4T r32.7.3 version.
It might be the problem with loading the GPU driver, nvargus-daemon, or after it.

Working versions:
[kirkstone-l4t-r32.7.x](https://github.com/OE4T/meta-tegra/tree/a0dfc1dcd080b10a1a9a8b5349e32efdda125cb3)
[dunfell](https://github.com/OE4T/meta-tegra/tree/fd7b1da0dc0fe42afa3203c884e1865d5a7b32fa)

**To Reproduce**
Compile and flash core-image-weston with L4T r32.7.4 and run on Jetson Nano 2GB

You can reproduce with the latest commits:
[kirkstone-l4t-r32.7.x](https://github.com/OE4T/meta-tegra/tree/9401c5a564d48f1f1f649f06f15702cd36814192)
[dunfell](https://github.com/OE4T/meta-tegra/tree/133446740965f2ec7c97694d163d47db4009aa83)

** Backtrace **
[dunfell_7.4_backtrace.txt](https://github.com/OE4T/meta-tegra/files/12595870/dunfell_7.4_backtrace.txt)

dwalkes:
Thanks @housefm666 
Each of the warnings look like this:
```
Starting nvargus-daemon: [OK]
[   10.653594] ------------[ cut here ]------------
[   10.658215] WARNING: CPU: 0 PID: 2147 at mm/slab_common.c:98 kmem_cache_create+0x1cc/0x258
[   10.666459] Modules linked in: g_serial nvgpu tegra_udrm userspace_alert dione_ir
[   10.673973] 
[   10.675459] CPU: 0 PID: 2147 Comm: nvphsd_setup.sh Not tainted 4.9.337-l4t-r32.7.4+ga1cf17bc0aa0 #1
[   10.684484] Hardware name: NVIDIA Jetson Nano 2GB Developer Kit (DT)
[   10.690822] task:         (ptrval) task.stack:         (ptrval)
[   10.696727] PC is at kmem_cache_create+0x1cc/0x258
[   10.701505] LR is at kmem_cache_create+0xd4/0x258
[...]

PR #1371, opened 2023-09-16
Haveged, nvpmodel, and nvphs startup fixes for dunfell

madisongh:
* Add a bbappend to fix its init script
* Minor fixes for the nvpmodel and nvphs startup files 

Related to #1364


madisongh:
The checklayer failure is due to dunfell not being compatible with ARM64 build hosts.  Will bypass the check for now and update the autobuilder to require an x86 instance for running layer checks on dunfell-based branches.

PR #1380, opened 2023-09-24
Incorporate signing server support into tegra-demo-distro

Lexmark-chad:
@madisongh , I wanted to get this up for early feedback.  I discussed this at a high level in the last (September) OE4T monthly meeting.  I'm starting to work now on getting `digsigserver` to support UEFI signing of various artifacts.  My desire was to try and start simple (Orin NX development module fused with development keys, Orin Nano devkit carrier board, tegra-demo-distro build) and work my way up to supporting our own custom carrier card, tegra build, and production keys.  I refactored how the UEFI signing worked with the intent of making it easier for others to follow the same path.  My approach was to try and corral all the `TEGRA_UEFI_DB_KEY` and `TEGRA_UEFI_DB_CERT` references spread out in various recipes and bring those back to the `tegra-uefi-signing.bbclass`.  Then one can add a `signing-server` override to `OVERRIDES` in their `local.conf` to cause the build to use the signing server equivalents of the functions you provided for supporting signing with local keys.  This approach cuts down on having to `.bbappend` recipes and reduces the friction required to utilize an external signing server.  In the meeting I discussed creating signed and unsigned equivalents of machine configurations using this override mechanism, but I think that would possibly create more clutter than it is worth and settled on the approach I described above.  There was good discussion in the meeting regarding whether an advanced feature like this should be brought into the demo distro so I understand if you want to keep things simple.  Any feedback you have is welcome.

### DIFF
--- a/classes/image_types_tegra.bbclass
+++ b/classes/image_types_tegra.bbclass
@@ -1,5 +1,9 @@
 inherit image_types image_types_cboot image_types_tegra_esp python3native perlnative kernel-artifact-names
 
+TEGRA_UEFI_SIGNING_CLASS ??= "tegra-uefi-signing"
+inherit ${TEGRA_UEFI_SIGNING_CLASS}
+TEGRA_UEFI_USE_SIGNED_FILES ??= "false"
+
 IMAGE_TYPES += "tegraflash"
 
 IMAGE_ROOTFS_ALIGNMENT ?= "4"
@@ -25,13 +29,13 @@ def tegra_dtb_extra_deps(d):
     deps = []
     if d.getVar('PREFERRED_PROVIDER_virtual/dtb'):
         deps.append('virtual/dtb:do_populate_sysroot')
-    if d.getVar('TEGRA_UEFI_DB_KEY') and d.getVar('TEGRA_UEFI_DB_CERT'):
+    if d.getVar('TEGRA_UEFI_USE_SIGNED_FILES') == "true":
         deps.append('tegra-uefi-keys-dtb:do_populate_sysroot')
     return ' '.join(deps)
 
 def tegra_bootcontrol_overlay_list(d, bup=False, separator=','):
     overlays = d.getVar('TEGRA_BOOTCONTROL_OVERLAYS').split()
-    if d.getVar('TEGRA_UEFI_DB_KEY') and d.getVar('TEGRA_UEFI_DB_CERT'):
+    if d.getVar('TEGRA_UEFI_USE_SIGNED_FILES') == "true":
         overlays.append('UefiDefaultSecurityKeys.dtbo')
         if bup and os.path.exists('UefiUpdateSecurityKeys.dtbo'):
             overlays.append('UefiUpdateSecurityKeys.dtbo')
@@ -64,7 +68,6 @@ TEGRAFLASH_ROOTFS_EXTERNAL = "${@'1' if d.getVar('TNSPEC_BOOTDEV') != d.getVar('
 ROOTFS_DEVICE_FOR_INITRD_FLASH = "${@tegra_rootfs_device(d)}"
 TEGRAFLASH_NO_INTERNAL_STORAGE ??= "0"
 OVERLAY_DTB_FILE ??= ""
-USE_UEFI_SIGNED_FILES ?= "${@'true' if d.getVar('TEGRA_UEFI_DB_KEY') and d.getVar('TEGRA_UEFI_DB_CERT') else 'false'}"
 
 TEGRA_EXT4_OPTIONS ?= "-O ^metadata_csum_seed"
 EXTRA_IMAGECMD:append:ext4 = " ${TEGRA_EXT4_OPTIONS}"
@@ -332,7 +335,7 @@ copy_dtbs() {
         fi
         bbnote "Copying KERNEL_DEVICETREE entry $dtbf to $destination"
         cp -L "${DEPLOY_DIR_IMAGE}/$dtbf" $destination/$dtbf
-	if ${USE_UEFI_SIGNED_FILES}; then
+	if ${TEGRA_UEFI_USE_SIGNED_FILES}; then
             cp -L "${DEPLOY_DIR_IMAGE}/$dtbf.signed" $destination/$dtbf.signed
 	fi
     done
@@ -345,7 +348,7 @@ copy_dtbs() {
             fi
             bbnote "Copying EXTERNAL_KERNEL_DEVICETREE entry $dtb to $destination"
             cp -L "${EXTERNAL_KERNEL_DEVICETREE}/$dtb" $destination/$dtbf
-	    if ${USE_UEFI_SIGNED_FILES}; then
+	    if ${TEGRA_UEFI_USE_SIGNED_FILES}; then
                 cp -L "${DEPLOY_DIR_IMAGE}/$dtb.signed" $destination/$dtbf.signed
 	    fi
         done
@@ -360,7 +363,7 @@ copy_dtb_overlays() {
     if [ -n "${IMAGE_TEGRAFLASH_INITRD_FLASHER}" ]; then
         extraoverlays="$extraoverlays L4TConfiguration-rcmboot.dtbo"
     fi
-    if ${USE_UEFI_SIGNED_FILES}; then
+    if ${TEGRA_UEFI_USE_SIGNED_FILES}; then
         extraoverlays="$extraoverlays UefiDefaultSecurityKeys.dtbo"
     fi
     for dtb in "$@" ${TEGRA_PLUGIN_MANAGER_OVERLAYS} $extraoverlays; do
--- a/classes/tegra-uefi-signing.bbclass
+++ b/classes/tegra-uefi-signing.bbclass
@@ -10,23 +10,30 @@ TEGRA_UEFI_DB_KEY ??= ""
 TEGRA_UEFI_DB_CERT ??= ""
 TEGRA_UEFI_SIGNING_TASKDEPS ?= "${@tegra_uefi_signing_deps(d, tasks=True)}"
 TEGRA_UEFI_SIGNING_DEPENDS ?= "${@tegra_uefi_signing_deps(d)}"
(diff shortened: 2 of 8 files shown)


madisongh:
Thanks @Lexmark-chad ... the refactoring makes sense to me (will review in detail later).  As for interfacing with digsigserver, I was thinking it might be better to create a new repo/layer, rather than putting the classes here or in the demo distro repo directly.  What do you think?

Lexmark-chad:
> As for interfacing with digsigserver, I was thinking it might be better to create a new repo/layer, rather than putting the classes/bbappends here or in the demo distro repo directly.

I think a new layer could make sense.  That's assuming that other production-level capabilities that you've experimented with in `tegra-test-distro` would make their way into this layer.  Otherwise, it might be overkill if it was just for this.  I don't consider myself advanced in regards to yocto layers, classes, and recipe structure so I'll want to check whether I can simply add a `tegra-uefi-signing.bbclass` to the proposed new layer and make sure that layer priorities causes the class with the signing server version of the signing routines to get picked up.

ichergui:
Thanks @Lexmark-chad. Should we use `meta-tegra-community` instead of creating new meta layer.

Lexmark-chad:
Ah, yes... Makes sense to me.

classes/tegra-uefi-signing.bbclass
```diff
@@ -13,20 +15,42 @@ TEGRA_UEFI_SIGNING_DEPENDS ?= "${@tegra_uefi_signing_deps(d)}"
...
-    truncate --size=%2048 "$1.signed"
-    cat "$1".sig.tmp >> "$1.signed"
-    rm "$1".sig.tmp
+    if [ -n "${TEGRA_UEFI_DB_KEY}" -a -n "${TEGRA_UEFI_DB_CERT}" ]; then
+        openssl cms -sign -signer "${TEGRA_UEFI_DB_CERT}" -inkey "${TEGRA_UEFI_DB_KEY}" -binary -in "$1" -outform der -out "$1".sig.tmp
+        cp "$1" "$1.signed"
+        truncate --size=%2048 "$1.signed"
+        cat "$1".sig.tmp >> "$1.signed"
+        rm "$1".sig.tmp
+    fi
```
madisongh, line 39: Based on how the `sign_bootimg()` function is calling this (up above in `image_types_cboot.bbclass`), I think you'll need to and an `else` clause here to copy `$1` to `$1.signed` if the key/cert variables are not set.

classes/tegra-uefi-signing.bbclass
```diff
@@ -1,3 +1,5 @@
+inherit signing_server
```
madisongh, line 1: Rather than directly inheriting the class here, how about we add a variable, something like `TEGRA_SIGNING_CLASS`, defaulting to the null string somewhere appropriate (maybe in the `tegra-common.inc` include file) and use that variable here - `inherit ${TEGRA_SIGNING_CLASS}`?  

That would allow another layer, or distro, to add the needed overrides in one place, without having to put specifics about the digsigserver implementation here in the BSP layer.  The same could be done in other places where signing steps might need modification - optee, BUP generation, etc.  We could even have different variables for each type of signing, if that seems appropriate.

I'd really like to give people a way to insert whatever signing mechanism they need here at the BSP layer level, rather than tying things to digsigserver specifically.

madisongh (changes requested):
See in-line

classes/tegra-uefi-signing.bbclass
Lexmark-chad: I like it...  I'll work on that.

Lexmark-chad:
@madisongh, I'm not sure if what I implemented was exactly what you had in mind, but your `${TEGRA_SIGNING_CLASS}` suggestion inspired the thought to apply it to `inherit tegra-uefi-signing` vs `inherit signing_server`.  Basically, make it polymorphic so the recipes can just call the tegra uefi signing functions and the local configuration can control which class's functions get called.  There still is some leakage of `TEGRA_UEFI_DB_*` references in `image_types_tegra.bbclass` so I'm going to take a peek at that next.

classes/tegra-uefi-signing.bbclass
Lexmark-chad: I unconditionally copied it since both the `if` and `else` clause would have executed it.

Lexmark-chad:
I think I've got everything implemented.  The last change moved `USE_UEFI_SIGNED_FILES` to `tegra-uefi-signing.bbclass` to remove the final references to the local signing keys.  Any class providing a different signing mechanism will be expected to define this in their class.  The auto-builder failures are complaining about parsing `Could not inherit file classes/${TEGRA_UEFI_SIGNING_CLASS}.bbclass` in one of the recipes.  I'm not sure how to resolve.

madisongh:
@Lexmark-chad A default definition (using `??=`) for `TEGRA_UEFI_SIGNING_CLASS` is probably needed.  Putting it in each recipe that has the `inherit` would work, and is probably the most proper place; another possibility would be to put it in the `layer.conf` file, since that's the only place we'd be sure of it getting parsed before the recipes, regardless of the target MACHINE setting.  The latter would be simpler, since it would only be one line, but is a bit unorthodox to put this kind of stuff in the layer config file.

An alternative would be to do something like `inherit ${@d.getVar('TEGRA_UEFI_SIGNING_CLASS') or ''}` instead, which, while a bit clunky-looking, should work.

classes/image_types_tegra.bbclass
```diff
@@ -25,13 +25,13 @@ def tegra_dtb_extra_deps(d):
     deps = []
     if d.getVar('PREFERRED_PROVIDER_virtual/dtb'):
         deps.append('virtual/dtb:do_populate_sysroot')
-    if d.getVar('TEGRA_UEFI_DB_KEY') and d.getVar('TEGRA_UEFI_DB_CERT'):
+    if d.getVar('USE_UEFI_SIGNED_FILES'):
```
madisongh, line 28: The definition of this variable is is in `tegra-uefi-signing.bbclass`, and will, presumably, also be set in any custom signing bbclass, so the signing class needs to be inherited here in `image_types_tegra` too.

Also, this check, plus the other python references to this variable, need to explicitly compare to `"true"`.  Just using `d.getVar()` will always succeed if the variable is defined, even if set to `"false"`.

And now that the usage of this variable spans multiple classes, it might be good to add a `TEGRA_` prefix to it, just to avoid possible conflicts with other layers.

classes/image_types_tegra.bbclass
```diff
@@ -25,13 +25,13 @@ def tegra_dtb_extra_deps(d):
...
         deps.append('tegra-uefi-keys-dtb:do_populate_sysroot')
     return ' '.join(deps)
 
 def tegra_bootcontrol_overlay_list(d, bup=False, separator=','):
     overlays = d.getVar('TEGRA_BOOTCONTROL_OVERLAYS').split()
-    if d.getVar('TEGRA_UEFI_DB_KEY') and d.getVar('TEGRA_UEFI_DB_CERT'):
+    if d.getVar('USE_UEFI_SIGNED_FILES'):
```
madisongh, line 34: See above

classes/image_types_tegra.bbclass
madisongh: Oh, and it probably wouldn't hurt to add a `??=` setting (to `"false"`) in this bbclass for this variable also, so it parses properly even if the signing class variable setting is null.

classes/tegra-uefi-signing.bbclass
madisongh: Right, I see now.

madisongh (changes requested):

classes/image_types_tegra.bbclass
```diff
@@ -332,7 +335,7 @@ copy_dtbs() {
         fi
         bbnote "Copying KERNEL_DEVICETREE entry $dtbf to $destination"
         cp -L "${DEPLOY_DIR_IMAGE}/$dtbf" $destination/$dtbf
-	if ${USE_UEFI_SIGNED_FILES}; then
+	if ${TEGRA_UEFI_USE_SIGNED_FILES} == "true"; then
```
madisongh, line 338: Here, in shell script, this could either be what you had before (`if ${TEGRA_UEFI_USE_SIGNED_FILES}`)  or it should be a valid POSIX shell construct (e.g., `if [ "${TEGRA_UEFI_SIGNED_FILES}" = "true" ]`).  The bare comparison with the double-equals isn't quite going to work.

classes/image_types_tegra.bbclass
```diff
@@ -345,7 +348,7 @@ copy_dtbs() {
             fi
             bbnote "Copying EXTERNAL_KERNEL_DEVICETREE entry $dtb to $destination"
             cp -L "${EXTERNAL_KERNEL_DEVICETREE}/$dtb" $destination/$dtbf
-	    if ${USE_UEFI_SIGNED_FILES}; then
+	    if ${TEGRA_UEFI_USE_SIGNED_FILES} == "true"; then
```
madisongh, line 351: See above

classes/image_types_tegra.bbclass
```diff
@@ -360,7 +363,7 @@ copy_dtb_overlays() {
     if [ -n "${IMAGE_TEGRAFLASH_INITRD_FLASHER}" ]; then
         extraoverlays="$extraoverlays L4TConfiguration-rcmboot.dtbo"
     fi
-    if ${USE_UEFI_SIGNED_FILES}; then
+    if ${TEGRA_UEFI_USE_SIGNED_FILES} == "true"; then
```
madisongh, line 366: See above

madisongh (changes requested):

classes/image_types_tegra.bbclass
Lexmark-chad: Well, that was rather sloppy of me.  I'll have a fix shortly.

madisongh:
Looks good now.  Thanks, @Lexmark-chad !

ichergui:
@Lexmark-chad @madisongh Is it possible to have these changes in our `master` branch ?

madisongh:
I have already put them in master and mickledore.

PR #1386, opened 2023-10-01
WIP: cleaning up duplicated code and settings

madisongh:
Several sets of changes here to reduce some of the duplication we had in the layer, and other miscellaneous cleanup to make things easier to maintain/extend in the future:

* Move the list of boot firmware binaries/files that were separately defined in the `tegra-boofiles` recipe and in the `image_types_tegra` class to a single per-SoC definition in the SoC-specific machine configurations.
* Clean up some per-SoC task appends in the `tegra-bootfiles` recipe
* Merge the per-SoC flash helper scripts to a single `tegra-flash-helper.sh` script, which contains logic for handling both chips
* Remove checking/handling of the `fuselevel` environment variable from the helper scripts, since public L4T releases *only* provide support for production chips.
* Switch from checked-in `flashvars` files to settings for those variables that are defined in the machine configurations


madisongh:
Further consolidation:

* Moved a couple of the `flashvars` file additions out of `image_types_tegra.bbclass` and into the `tegra-flashvars` recipe
* Unified the `create_tegraflash_pkg` function in `image_types_tegra.bbclass` so it works for both SoC families
* Hoisted the directory-population code to a common function shared between `create_tegraflash_pkg` and `oe_make_bup_payload`

Also renamed our `classes` directory to `classes-recipe`, since all classes in the layer are only used in recipe scope.

madisongh:
I've now broken out the staging of the EKB/EKS image from the tegra-bootfiles recipe as well, to make it easier to replace the default copy from L4T with a custom copy.

I've tested builds, including BUP generation, across all machines, and have tested flashing with AGX Orin, Orin NX, AGX Xavier, and Xavier NX kits.  Have also tested in-build signing and post-build signing, flashing a secured Xavier NX.  Will now move this out of draft.

ichergui (approved):
Looks great @madisongh 
Thanks a lot :slightly_smiling_face:

quaresmajose:
the https://github.com/OE4T/meta-tegra/commit/733b97877941d2a037cdea70df9014751b3bd3ad was not backported to the kirkstone branch and I suspect it should have been too. What is your opinion, am I right in my analysis?

madisongh:
master has the unified helper script that always sets `fuselevel_production`.  The helper scripts in kirkstone don't do this, which is why the `fuselevel` variable is still passed to them from `initrd-flash`.

quaresmajose:
Thanks for the clarification and sorry for the noise on this old subject

PR #1391, opened 2023-10-04
recipes-bsp/uefi: make UefiUpdateSecurityKeys optional

Lexmark-chad:
I took a stab at making this optional.  It builds with a db key and cert locally defined, but it is untested as I'm not quite setup to test that yet.


Lexmark-chad:
I think there is more work to do here.  I see that `tegra_bootcontrol_overlay_list` was recently re-worked and I think there is more cleanup that can be done.

Lexmark-chad:
I think this is ready for review now.

ichergui:
I will review it and will share my thoughts

madisongh:
I have manually merged this with a small tweak to the overlay inclusion logic in image_types_tegra so the update key overlay still only gets included for update payloads.

madisongh:
Thanks, @Lexmark-chad !

PR #1392, opened 2023-10-07
initrd-flash: provision to use separate dtb file

atharvanan1:
- initrd-flash requires `usb2-0` in `otg` mode, this provides a way to flash without losing ability to customize device-specific devicetree.


atharvanan1:
This is probably the wrong fix for this. Turns out, using `DTBFILE` in `.env.initrd-flash` ends up with device getting flashed with said DTBFILE.

PR #1394, opened 2023-10-08
initrd-flash: provision to use separate dtb file

atharvanan1:
- initrd-flash requires usb2-0 in otg mode, this provides a way to flash without losing ability to customize device-specific devicetree.


madisongh:
@atharvanan1 Can you provide more info on exactly what was missing from/changed in your device tree that caused the flashing problem?

classes/image_types_tegra.bbclass
```diff
@@ -32,6 +32,7 @@ TEGRA_BUPGEN_STRIP_IMG_NAMES ??= ""
 TEGRA_BUPGEN_STRIP_CMD ??= "${@tegraflash_bupgen_strip_cmd(d)}"
 
 DTBFILE ?= "${@os.path.basename(d.getVar('KERNEL_DEVICETREE').split()[0])}"
+INITRD_DTBFILE ?= "${@os.path.basename(d.getVar('INITRD_KERNEL_DEVICETREE').split()[0]) if d.getVar('INITRD_KERNEL_DEVICETREE') else '${DTBFILE}'}"
```
madisongh, line 35: I think it's OK to limit the `INITRD_KERNEL_DEVICETREE` variable to naming just one device tree, so the split is probably not needed here.  Also, it's probably better to default to a null string instead of ${DTBFILE} if it's not defined at all.

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -243,6 +243,7 @@ sign_binaries() {
     fi
     if [ "$CHIPID" = "0x18" ]; then
 	"$here/tegra-signimage-helper" --chip "$CHIPID" --nosplit --type kernel initrd-flash.img
+	"$here/tegra-signimage-helper" --chip "$CHIPID" --nosplit --type kernel-dtb ${INITRD_DTBFILE}
```
madisongh, line 246: Let's let INITRD_DTBFILE be null if there is no separate device tree defined for initrd flashing, and just add a check for it being non-null here to see if the extra signing is needed.

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -252,11 +253,12 @@ prepare_for_rcm_boot() {
     if [ $have_odmsign_func -eq 1 ]; then
 	if [ -e "rcmbootcmd.txt" ]; then
 	    sed -e"s,boot.img,initrd-flash.img," rcmbootcmd.txt > rcm-boot.sh
+	    sed -e"s,${DTBFILE},${INITRD_DTBFILE}," rcmbootcmd.txt > rcm-boot.sh
 	    chmod +x rcm-boot.sh
 	elif [ "$CHIPID" = "0x18" ]; then
 	    local binsfx=".encrypt"
 	    local ksfx=".encrypt"
-	    local kdtbfilebase=$(basename kernel_$DTBFILE .dtb)
+	    local kdtbfilebase=$(basename $INITRD_DTBFILE .dtb)
```
madisongh, line 261: Here's where we should fall back to `kernel_$DTBFILE` if `$INITRD_DTBFILE` is null.

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -272,7 +274,7 @@ prepare_for_rcm_boot() {
 					    --bins kernel=initrd-flash_sigheader.img${ksfx},kernel_dtb=${kdtbfilebase}_sigheader.dtb${ksfx},sce_fw=camera-rtcpu-sce_sigheader.img${binsfx},adsp_fw=adsp-fw_sigheader.bin${binsfx} \
 					    --cmd rcmboot --add="--securedev" flash_signed.sh || return 1
 	else
-	    "$here/rewrite-tegraflash-args" -o rcm-boot.sh --bins kernel=initrd-flash.img,kernel_dtb=kernel_$DTBFILE --cmd rcmboot --add="--securedev" flash_signed.sh || return 1
+	    "$here/rewrite-tegraflash-args" -o rcm-boot.sh --bins kernel=initrd-flash.img,kernel_dtb=${INITRD_DTBFILE} --cmd rcmboot --add="--securedev" flash_signed.sh || return 1
```
madisongh, line 277: Likewise here - use the normal kernel DTB unless the extra one is defined.

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -291,7 +293,7 @@ run_rcm_boot() {
 	./rcm-boot.sh || return 1
     else
 	local cmd="MACHINE=$MACHINE BOARDID=$BOARDID FAB=$FAB BOARDSKU=$BOARDSKU BOARDREV=$BOARDREV CHIPREV=$CHIPREV fuselevel=$fuselevel \
-	       $helper --rcm-boot $signargs flash.xml.in $DTBFILE $EMMC_BCTS $ODMDATA initrd-flash.img $ROOTFS_IMAGE"
+	   $helper --rcm-boot $signargs flash.xml.in $INITRD_DTBFILE $EMMC_BCTS $ODMDATA initrd-flash.img $ROOTFS_IMAGE"
```
madisongh, line 296: and again here.

madisongh (changes requested):
I think I see where you're going with this, and the idea looks OK to me.  Using `INITRD_` as the prefix on the new variables is a bit misleading, so it would be better to use `INITRD_FLASHER_` or maybe `RCMBOOT_INITRD_` or something else that indicates it's specific to the initrd flash mechanism.

And see in-line for some suggestions on deferring the fallback to using the normal kernel dtb to later in the process.  I mainly concerned about not signing the same file twice when secure boot is enabled.

I also need to go through other parts of the layer to double-check that the added DTB will get handled appropriately for the during-build and post-build signing use cases, as well as making sure it does not get included BUP payloads.

atharvanan1:
> Can you provide more info on exactly what was missing from/changed in your device tree that caused the flashing problem?

My device-tree was missing https://github.com/OE4T/linux-tegra-4.9/blob/oe4t-patches-l4t-r32.7.4/nvidia/platform/t19x/jakku/kernel-dts/common/tegra194-p3668-common.dtsi#L127 mode, we are setting it to host mode. Along with that, few other places where xusb and external-connections are modified.

classes/image_types_tegra.bbclass
atharvanan1: Makes sense. This is resolved.

atharvanan1:
> And see in-line for some suggestions on deferring the fallback to using the normal kernel dtb to later in the process. I mainly concerned about not signing the same file twice when secure boot is enabled.

Thanks for your valuable feedback!

atharvanan1:
Initial tests loks good for non-secureboot non-signed stuff.

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -243,20 +243,26 @@ sign_binaries() {
...
+	if [ ! -z "${INITRD_FLASHER_DTBFILE}" ]; then
+	    "$here/tegra-signimage-helper" --chip "$CHIPID" --nosplit --type kernel-dtb ${INITRD_FLASHER_DTBFILE}
+	fi
     fi
     . ./boardvars.sh
     return 0
 }
 
 prepare_for_rcm_boot() {
+    if [ -z "${INITRD_FLASHER_DTBFILE}" ]; then
```
madisongh, line 255: Redefining the global variable as local here could be confusing later.  Could you do something like
```
local flasherdtb="${INITRD_FLASHER_DTBFILE:-kernel_${DTBFILE}}"
```
instead, and then use `$flasherdtb` below?

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -272,13 +278,16 @@ prepare_for_rcm_boot() {
...
 	else
-	    "$here/rewrite-tegraflash-args" -o rcm-boot.sh --bins kernel=initrd-flash.img,kernel_dtb=kernel_$DTBFILE --cmd rcmboot --add="--securedev" flash_signed.sh || return 1
+	    "$here/rewrite-tegraflash-args" -o rcm-boot.sh --bins kernel=initrd-flash.img,kernel_dtb=${INITRD_FLASHER_DTBFILE} --cmd rcmboot --add="--securedev" flash_signed.sh || return 1
 	fi
 	chmod +x rcm-boot.sh
     fi
 }
 
 run_rcm_boot() {
+    if [ -z "${INITRD_FLASHER_DTBFILE}" ]; then
```
madisongh, line 288: See above

madisongh (changes requested):
Looks good overall, just a minor change needed (see in-line).  Thanks!

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
atharvanan1: Should be resolved now.

madisongh (approved):

PR #1404, opened 2023-10-18
Dunfell: tegra-boot-tools: update to 3.1.2

paroque28:
tegra-boot-tools: update to 3.1.2


madisongh:
I'm nervous about making a big jump like this in dunfell.  Is there a specific fix in the tools that you're looking for?  It would be safer to back-port that as a patch to the 2.5.x version we're using in this branch.

paroque28:
@madisongh I was looking into integrating https://github.com/OE4T/tegra-boot-tools/pull/30 . But I've already integrated this change into my OS after some testing. We maintain our own branch for Dunfell, so if you feel like this is a big change, and no one else needs it right now, then we can close the PR.

madisongh:
@paroque28 That fix is a good one to back-port, so I've issued a 2.5.3 version of the tools with that fix added and updated the recipe here for the new version.

PR #1405, opened 2023-10-20
Add support for Qt 6

sapiippo:
Add separate bbappends for Qt 6 (https://code.qt.io/cgit/yocto/meta-qt6.git). Enable the EGLFS EGLDevice platform support needed for NVIDIA boards. Disable GBM for qtbase, otherwise Qt tries to use wrong eglfs backend.

### DIFF
--- a/external/qt6-layer/recipes-qt/qt6/qtbase_%.bbappend
+++ b/external/qt6-layer/recipes-qt/qt6/qtbase_%.bbappend
@@ -1,0 +1,4 @@
+PACKAGECONFIG_GRAPHICS:append:tegra = " ${@bb.utils.contains('DISTRO_FEATURES', 'x11', '', 'eglfs-egldevice', d)}"
+
+QT_QPA_DEFAULT_EGLFS_INTEGRATION ?= "${@bb.utils.contains('PREFERRED_RPROVIDER_tegra-gbm-backend', 'tegra-libraries-gbm-backend', 'eglfs_kms_egldevice', 'eglfs_kms', d)}"
+EXTRA_OECMAKE:append:tegra = " -DQT_QPA_DEFAULT_EGLFS_INTEGRATION=${QT_QPA_DEFAULT_EGLFS_INTEGRATION}"


kekiefer:
For meta-qt6, we should be overriding `PACKAGECONFIG_GRAPHICS` here, so that derivative layers can more easily specify their own settings (for example, using the gbm backend).

Also in this layer we should apply platform-specific overrides.

I have one question as I'm not an X11 user, but the [default graphics packageconfig in meta-qt6 for x11](https://code.qt.io/cgit/yocto/meta-qt6.git/tree/recipes-qt/qt6/qtbase_git.bb#n51) only enables `gl` and not `kms` etc. Here I'm assuming that the qt build could fail if we don't filter, and I could be wrong about this - in which case we could leave it alone. In total, the following makes sense:

```
PACKAGECONFIG_GRAPHICS:remove:tegra = "gbm"
PACKAGECONFIG_GRAPHICS:append:tegra = "bb.utils.contains('DISTRO_FEATURES', 'x11', '', ' eglfs-egldevice', d)"
```

While we're here - just in case anyone is seeing this and wondering about gbm support. In fact, Qt's gbm does work, with the non-nvidia gbm backend:

```
PREFERRED_RPROVIDER_tegra-gbm-backend = "tegra-udrm-gbm"
```

I did need a small patch to create contexts:

```
 src/gui/opengl/platform/egl/qeglplatformcontext.cpp | 2 --
 1 file changed, 2 deletions(-)

diff --git a/src/gui/opengl/platform/egl/qeglplatformcontext.cpp b/src/gui/opengl/platform/egl/qeglplatformcontext.cpp
index 0a6e83e875..734ec74c8c 100644
--- a/src/gui/opengl/platform/egl/qeglplatformcontext.cpp
+++ b/src/gui/opengl/platform/egl/qeglplatformcontext.cpp
@@ -95,8 +95,6 @@ QEGLPlatformContext::QEGLPlatformContext(const QSurfaceFormat &format, QPlatform
     contextAttrs.append(format.majorVersion());
     const bool hasKHRCreateContext = q_hasEglExtension(m_eglDisplay, "EGL_KHR_create_context");
     if (hasKHRCreateContext) {
-        contextAttrs.append(EGL_CONTEXT_MINOR_VERSION_KHR);
-        contextAttrs.append(format.minorVersion());
         int flags = 0;
         // The debug bit is supported both for OpenGL and OpenGL ES.
         if (format.testOption(QSurfaceFormat::DebugContext))
```

I'm not really sure if there's a way to include both the `gbm` and `eglfs-egldevice` backends, and more simply change the priority of them, or allow a selection in this override to choose one or the other, but I wish we could.

ichergui:
Hey @sapiippo @kekiefer 
I totally agree with @kekiefer  

@sapiippo: Did you tried `QT6` on `Xavier` or `Orin` series or both ? 
I did try `Q6` on `Orin` series and I was happening some issues.

sapiippo:
Absolute right about the x11, I had this prepared already, but in haste pushed older commit.

I wasn't aware of the alternative gbm backend, but there is way to keep both if we want.
We could have following bits to selected the preferred Qt backend

```
QT_QPA_DEFAULT_EGLFS_INTEGRATION ?= "${@bb.utils.contains('PREFERRED_RPROVIDER_tegra-gbm-backend', 'tegra-libraries-gbm-backend', 'eglfs_kms_egldevice', 'eglfs_kms', d)}"
EXTRA_OECMAKE:append:tegra = " -DQT_QPA_DEFAULT_EGLFS_INTEGRATION=${QT_QPA_DEFAULT_EGLFS_INTEGRATION}"
```


@ichergui: We've been testing Qt on AGX Orin devkit. What issues did you run into?

kekiefer:
> I wasn't aware of the alternative gbm backend, but there is way to keep both if we want.
> We could have following bits to selected the preferred Qt backend
> `QT_QPA_DEFAULT_EGLFS_INTEGRATION ?= `

I like this approach. Distro layers won't require a qtbase bbappend to choose the backend.

kekiefer:
lgtm

kekiefer (approved):

PR #1411, opened 2023-11-09
tegra-uefi-capsule: add TEGRA_UEFI_CAPSULE_SIGNING_EXTRA_DEPS

Lexmark-chad:
- The signing server version of the class that signs uefi capsules has some task dependencies that must be met.

### DIFF
--- a/recipes-bsp/uefi/tegra-uefi-capsules_35.4.1.bb
+++ b/recipes-bsp/uefi/tegra-uefi-capsules_35.4.1.bb
@@ -7,6 +7,8 @@ inherit tegra-bup deploy
 TEGRA_UEFI_CAPSULE_SIGNING_CLASS ??= "tegra-uefi-capsule-signing"
 inherit ${TEGRA_UEFI_CAPSULE_SIGNING_CLASS}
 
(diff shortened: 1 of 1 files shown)


Lexmark-chad:
More work to do.  The way dependencies are expressed in meta-tegra/classes/tegra-uefi-signing.bbclass and meta-tegra-signing-server/classes/digsigserver-uefi-capsule-signing.bbclass need to be made consistent.

Lexmark-chad:
I considered changing to this:

```
do_compile[depends] += "${TEGRA_UEFI_CAPSULE_SIGNING_EXTRA_DEPS}"
```
by defining `TEGRA_UEFI_CAPSULE_SIGNING_EXTRA_DEPS` in `recipes-bsp/uefi/tegra-uefi-capsules_35.4.1.bb` and moving `${@bup_dependency(d)}` there as well as to the signing server equivalent class.  But, I think it is correct that `bup_dependency` stay in the recipe since it is a dependency of the recipe itself and not of the signing class.  Only class specific dependencies should be defined in `TEGRA_UEFI_CAPSULE_SIGNING_EXTRA_DEPS`.  I think what I've submitted is appropriate.

PR #1424, opened 2023-11-27
tegra-helper-scripts: Fix error handling errors when flashing + Formatting

synther:
Function calls under `| tee -a $logfile` don't propagate an error code. Some flashing errors are ignored without the fix.

And a separate commit for some formatting.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
@@ -116,6 +116,7 @@ wait_for_rcm() {
 }
 
 # shellcheck disable=SC2120  # Don't warn about no arguments are passed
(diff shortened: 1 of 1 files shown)


quaresmajose:
Good finding!
Thanks for your patch

This issue can be also fixed changing the shell options with `set`:
```
#!/usr/bin/env bash
set -o errexit
set -o pipefail
...
```
https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html

madisongh:
`errexit` is probably too much, but I definitely missed `set -o pipefail` in the script.  Setting that should simplify things, eliminating the need for extra wrapper functions.

ichergui:
@madisongh  should we close this PR ?

ichergui:
Didn't get answer from a while.
closing this PR.

PR #1427, opened 2023-11-30
libnvidia-container-tools: rdepend on ldconfig

ricardosalveti:
nvidia-container-cli requires access to the host ldconfig at runtime, so have it as a runtime dependency explicitly.


ichergui:
Hi @ricardosalveti 
Thanks for the PR.
Could you please share the context of this change ?
I find it strange that **ldconfig** is needed at runtime.

ricardosalveti:
> Hi @ricardosalveti Thanks for the PR. Could you please share the context of this change ? I find it strange that **ldconfig** is needed at runtime.

It was also surprising to me, but we found as we tried to reduce the image size which ended up not having ldconfig installed at all:

```
fio@jetson-agx-xavier-devkit:~$ docker run -it --rm --net=host --runtime nvidia -e DISPLAY=$DISPLAY -v /tmp/.X11-unix/:/tmp/.X11-unix nvcr.io/nvidia/l4t-cuda:11.4.14-runtime
docker: Error response from daemon: failed to create task for container: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: error during container init: error running hook #0: error running hook: exit status 1, stdout: , stderr: nvidia-container-cli: ldcache error: open failed: /sbin/ldconfig: no such file or directory: unknown.
```

Looking further there indeed logic to update the ldcache from the host as cli gets executed, as you can find at https://github.com/NVIDIA/libnvidia-container/blob/5e135c17d6dbae861ec343e9a8d3a0d2af758a4f/src/nvc_ldcache.c#L348.

PR #1449, opened 2023-12-26
external/virtualization-layer: update container support recipes

madisongh:
This brings container support recipes up to the versions used in L4T R35.4.1/JetPack 5.1.2.

* Rename libnvidia-container-tools recipe to the more accurate 'libnvidia-container', and update to v1.10.0 (with patch consolidiation and refresh)
* Fix the makefiles for libnvidia-container and libnvidia-container-jetson to eliminate the build paths that were getting embedded in the RPC files
* Include the full set of DEBUG_PREFIX_MAP flags when building libnvidia-container, which fixes more embedded build paths
* Update nvidia-container-toolkit to v1.11.0
* Switch to static linking for nvidia-container-toolkit to work around panic on startup in nvidia-ctk
* Hard-code the toolkit configuration to "legacy" mode, which uses the older prestart hook method for mapping the paths in the CSV file, until our hacks for the passthrough tree and for mapping the Debian/Ubuntu multi-arch library paths to the normal host paths get migrated to the Go code in nvidia-container-toolkit.
* Update nvidia-docker to v2.11.0


madisongh:
See discussion at #1426

PR #1454, opened 2024-01-08
external/openembedded-layer: tensorrt-plugins: fixes for new protobuf

kekiefer:
Due to updates to protobuf and gcc, make some updates to the build. Namely, this means using C++17 as required by protobuf and making sure we link to the system protobuf.

Applicable to master and nanbield


ichergui:
Thanks @kekiefer for the PR. I didn't test that for awhile.

PR #1472, opened 2024-02-08
Updates for L4T R36.2.0 DP

ichergui:


madisongh (commented):
The only question I have is around the `libnvjpeg` recipe... we already install `libnvjpeg` from the BSP in the `tegra-libraries-multimedia` recipe.  Is there any conflict between that one and the one installed by this new recipe?

Otherwise, everything looks good

ichergui:
Thanks @madisongh. 
Good catch. Let me double check that and will get back to you.

ibstewart:
Ah sorry, I missed that the libnvjpeg was installed elsewhere. I added that recipe because the dGPU CUDA package also includes libnvjpeg and so this allowed the dGPU recipe to use a bbappend for it as I do for the other CUDA packages. Feel free to remove the libnvjpeg recipe and I can revisit the dGPU recipe for this.

ichergui:
@madisongh  I removed the nvjpeg recipe. 
Thanks

PR #1475, opened 2024-02-12
nvidia-container-toolkit: Add version and gitCommit to nvidia-container-runtime --version

dchvs:
This feature enables the display of the version information for the `nvidia-container-toolkit` binaries. After this change, the binary versioning will appear as follows:

```bash
nvidia-container-runtime --version
NVIDIA Container Runtime version 1.11.0
commit: 670+d9de4a0
.
.
.
```


ichergui (approved):
LGTM

PR #1477, opened 2024-02-13
external/openembedded-layer: opencv: 4.8 -> 4.9

quaresmajose:


madisongh:
Thanks... I've pushed a more complete update, which was a bit more involved than just renaming the bbappend.

quaresmajose:
Basically I just fixed the bitbake parsing, sorry for not update the patches.
Thanks

PR #1488, opened 2024-02-21
AGX Xavier missing fab

lfdmn:
The 401 fab was added in dunfell-l4t-r32.5.0 but missing in this branch


madisongh:
@lfdmn I just cherry-picked over your patch from the 32.5 branch rather than merging this.

PR #1511, opened 2024-03-21
Fix override syntax error in libglvnd-tegra

lms-ts:


madisongh:
Thanks!

ichergui:
Thanks @lms-ts

PR #1512, opened 2024-03-21
Fix override syntax errors

lms-ts:
Found additional override syntax errors by grepping for them. That should probably be all.


ichergui (approved):
LGTM. Thanks @lms-ts

madisongh:
Thanks... I'd appreciate it if you could break this into two commits specifically mentioning each recipe in the commit message, and add the requisite `Signed-off-by:` line to each.

lms-ts:
Done.

PR #1527, opened 2024-04-11
Refactor UEFI capsule generation

igoropaniuk:
Refactor UEFI capsule generation. Drop BUP-specific recipes and move all logic to UEFI capsule
generation recipe, as suggested in [1]

[1] https://github.com/OE4T/meta-tegra/discussions/1492


ichergui:
Thanks @igoropaniuk for the PR. I will check the changes and get back to you.

ichergui (approved):
LGTM.
Did you try any update mechanism ?

igoropaniuk:
@ichergui not really, I've verified only that cap files  (with the same size as without these changes) are generated and deployed. I can test update mechanism if needed, but I haven't touched anything in the way how that capsule updates are processed in runtime

ichergui:
@igoropaniuk Please do

recipes-bsp/uefi/tegra-uefi-capsules_35.5.0.bb
```diff
@@ -2,54 +2,67 @@ DESCRIPTION = "Generate UEFI capsules for bup paylods"
...
-    sign_uefi_capsules
+
+    if [ ! -z "${INITRAMFS_IMAGE}" ]; then
+        initramfs_symlink_name=${INITRAMFS_IMAGE}-${MACHINE}
+        # Generate BUP images
+        oe_make_bup_payload ${DEPLOY_DIR_IMAGE}/${initramfs_symlink_name}.cboot
+
+        # Generate UEFI capsules
+        generate_uefi_capsule ${WORKDIR}/bup-payload/bl_only_payload tegra-bl.cap
+        generate_uefi_capsule ${WORKDIR}/bup-payload/kernel_only_payload tegra-kernel.cap
```
madisongh, line 27: This should be conditional on the existence of the kernel-only payload.  It's technically possible to remove the kernel and DTB partitions from a storage layout, in which case no such payload would get generated.

recipes-bsp/uefi/tegra-uefi-capsules_35.5.0.bb
```diff
@@ -2,54 +2,67 @@ DESCRIPTION = "Generate UEFI capsules for bup paylods"
...
 
 addtask deploy after do_install
-
+do_compile[depends] += "virtual/kernel:do_deploy tegra-flashtools-native:do_populate_sysroot dtc-native:do_populate_sysroot"
+do_compile[depends] += "python3-pyyaml-native:do_populate_sysroot"
+do_compile[depends] += "tegra-redundant-boot-rollback:do_populate_sysroot tegra-bootfiles:do_populate_sysroot"
+do_compile[depends] += "coreutils-native:do_populate_sysroot ${TEGRA_ESP_IMAGE}:do_image_complete virtual/secure-os:do_deploy"
+do_compile[depends] += "${TEGRA_SIGNING_EXTRA_DEPS}"
 do_compile[depends] += "${@bup_dependency(d)} ${TEGRA_UEFI_CAPSULE_SIGNING_EXTRA_DEPS}"
+addtask compile before do_build
```
madisongh, line 68: Is this needed?  do_compile should already be in the standard list of tasks for every recipe.

recipes-bsp/uefi/tegra-uefi-capsules_35.5.0.bb
```diff
@@ -2,54 +2,67 @@ DESCRIPTION = "Generate UEFI capsules for bup paylods"
...
-DEPENDS += "tegra-bup-payload"
+TEGRA_SIGNING_EXTRA_DEPS ??= ""
 
 GUID:tegra194 ?= "be3f5d68-7654-4ed2-838c-2a2faf901a78"
 GUID:tegra234 ?= "bf0d4599-20d4-414e-b2c5-3595b1cda402"
 
 do_compile() {
-    sign_uefi_capsules
+
+    if [ ! -z "${INITRAMFS_IMAGE}" ]; then
```
madisongh, line 20: Everything below here should be run regardless of whether INITRAMFS_IMAGE is defined or not.  Instead, the parameter passed to `oe_make_bup_payload` will differ, I think, depending on whether we have a bare kernel/kernel with bundled initramfs or a kernel with separate initrd.  Either case would still be a cboot file, but the base file name would be different.

madisongh (changes requested):
Looks like a good start on this, thanks!

A couple of additional cleanups: the `tegra-bup` bbclass can be removed with these changes, and the `bup_dependency()` function reference could just be replaced with a dependency on the do_image_complete task for the initrd if the build is using a separate initrd image.

Also, the `oe_make_bup_payload()` function could be moved out of `image_types_tegra.bbclass` and directly into the `do_compile` function of the capsules recipe, as this would be the only user of it.

madisongh:
@Lexmark-chad It would be good for you to review this as well, if you can

classes-recipe/tegra-uefi-capsule-signing.bbclass
```diff
@@ -16,9 +16,11 @@ UEFI_CAPSULE_TRUSTED_PUBLIC_CERT ?= "${PYTHON_BASETOOLS}/Pkcs7Sign/TestRoot.pub.
 
 # Override this function to have a secure signing server
 # perform the capsule signing.
-sign_uefi_capsules() {
+generate_uefi_capsule() {
```
Lexmark-chad, line 19: Structurally I like what has been done, but it is going to introduce breaking changes for anyone who has overridden `sign_uefi_capsules`.  The effect of renaming the class function is to not invoke someone's overridden signed implementation leaving capsules signed with default keys.  Capsule updates will fail.

recipes-bsp/uefi/tegra-uefi-capsules_35.5.0.bb
kekiefer: My understanding is that the kernel payload as a capsule doesn't work to start with. If this is correct, it might make sense to omit it altogether.

classes-recipe/tegra-uefi-capsule-signing.bbclass
Lexmark-chad: There are several things that make this a breaking change:  the class function name change, class function is parameterized, and the location of BUP images.  All good changes, but breaking changes none-the-less.  Chalk that up to maybe not the best first pass implementation.  Just deploying the BUP images to DEPLOY_DIR won't solve anything because the overridden class function won't get invoked.  It could be decided that maybe not many have even adopted an external signing server yet for generating & signing the capsules (it just went last year) in which case the PR could be merged.  Not sure the best way to try to communicate the change other than announcing on the gitter channel so people don't get caught by surprise.

ichergui:
@igoropaniuk  Any update about this PR ?

igoropaniuk:
@ichergui I'm currently traveling, will get back to it when I'm back (in a week)

ichergui:
Hi @igoropaniuk 
Please update this PR when you have chance.
Let us know if you need help.

ichergui:
Hi @igoropaniuk 
Thanks for the updates. 
Could you please rebase your banch on top of master then push again ?

recipes-bsp/uefi/tegra-uefi-capsules_35.5.0.bb
```diff
@@ -11,45 +11,84 @@ TEGRA_UEFI_CAPSULE_SIGNING_EXTRA_DEPS ??= ""
...
+	    BUP_generator.py --contents --check $bup
+    done
+    mv ${WORKDIR}/bup-payload/${BUP_PAYLOAD_DIR}/* .
+    cd "$oldwd"
+
+    # Deploy BUP payloads
+    for f in ${WORKDIR}/bup-payload/*_only_payload; do
+        [ -e $f ] || continue
+        sfx=$(basename $f _payload)
+        # We need to have bup-payload images to be deployed to ${DEPLOY_DIR_IMAGE} before
```
igoropaniuk, line 45: what explained in this comment - is how it was done in the previous implementation, where CONVERSION_CMD was used. TBH I'd prefer to split that recipe on two, and add compile time dependency, so capsule-generation recipe depends bup-recipe

This all applies if we still want to keep `sign_uefi_capsules` function implementation at it is (as it expects BUP images to be deployed to DEPLOY_DIR_IMAGE in advance)

igoropaniuk: @madisongh @Lexmark-chad please advise what action needs to be taken

igoropaniuk:
@ichergui I've pushed rebased version, but looks like sanity checks are still failing (and issues reported are not related to my changes):

```
ERROR: /scratch/buildworker/meta-tegra-master-checklayer/build/poky/meta-tegra/recipes-devtools/cuda/nativesdk-cuda-environment_1.0.bb: Using S = ${WORKDIR} is no longer supported
ERROR: /scratch/buildworker/meta-tegra-master-checklayer/build/poky/meta-tegra/recipes-bsp/tegra-binaries/tegra-helper-scripts-native_35.5.0.bb: Using S = ${WORKDIR} is no longer supported
ERROR: Parsing halted due to errors, see error messages above
```

madisongh:
Yeah, the changes for disallowing `S = "${WORKDIR}" have gone into OE-Core master... I'll try and make some time to work on that today.

recipes-bsp/uefi/tegra-uefi-capsules_35.5.0.bb
igoropaniuk: @madisongh sounds good to me. 

Should we still deploy BUP images to `${DEPLOYDIR}` in `do_deploy()` task, or they are not supposed to used anymore?

I was thinking about generating both BUP images and capsules in `${B}` dir, but then deploying all of them in `do_deploy()`

madisongh: BUPs aren't supposed to be used any more, so let's not deploy them, just the capsules.

ichergui: Totally agree with @madisongh. Let's use only capsules 
Thanks @igoropaniuk

igoropaniuk: done!

recipes-bsp/uefi/tegra-uefi-capsules_35.5.0.bb
```diff
@@ -11,45 +11,88 @@ TEGRA_UEFI_CAPSULE_SIGNING_EXTRA_DEPS ??= ""
...
+    # Generate UEFI capsules
     sign_uefi_capsules
+
+    # Check if capsules were generated successfully
+    if [ ! -e ${B}/tegra-bl.cap ]; then
+        bberror "${B}/tegra-bl.cap wasn't generated"
+    fi
+    if [ ! -e ${B}/tegra-kernel.cap ]; then
+        bberror "${B}/tegra-kernel.cap wasn't generated"
+    fi
```
madisongh, line 58: Almost... at least for the kernel one, the test should be something like `[ -e ${B}/${BUPFILENAME}.kernel.bup_payload -a ! -e ${B}/tegra-kernel.cap ]`.   There are cases where no kernel BUP will get created, and that's OK.  (No one should be using capsules to update their kernels anyway.)

madisongh (commented):
See in-line

recipes-bsp/uefi/tegra-uefi-capsules_35.5.0.bb
igoropaniuk: fixed, thanks @madisongh

igoropaniuk:
@ichergui @madisongh @Lexmark-chad @kekiefer do you have any additional comments/objections/suggestions? thanks

madisongh (approved):

madisongh:
I think we're good.  I'm running a couple of build tests in preparation for merging. Thanks!

ichergui (approved):
LGTM. Thanks @igoropaniuk

kekiefer (approved):

Lexmark-chad (approved):
New changes look good.

PR #1535, opened 2024-04-18
initrd-flash: Add custom extra init

mickeyproc:
This makes it possible for extra custom steps to take before the final reboot step of the initrd flashing process.

### DIFF
--- a/recipes-core/initrdscripts/tegra-flash-init/init-extra.sh
+++ b/recipes-core/initrdscripts/tegra-flash-init/init-extra.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+# Run custom extra init script(s) provided from seperate recipes.
+
+init_extra_dir="/init-extra.d"
+
+if [ -d $init_extra_dir ] && [ "$(ls -A $init_extra_dir)" ]; then
+  for init_extra in "$init_extra_dir"/*; do
+    ./"$init_extra" 2>&1 | tee "/tmp/flashpkg/flashpkg/logs/custom-extra-$(basename "$init_extra").log"
+  done
+else
+  echo "No init_extra was found, ignoring" > /tmp/flashpkg/flashpkg/logs/custom-extra.log
+fi
--- a/recipes-core/initrdscripts/tegra-flash-init_1.0.bb
+++ b/recipes-core/initrdscripts/tegra-flash-init_1.0.bb
@@ -4,6 +4,7 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda
 
 SRC_URI = "\
     file://init-flash.sh \
+    file://init-extra.sh \
     file://program-boot-device.sh \
     file://initrd-flash.scheme.in \
 "
@@ -14,6 +15,8 @@ S = "${WORKDIR}"
 
 do_install() {
     install -m 0755 ${WORKDIR}/init-flash.sh ${D}/init
+    install -m 0755 ${WORKDIR}/init-extra.sh ${D}/init-extra
+    install -m 0755 -d ${D}/init-extra.d
     install -m 0555 -d ${D}/proc ${D}/sys
     install -m 0755 -d ${D}/dev ${D}/mnt ${D}/run ${D}/usr
     install -m 1777 -d ${D}/tmp
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
@@ -371,6 +371,7 @@ generate_flash_package() {
 	echo "export-devices $ROOTFS_DEVICE" >> "$mnt/flashpkg/conf/command_sequence"
     fi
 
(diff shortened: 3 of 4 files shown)


recipes-core/initrdscripts/tegra-flash-init_1.0.bb
```diff
@@ -14,6 +14,9 @@ S = "${WORKDIR}"
 
 do_install() {
     install -m 0755 ${WORKDIR}/init-flash.sh ${D}/init
+    if [ -f "${EXTRA_INIT}" ]; then
```
madisongh, line 17: It wouldn't hurt to add a default initializer for this variable at the top of the recipe, i.e.
```
EXTRA_INIT ??= ""
```

madisongh (changes requested):
Thanks for this... If you could update your commit message with a Signed-off-by, per our [contributing guidelines](https://github.com/OE4T/meta-tegra/blob/master/CONTRIBUTING.md) and address the minor comment I have in-line, it will be good to go.

recipes-core/initrdscripts/tegra-flash-init_1.0.bb
madisongh: That's a good idea, but I'd move the logic into the `init_extra` script itself, which we could supply with the recipe, and users could just install their customizations into a directory, say, `/init_extra.d`.  That would eliminate the need to add a bbappend to this recipe altogether - the recipes to install the custom extension scripts could be separate, and would just need to be included in the `tegra-initrd-flash-initramfs` image. We already have a `TEGRA_INITRD_FLASH_INSTALL` variable (which defaults to empty) that could be set in a distro config file to add the custom extension packages that install those scripts.

mickeyproc: Good input. I'll get on it and update the PR.

recipes-core/initrdscripts/tegra-flash-init_1.0.bb
```diff
@@ -14,6 +14,8 @@ S = "${WORKDIR}"
 
 do_install() {
     install -m 0755 ${WORKDIR}/init-flash.sh ${D}/init
+    install -m 0755 ${WORKDIR}/init_extra.sh ${D}/init_extra
```
mickeyproc, line 17: The script to install is not included.
Also, there are inconsistency with underscore and hyphen, so that needs to be fixed as well.

mickeyproc: Updated PR with fixes. 
This has been verified with a clean build of core image with added recipe (custom-tegra-flash-init-extra):

```
DESCRIPTION = "Custom extra init for tegraflash initrd flashing"

LICENSE="CLOSED"

SRC_URI = "file://custom-extra-init.sh"

RDEPENDS:${PN} = " util-linux-findfs"

do_install () {
    install -m 0755 -d ${D}/init-extra.d
    install -m 0755 ${WORKDIR}/custom-extra-init.sh ${D}/init-extra.d/custom-extra-init
}

FILES:${PN} = "/"
```

And with distro config updated with:

```
TEGRA_INITRD_FLASH_INSTALL = "custom-tegra-flash-init-extra"
```

recipes-core/initrdscripts/tegra-flash-init/init-extra.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
madisongh, line 1: If you change this to `/bin/sh`, we shouldn't need to add bash to the RDEPENDS.  I don't think there are any bashisms in this script, so that should be OK.

recipes-core/initrdscripts/tegra-flash-init_1.0.bb
```diff
@@ -25,5 +28,5 @@ do_install() {
 }
 
 FILES:${PN} = "/"
-RDEPENDS:${PN} = "util-linux-blkdiscard tegra-flash-reboot mtd-utils e2fsprogs-mke2fs libusbgx-tegra-initrd-flash watchdog-keepalive gptfdisk"
+RDEPENDS:${PN} = "util-linux-blkdiscard tegra-flash-reboot mtd-utils e2fsprogs-mke2fs libusbgx-tegra-initrd-flash watchdog-keepalive gptfdisk bash"
```
madisongh, line 31: See above.  It would be preferable to keep bash out of the dependencies here.

madisongh (changes requested):
Generally looks good, thanks!  See in-line for one more small change to make.

recipes-core/initrdscripts/tegra-flash-init/init-extra.sh
mickeyproc: Makes sense. Removed the bash dependency.

recipes-core/initrdscripts/tegra-flash-init_1.0.bb
mickeyproc: Removed the bash dependency.

mickeyproc: Removed the bash dependency.

madisongh (approved):
LGTM.  Thanks!

PR #1539, opened 2024-04-22
l4t-extlinux-config.bbclass: add a missing boot prefix

liuming50:
A boot prefix '/boot/' is missing from FDT line, fix it.


madisongh:
Is there a reason why you can't specify the full path as the value for the variable, as is done for `UBOOT_EXTLINUX_KERNEL_IMAGE`?

liuming50:
> Is there a reason why you can't specify the full path as the value for the variable, as is done for `UBOOT_EXTLINUX_KERNEL_IMAGE`?

@madisongh you are absolutely right, closing this MR.

PR #1542, opened 2024-04-28
python3-kconfiglib-tegra: specify python package name properly

loblik:
Explictly specify python package name since this recipe does not follow python3-${PN} scheme.


ichergui (approved):
Thanks @loblik.

madisongh:
Thanks... I pushed the same fix for this yesterday, so it should already be taken care of.

PR #1543, opened 2024-04-28
p3768-0000-p3767-0001: allign machine config with stock SDK

loblik:
It seems official SDK manager is loading different BPFDTB for Orin NX 8Gb.

Also SKU is set 0001 as otherwise 0000 is used from included 16Gb Orin NX variant.


madisongh:
I had pushed the BPFDTB change yesterday... The TEGRA_BOARDSKU change is a good one, though, so if you could rebase this on top of latest, I'll merge that.  Thanks!

PR #1544, opened 2024-04-29
external/openembedded-layer: tensorrt-plugins: use CMAKE_CXX_STANDARD 17

kekiefer:
Previously the 17 standard was needed just for some plugins, and leaving 14 was fine for the rest of the project. With recent changes to absl, now we need the 17 standard for other parts as well.

This change builds on the do_configure:prepend sed fixup to include the CMakeLists.txt CMAKE_CXX_STANDARD being forced in a couple more places.


kekiefer:
This is applicable to master and scarthgap

PR #1548, opened 2024-05-07
p3768-0000-p3767-0001: set correct SKU for Orin NX 8Gb

loblik:
Board SKU fix was not picked for r36.3.0 branch.

So far this version seems to works fine for me.


madisongh:
Yep, there's still some cleanup on that branch I'm catching up on.  Thanks!

PR #1558, opened 2024-05-14
cuda-nvvm: update recipe

ichergui:
- update sha256sum
- restructure the DEB package to facilitate compile and install tasks

### DIFF
--- a/recipes-devtools/cuda/cuda-nvvm_12.2.140-1.bb
+++ b/recipes-devtools/cuda/cuda-nvvm_12.2.140-1.bb
@@ -3,7 +3,11 @@ L4T_DEB_GROUP = "cuda-nvcc"
 
 require cuda-shared-binaries.inc
 
-MAINSUM = "c6931d80a2c823e4c0a42dd1371e7bda613be2444393a67970e32f05dd0ccfd2"
+MAINSUM = "1ce8fac08a532f9202f1395a4b2c7be78cf9976663373cf563d79e7164b3b158"
 MAINSUM:x86-64 = "7cd46649dbad60b59860f0965d92eea87f783abfe2cbd2c9b49a0c8544161e3a"
+
+FILES:${PN} = "${prefix}/local/cuda-${CUDA_VERSION}"
+FILES:${PN}-dev = ""
+INSANE_SKIP:${PN} += "dev-so"
 
 BBCLASSEXTEND = "native nativesdk"


recipes-devtools/cuda/cuda-nvvm_12.2.140-1.bb
```diff
@@ -3,7 +3,14 @@ L4T_DEB_GROUP = "cuda-nvcc"
 
 require cuda-shared-binaries.inc
 
-MAINSUM = "c6931d80a2c823e4c0a42dd1371e7bda613be2444393a67970e32f05dd0ccfd2"
+MAINSUM = "1ce8fac08a532f9202f1395a4b2c7be78cf9976663373cf563d79e7164b3b158"
 MAINSUM:x86-64 = "7cd46649dbad60b59860f0965d92eea87f783abfe2cbd2c9b49a0c8544161e3a"
 
+do_compile:prepend() {
+    mv ${B}/usr/local/cuda-${CUDA_VERSION}/nvvm/* ${B}/usr/local/cuda-${CUDA_VERSION}/
```
quaresmajose, line 10: It will be better to move the entire directory but only if it exists:
```
if [ -d ${B}/usr/local/cuda-${CUDA_VERSION}/nvvm ]; then
    mv ${B}/usr/local/cuda-${CUDA_VERSION}/nvvm/* ${B}/usr/local/cuda-${CUDA_VERSION}/
    rm -rf ${B}/usr/local/cuda-${CUDA_VERSION}/nvvm
fi 
```

recipes-devtools/cuda/cuda-nvvm_12.2.140-1.bb
```diff
@@ -3,7 +3,14 @@ L4T_DEB_GROUP = "cuda-nvcc"
...
+    mv ${B}/usr/local/cuda-${CUDA_VERSION}/lib64 ${B}/usr/local/cuda-${CUDA_VERSION}/lib
```
quaresmajose, line 11: same check as abobe
```
if [ -d ${B}/usr/local/cuda-${CUDA_VERSION}/lib64 ]; then
    mv ${B}/usr/local/cuda-${CUDA_VERSION}/lib64 ${B}/usr/local/cuda-${CUDA_VERSION}/lib
fi 
```

quaresmajose (commented):
Using `mv` is very sensitive and easy to fail.
For example, this specific case will fail if you run the same task more than 1 time.

```
bitbake cuda-nvvm -c do_compile -f
```

recipes-devtools/cuda/cuda-nvvm_12.2.140-1.bb
madisongh: I don't understand the need for these moves in the first place.  Is that fixing a problem of some kind?

ichergui: Yes, the `nvvm` is not packaged same as the other cuda packages

madisongh: But it works fine as-is already (at least for x86_64).  What problem did you encounter that you're trying to fix with this change?

ichergui:
@madisongh  Kindly see the following logs 

```
ERROR: cuda-nvvm-12.2.140-1-r0 do_package: QA Issue: cuda-nvvm: Files/directories were installed but not shipped in any package:
  /usr
  /usr/local
  /usr/local/cuda-12.2
  /usr/local/cuda-12.2/nvvm
  /usr/local/cuda-12.2/nvvm/bin
  /usr/local/cuda-12.2/nvvm/libdevice
  /usr/local/cuda-12.2/nvvm/include
  /usr/local/cuda-12.2/nvvm/lib64
  /usr/local/cuda-12.2/nvvm/bin/cicc
  /usr/local/cuda-12.2/nvvm/libdevice/libdevice.10.bc
  /usr/local/cuda-12.2/nvvm/include/nvvm.h
  /usr/local/cuda-12.2/nvvm/lib64/libnvvm.so
  /usr/local/cuda-12.2/nvvm/lib64/libnvvm.so.4.0.0
  /usr/local/cuda-12.2/nvvm/lib64/libnvvm.so.4
Please set FILES such that these items are packaged. Alternatively if they are unneeded, avoid installing them or delete them within do_install.
cuda-nvvm: 14 installed and not shipped files. [installed-vs-shipped]
```
This is the issue I'm trying to fix. 
I can override the `install` task instead of adding `prepend` to `compile` task.

madisongh:
Better fix is to set `FILES:${PN}` similar to what's in the `cuda-nvcc` recipe.

ichergui:
I will update the recipe

PR #1561, opened 2024-05-15
Update recipes for L4T R36.3.0 GA

ichergui:


ichergui:
Closing since @madisongh updated a bunch of deb packages as well

PR #1569, opened 2024-05-30
tegra-uefi-capsules: add missing DTB_EXTRA_DEPS dependency

kekiefer:
The collection of dtb overlays used during tegraflash package creation can come from sources other than l4t or the kernel. Most commonly this is the virtual/dtb package, but it can be overridden, and is specified by the DTB_EXTRA_DEPS variable.

The migration from creating bup payloads to creating only capsules inadvertantly dropped this dependency, so this change adds it back.


madisongh:
Thanks

kekiefer:
Thanks, Matt. Please cherry-pick for scarthgap as well.

madisongh:
Yep, will do that shortly.

PR #1572, opened 2024-06-03
Add optional TBCDTB file

TB-ME:
Add the ability to users to specify a TBCDTB_FILE in the flashvars so that a different device tree can be used during UEFI than the kernel. This is helpful to stop other devices connected over UART from interrupting the boot process as listed here https://github.com/OE4T/meta-tegra/discussions/1510. I can only test on a tegra194 not a tegra234, meaning that this PR does not add the TBCDTB_FILE variable to tegra234 boards.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra194-flash-helper.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra194-flash-helper.sh
@@ -386,6 +386,11 @@ else
     cp "$dtb_file" "$kernel_dtbfile"
 fi
 
+tbc_dtb_file="$TBCDTB_FILE"
+if [ -z "$tbc_dtb_file" ]; then
+	tbc_dtb_file="$dtb_file"
+fi
+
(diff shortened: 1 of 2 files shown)


madisongh:
Basically looks OK, but it should cover the t234 platforms, too.  If you could update  the tegra234 helper script, even if you can't actually test it, that would help. I'll run through some manual tests before merging.

Also, could you add a Signed-off-by line to your commit message(s), per our [contributing guidelines](https://github.com/OE4T/meta-tegra/blob/eda302c9ed792e1d61912f33f9dddd09d97c5e5b/CONTRIBUTING.md)?

TB-ME:
I've updated the tegra234 helper script. I've added `bootloader_dtb` argument to `BINSARGS` because it was present in the tegra194 helper script.

ichergui (approved):
LGTM

madisongh:
I have merged this manually, with a small tweak for the t234 helper script. Thanks!

madisongh:
I've also adapted this change to the master, scarthgap, and scarthgap-l4t-r35.x branches.

TB-ME:
Thank you

PR #1575, opened 2024-06-06
tegra-uefi-capsules: conditionally install capsules

igoropaniuk:
Install capsules to rootfs only when TEGRA_UEFI_CAPSULE_INSTALL_DIR is set.


ichergui (approved):
LGTM

PR #1577, opened 2024-06-08
remove extra esp dependency

loblik:
Remove unnecessary dependency between firmware BUP and EFI system partition image. This creates circular dependency if you want to place your kerneln on ESP.


loblik:
Should this also get to master somehow? I'm not sure how development model and branching works here. Master branch is supposed to be used with next Yocto release, and scarthgap is stable branch mostly for bugfixes?

madisongh:
I usually take care of cherry-picking fixes to branches where they're needed.  I like to have fixes go into master first, but I know that's not easy for everyone to do.

Issue #1576, opened 2024-06-07
nvidia-kernel-oot: Conflicts with intree kernel modules of the same name


lms-ts:
There seems to be an error when installing intree kernel modules that are also RPROVIDED by nvidia-kernel-oot.

Relevant prerequisites:
Branch: scarthgap
```PREFERRED_PROVIDER_virtual_dtb = "nvidia-kernel-oot"```
```IMAGE_INSTALL += "kernel-modules```
Only ```nvidia-kernel-oot-base``` is explicitly added to the custom image.

When I try to build a custom image the following error occurs:
```
ERROR: jetson-image-custom-1.0-r0 do_rootfs: \
Unable to find package with name 'kernel-module-tegra-drm-5.15.136-l4t-r36.3-1009.9+g46cdb595bebc' \
in SPDX file /home/tschuster/project-lms/build/tmp/deploy/spdx/by-hash/jetson_agx_orin_orito/ \
sstate:nvidia-kernel-oot:jetson_agx_orin_orito-lms-linux:36.3.0:r0:jetson_agx_orin_orito:12: \
/nv-kernel-module-tegra-drm-5.15.136-l4t-r36.3-1009.9+g46cdb595bebc.spdx.json     
```
This error occurs for all conflicting modules. 

Possible solution:
By filling the ```RPROVIDES``` and ```RREPLACES``` variables of the oot modules, intree kernel modules of the same name are replaced and the above error is resolved.

nvidia-kernel-oot_36.3.0.bb:
```
-        d.appendVar('RPROVIDES:' + oot_pkg, ' ' + newprovides + ' ' + newprovides_virt)
+        newprovides_str = ' ' + newprovides + ' ' + newprovides_virt
+        d.appendVar('RPROVIDES:' + oot_pkg, newprovides_str)
+        d.appendVar('RREPLACES:' + oot_pkg, newprovides_str)
+        d.appendVar('RCONFLICTS:' + oot_pkg, newprovides_str)
```

madisongh:
I can't reproduce this in master, and I think that's because there's a fix for dependency handling in `create-spdx-2.2.bbclass` there that isn't in scarthgap.  Unfortunately, the fix depends on some bitbake changes that haven't made their way to scarthgap, so I can't easily test that hypothesis.

[...]

PR #1578, opened 2024-06-08
nvidia-kernel-oot: Set RREPLACES and RCONFLICTS variables to replace …

lms-ts:
https://github.com/OE4T/meta-tegra/issues/1576


madisongh:
I've implemented this a bit differently, but you should find that #1576 is fixed now.

PR #1579, opened 2024-06-11
tegra-flash-init: allow user to override script variables

loblik:
On custom board the QSPI flash can be under different name (like mtd1). Allow user to specify a device name by creating an environment file in bbappend.

### DIFF
--- a/recipes-core/initrdscripts/tegra-flash-init_1.0.bb
+++ b/recipes-core/initrdscripts/tegra-flash-init_1.0.bb
@@ -14,6 +14,13 @@ UNPACKDIR = "${S}"
 
 COMPATIBLE_MACHINE = "(tegra)"
 
+MTD_DEV ?= "${@d.getVar('OTABOOTDEV').replace('mtdblock','mtd')}"
+
+do_configure() {
+    sed -e's,@MTD_DEV@,${MTD_DEV},g' \
+        ${UNPACKDIR}/program-boot-device.sh > ${UNPACKDIR}/program-boot-device
+}
+
 do_install() {
     install -m 0755 ${UNPACKDIR}/init-flash.sh ${D}/init
     install -m 0755 ${UNPACKDIR}/init-extra.sh ${D}/init-extra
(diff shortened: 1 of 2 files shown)


loblik:
Not sure if this proper way to do this. I know there is .env.initrd-flash in flash packages, and arguments get loaded at runtime. 

But is this really needed? Is ramdisk supposed to be universal for all the targets? So I could just swap the .env.initrd-flash file and flash completely different board using the same image?

If this is the case, this change would break this of course.

madisongh:
I'd rather see this as a build-time (bitbake) variable, or could even be derived from the `OTABOOTDEV` variable we already have in the machine config files. The initrd is built per-MACHINE, which describes a specific combination of module and carrier, so the SPI flash device name is very, very likely to be the same between flashing runs.

recipes-core/initrdscripts/tegra-flash-init_1.0.bb
```diff
@@ -14,6 +14,14 @@ UNPACKDIR = "${S}"
 
 COMPATIBLE_MACHINE = "(tegra)"
 
+OTABOOTDEV ?= "/dev/mtdblock0"
```
madisongh, line 17: This should be `??=` for defaulting.

recipes-core/initrdscripts/tegra-flash-init_1.0.bb
```diff
@@ -14,6 +14,14 @@ UNPACKDIR = "${S}"
...
+
+do_configure() {
+    MTDBLOCK_DEV=$(basename ${OTABOOTDEV})
+    MTD_DEV_NUM="${MTDBLOCK_DEV##mtdblock}"
+    echo "MTD_DEV=mtd${MTD_DEV_NUM}" > ${WORKDIR}/tegra-flash.env
```
madisongh, line 22: Instead of deferring the substition to runtime, I'd suggest just setting MTD_DEV as a bitbake variable (e.g. `MTD_DEV ?= "${@d.getVar('OTABOOTDEV').replace('mtdblock','mtd')}"`),  and using a token like `@MTD_DEV@` in the script itself (in place of the `/dev/${MTD_DEV}` you have now).
 
Then add a `sed -i` command in the `do_install` step to replace the token with the variable's value.

That's the more typical pattern we follow in other recipes for handling values known at build time (and is more typical across other Yocto layers generally).

Then you don't need to install the extra `tegra-flash.env` file.

madisongh (changes requested):

recipes-core/initrdscripts/tegra-flash-init_1.0.bb
loblik: Ok, it's not needed anymore.

madisongh (approved):
Looks good, thanks!

PR #1596, opened 2024-06-25
add support for Jetson AGX Xavier with new BOM according to PCN 208560

PercevalSA:
I can open a pull request per dunfell branch if this one is validated.

add support for PCN 208560 from L4T 32.7.4
Extracted from l4t_generate_soc_bup.sh for BOARDID=2888 and board=jetson-agx-xavier-devkit

SKU
0001 for Jetson AGX Xavier (16 GB RAM)
0004 for Jetson AGX Xavier (32 GB RAM)

FAB
400 for old model with classic BOM
401 and upper for new models with new BOM

### DIFF
--- a/conf/machine/include/tegra194.inc
+++ b/conf/machine/include/tegra194.inc
@@ -31,7 +31,8 @@ TEGRA_BUPGEN_SPECS ?= "fab=400;boardsku=0001;boardrev=H.0 \
                        fab=400;boardsku=0001;boardrev=J.0 \
                        fab=400;boardsku=0001;boardrev=L.0 \
                        fab=400;boardsku=0004;boardrev= \
-                       fab=401;boardsku=0004;boardrev="
+                       fab=401;boardsku=0001;boardrev= \
+                       fab=401;boardsku=0004;boardrev= "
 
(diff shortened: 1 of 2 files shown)


madisongh:
The PCN states that L4T R32.7.2 has support for this updated BOM. I'm not sure that just hacking in the new FAB/SKU combinations to the R32.5.0-based branch is really going to work.

Just upgrading to the `dunfell` branch (which is based off R32.7.4) would be the better approach. A bigger change, but even better one, would be to move to `kirkstone-l4t-r32.7.x`, the current active LTS branch for the R32-series BSP, since `dunfell` is past its end-of-life date.

PercevalSA:
You are perfectly right. I forgot to add the corresponding DTBs from nvidia. This is added to the PR now.

About the dunfell EOL. We are still using dunfell and R32.5.x because right now we cannot update our devices in production without any risk. We are stuck for the moment. We will migrate to kirkstone eventually in one year or two.

I'm eager to have that change in the BSP layer on that branch because that is the one I'm using right now. 

If you do not want to have any modification done to that branch because you consider it dead for now it is fine by me. I will keep those modifications in my own tree :) I just wanted to share what I did.

PR #1610, opened 2024-07-09
Support separate virtual/dtb provider

kekiefer:
Small updates that re-enable the use of a separate virtual/dtb provider for r36. Now we can simply change the `KERNEL_DEVICETREE` variable and update the `PREFERRED_PROVIDER_virtual/dtb`. The actual recipe to build the devicetrees has to pull in the oot-modules sources to build them, so an update to tegra demo distro will be subsequently due.


ichergui:
Hi @kekiefer 
Thanks for this PR. 
Is this an improvement to the current implementation or you are fixing an issue ?

kekiefer:
This is a fix needed to make it possible to define a new provider of virtual/dtb. Without it, the required oot-modules cannot be built since providers of a package are mutually exclusive..

ichergui:
Got it. Thanks

maximeborges:
AFAIK there was no update of the `tegra-demo-distro` using those changes properly.
Could you explain or give an example on the right way of using a custom DTS with `nvidia-kernel-oot-dtb`?

PR #1612, opened 2024-07-10
cuda-gcc.bbclass: Define TOOLCHAIN variable

kraj:
TOOLCHAIN variable is indicative of which compiler toolchain should be used and for cuda-gcc class its GCC obviously, therefore make it clear here, this helps in builds when default toolchain is clang/llvm ( using meta-clang ) and defaulting TOOLCHAIN = "clang"


ichergui (approved):
LGTM. Thanks @kraj

PR #1619, opened 2024-07-13
cuda-samples: Add missing runtime dep

kraj:
Fixes

opkg install cuda-samples
 * Solver encountered 1 problem(s):
 * Problem 1/1:
 *   - conflicting requests
 *   - nothing provides libcublas >= 12.2.5.6-1 needed by cuda-samples-12.2-r0.8.armv8a_tegra234
 *


madisongh (changes requested):
The runtime dependency additions don't look like the right solution to the problem you're seeing.  The S/UNPACKDIR fix looks good, though, thanks.

kraj:
> The runtime dependency additions don't look like the right solution to the problem you're seeing. The S/UNPACKDIR fix looks good, though, thanks.

yeah, thanks for feedback, I am using opkg for package management and ran into these problems, I saw that the related info is there in runtime-deps but somehow its not encoded into final .ipk files. Perhaps a ipk backend specific issue or some such and usually its only happening for lib to lib package deps. I will close this PR and open a new one just for UNPACKDIR for now. I have to narrow down the opkg issue a bit more.

PR #1630, opened 2024-07-18
tegra-libraries-eglcore: Create a symlink instead of hardlink

kraj:
And force the creation, otherwise O_P_M update does not work and fails like below

Configuring tegra-libraries-eglcore.
ln: failed to create hard link '/usr/lib/aarch64-linux-gnu/tegra-egl/libEGL_nvidia.so.0': File exists error: pkg_run_script: package "tegra-libraries-eglcore" postinst script returned status 1.


madisongh:
Unfortunately, making that a soft link will break container support - the container runtime is expecting to see a real file at that location, and will fail to mount it inside the container if it isn't.

We could `rm -f` the link before running `ln`, maybe?

kraj:
> Unfortunately, making that a soft link will break container support - the container runtime is expecting to see a real file at that location, and will fail to mount it inside the container if it isn't.
> 
> We could `rm -f` the link before running `ln`, maybe?

or use `ln -f`, I will update PR

PR #1647, opened 2024-08-01
initrd-flash.sh: fix sed command when using SBK

Lexmark-chad:
- When secure boot is enabled the resulting doflash.sh contains 'uefi_jetson_with_dtb_aligned_blob_w_bin_sigheader_encrypt.bin.signed' vs 'uefi_jetson_with_dtb_sigheader_encrypt.bin.signed' when secure boot is not enabled.  Update the sed command to work with both.

- The command to replace '_aligned_blob_w_bin' with '' must appear before the command to replace 'uefi_jetson' with 'rcmboot_uefi_jetson'.


madisongh:
Thanks! I've also cherry-picked this over to scarthgap-l4t-r35.x.

PR #1658, opened 2024-08-14
[scarthgap-l4t-r35.x] add l4t_version.bbclass

zach-welch-aquabyte:
The global tegra-mender-setup.bbclass from the meta-mender-community layer uses the L4T_VERSION variable to define a `l4t-mender-<version>` machine override, so that variable also needs to be available globally.


classes-recipe/l4t_bsp.bbclass
```diff
@@ -3,7 +3,7 @@ L4T_BSP_NAME ??= "release"
 L4T_SRCS_NAME ??= "sources"
 L4T_BSP_PREFIX ??= "Jetson"
 
-L4T_VERSION ?= "35.5.0"
+INHERIT += "l4t_version"
```
madisongh, line 6: You should just directly `inherit l4t_version` here, I think

madisongh (changes requested):
Directly inheriting the new class should fix the checklayer failure.

classes-recipe/l4t_bsp.bbclass
zach-welch-aquabyte: That was indeed my first instinct here, and this is what I got for the trouble:
```
ERROR: ParseError at .../meta-tegra/classes-recipe/l4t_bsp.bbclass:6: Could not inherit file classes/l4t_version.bbclass
```
It seems that BitBake will not search inside `classes-global` from the `l4t_bsp` recipe class; however the global `tegra-mender-setup` class can `inherit l4t_version` directly.

Moving this to `classes` allows `inherit` to work for both, but it feels... dirty.

madisongh (approved):

PR #1659, opened 2024-08-14
[scarthgap-l4t-r35.x] linux-tegra: properly quote KERNEL_ARGS

zach-welch-aquabyte:
This change permits passing kernel command line arguments that include quoted strings, including dyndbg.  Without single quotes, the recipe silently produces incorrect results in such cases.


madisongh (changes requested):
I'm surprised I've never run into this before. I would have expected bitbake to handle the quoting for us when writing out the shell script, but I guess it doesn't.

There's another instance of this in `classes-recipe/image_types_cboot.bbclass`. If you could add a commit to cover that one, too, this will be good to go.

Thanks!

zach-welch-aquabyte:
I adjusted `image_types_cboot`, squashed that change into the previous commit, and rebased to keep the history linear. Thanks for the prompt turnaround.

madisongh (approved):

PR #1661, opened 2024-08-15
initrd-flash: Add custom extra-pre-wipe

Red-Dragon-99:
This introduces a hook to the initrd-flash script which makes it possible to run custom scripts before erasing the storage disk via "erase-nvme" or overwriting it via "export-devices".

Note: There exists already an "extra" hock to run future scripts before the final reboot step. Similar to that, any future recipe intended to add custom extra pre-wipe scripts can use the TEGRA_INITRD_FLASH_INSTALL variable in the distro config to add the new package(s).

Furthermore, note that these custom scripts will need to be installed with executable permission set in the
/init_extra-pre-wipe.d folder of the tegraflash initrd image.


Red-Dragon-99:
This PR is based on my proposal [here](https://github.com/OE4T/meta-tegra/discussions/1648).

PR #1676, opened 2024-08-27
init-flash.sh: Handle missing UDC initialization

paroque28:
This update adds a check for a missing UDC (USB Device Controller) variable to prevent errors related to uninitialized or absent UDCs, stopping the execution of the boot process if none is found.

A missing UDC can result from various issues, such as an incorrect Device Tree Blob (DTB) configuration or misconfigured power regulators. By exiting early with a clear error message, this change helps diagnose and address the root cause more effectively.


paroque28:
@madisongh sorry for the delay. Done!

PR #1680, opened 2024-08-29
Support for linux-yocto as kernel provider

kekiefer:
Adds support for linux-yocto 6.6 and 6.10 against L4T 36.4.0.

A separate pull request will be opened against scarthgap/36.4.0 to backport support for linux-yocto 6.6 only there.


recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot_36.3.0.bb
akfaro, line 1: I think, this is relatively new for Nvidia: They seem to have git repos for their sources, at least I found them only recently.
Especially, there is this repo for the nvidia OOT modules: https://nv-tegra.nvidia.com/r/gitweb?p=linux-nv-oot.git;a=summary

AFAICT, the blob currently referenced by this recipe's SRC_URI corresponds to the branch [l4t/l4t-r36.3](https://nv-tegra.nvidia.com/r/gitweb?p=linux-nv-oot.git;a=shortlog;h=refs/heads/l4t/l4t-r36.3), resp. tag [jetson_36.3](https://nv-tegra.nvidia.com/r/gitweb?p=linux-nv-oot.git;a=commit;h=564ce2a709cfb300aa2cb092a708452d2ce2c274). Your patches are applied to this revision
The unreleased changes seem to be developed in the branch [rel-36](https://nv-tegra.nvidia.com/r/gitweb?p=linux-nv-oot.git;a=shortlog;h=refs/heads/rel-36), and there are commits that claim improving compatibility with newer Kernel versions.

IMO, it should be considered to reference the git repo itself, or to fork it in the OE4T project if some of the proposed changes are covered by newer commits there.

kekiefer: @akfaro good call-out. It is preferable, at least, to apply the patches from there (where they exist) rather than re-inventing these. I have some concern that the modifications we make to build will make maintaining a _git version of the oot modules difficult, but it seems worth taking that into consideration as well.

madisongh: > I have some concern that the modifications we make to build will make maintaining a _git version of the oot modules difficult

The nv-tegra.nvidia.com site has had reliability problems in the past, too, which has made me nervous about relying on it for any of our recipes.

kekiefer:
Rebased on master, after fixes to egl/gbm, display output works now

recipes-kernel/linux/linux-yocto_6.6.bbappend
```diff
@@ -0,0 +1,117 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
+
+TEGRA_UEFI_SIGNING_CLASS ??= "tegra-uefi-signing"
+
+inherit l4t_bsp ${TEGRA_UEFI_SIGNING_CLASS}
+
+SRC_URI += " \
+    file://0001-memory-tegra-Add-Tegra234-clients-for-RCE-and-VI.patch \
+    file://0002-hwmon-ina3221-Add-support-for-channel-summation-disa.patch \
+    file://0003-cpufreq-tegra194-save-CPU-data-to-avoid-repeated-SMP.patch \
```
ichergui, line 10: Thanks @kekiefer for this PR 
I noticed three patches are specific for tegra194. Could you please confirm that ? 
Could we remove them in the case they are specific for tegra194 because we support only tegra234 family.

recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot_36.3.0.bb
ichergui: We can not rely on the nv-tegra.nvidia.com at the moment. I had the same issues mentioned by @madisongh .

ichergui (changes requested):

akfaro:
Thanks for the proposal, @kekiefer!
I have successfully tested this on a custom Orin NX based board, with the proposed changes applied via `bbappend` files in my board/distro layer.

In order to flash the image via the `initrd-flash` script, I needed a minor change to the init script because the USB-gadget sysfs interface seems to have changed slightly, see attached 
[init-flash-update-UDC-sysfs-interface.patch](https://github.com/user-attachments/files/16836835/init-flash-update-UDC-sysfs-interface.patch)

kekiefer:
> In order to flash the image via the `initrd-flash` script, I needed a minor change to the init script because the USB-gadget sysfs interface seems to have changed slightly, see attached [init-flash-update-UDC-sysfs-interface.patch](https://github.com/user-attachments/files/16836835/init-flash-update-UDC-sysfs-interface.patch)

@akfaro thanks for tracking that one down! I'll rework this a bit so it can function across different kernels.

recipes-kernel/linux/linux-yocto_6.6.bbappend
kekiefer: @ichergui these patches were included as part of the [list nvidia provided](https://docs.nvidia.com/jetson/archives/r36.3/DeveloperGuide/SD/Kernel/BringYourOwnKernel.html#upstream-patches) that were not yet included in the linux-yocto kernel. As these are patches going upstream anyhow, and there are a lot of shared fabric between t194 and t234 SOCs, I think it's safest to include them.

kekiefer:
Today I've updated two things:
* initrd flashing should be working on NX as well as AGX machines
* fixed rtw 8822ce wifi configuration. I tried this again and realized I hadn't committed everything here. Looking into the crash, it's related to enabling ieee80211w (protected management frames) used by WPA3. Using WPA2 wifi will work.

kekiefer:
There is a crash connecting wifi related to enabling ieee80211w (protected management frames) in wpa_supplicant used by WPA3, and has nothing to do with the 8822ce drivers as it occurs with other adapters as well. The problem relates crypto offload to tegra_se (built by the oot modules) as blocking this module from loading allows this function to work.

recipes-kernel/linux/linux-yocto_6.6.bbappend
```diff
@@ -0,0 +1,118 @@
...
+# All of our device trees are out-of-tree
+KERNEL_DEVICETREE:forcevariable = ""
+
+set_scmversion() {
+    if [ "${SCMVERSION}" = "y" -a -d "${S}/.git" ]; then
+        head=$(git --git-dir=${S}/.git rev-parse --verify --short HEAD 2>/dev/null || true)
+        [ -z "$head" ] || echo "+g$head" > ${S}/.scmversion
+    fi
+}
+do_kernel_checkout[postfuncs] += "set_scmversion"
```
akfaro, line 32: I have the impression that this is no longer required:
* Apparently, the file `.scmversion` is no longer used since Kernel v6.3:
https://git.yoctoproject.org/poky/commit/meta/classes/kernel.bbclass?h=kirkstone&id=2b7c113459a602c91badaa7543d02811feac0151
* Also, the variable SCMVERSION is not set by default in `linux-yocto`, unlike `linux-jammy-nvidia-tegra`

kekiefer: Thanks I removed this

kekiefer:
The crash in tegra_se is now resolved by porting a patch from the nv-tegra git repos. I intend to update some of the other oot module patches that do the same as the ones I've created, but the function should be the same. Functionally though, now I think this PR is about where it needs to be.

recipes-kernel/linux/linux-yocto/defconfig
madisongh, line 1: Could this be done without having to provide a full defconfig? Using the in-tree defconfig and providing modifications with config fragments would be more manageable.

recipes-kernel/linux/linux-yocto_6.6.bbappend
madisongh, line 1: This bbappend is going to wreak havoc on other machines in other layers that also use linux-yocto 6.6.

It either needs to be its own recipe, in which case marking it as tegra-only via COMPATIBLE_MACHINE would be fine, or the bbappend needs to be reworked to remove the COMPATIBLE_MACHINE setting and changing all the task/variable modifications to be `:tegra`-specific.

madisongh (changes requested):

recipes-kernel/linux/linux-yocto_6.6.bbappend
kekiefer: Yes of course, no problem

kekiefer: Additional bits for linux-yocto are now in machine-specific overrides. Also, I've moved the tegra-specific fixups to an include file that I only pick up for `tegra` in `MACHINEOVERRIDES` to pick these up for multiple kernel recipes, and keep them out of non-tegra builds. Functionally this should be good, but do you reckon this will happy the layer checkers as well?

recipes-kernel/linux/linux-yocto/defconfig
kekiefer: Phew, well that dread wasn't for no reason. I've isolated what's required now, and there's quite a few things on top of genericarm64, but it should be much more manageable now. As we continue with this, it would make sense to get a tegra branch into yocto kernel meta, but I don't think we're quite there yet.

kekiefer:
I've updated this pr to address a few prior concerns. I still need to go back through the oot patches to pick up nvidia's changes.

What's stopping me now is that master has cut a 6.10 linux-yocto. The kernel incorporates the patches we needed to add for 6.6 (yay) but the oot modules are pretty broken (boo). However, nvidia's downstream git repo has fixes for building with 6.10, so I'm going to spend a little more time and decide if it worth trying to support 6.10 yet, or waiting for the next L4T release with the next stamp for the oot modules that should be compatible with 6.10.

kekiefer:
Well, I got the oot modules built and running with the 6.10 kernel, but it took 30+ patches from nvidia's git repo to make that happen. This is, I think, a few too many to pull into a bsp layer.

I'd like some feedback regarding how to proceed. These are the options:

1. We can get 6.6 done today, going with the current reasonably minimal changes to the kernel and oot modules maintained in-tree. Support for building the 6.10 is left off until we get r36.4 from NVIDIA, at which point the transition to 6.10 should be painless.
2. We add 6.10 support and 30+ patches for the oot modules so we can build against 5.15, 6.6, and 6.10
3. We add 6.10 support, and support a _git build for the oot modules to use with the 6.6 and 6.10 kernels. Based on feedback regarding the lack of stability of the nv-tegra git, we start mirroring the 3 oot modules repos on our github.

madisongh:
That's great, @kekiefer .  We can discuss the options at our next video meeting, assuming you can attend then.

On the whole, though: `scarthgap` core is on the 6.6 kernel, so using that as a starting point wouldn't be too bad, since we could back-port there. Updating to 6.10 could then happen in master, even if we haven't yet got the 36.4 drop from NVIDIA, since the 30+ patches would be relatively short-lived.

With that experience behind us (or while it's going on), we can consider the longer-term approach. Mirroring the relevant NVIDIA OOT repos (I count 7, 5 with drivers and 2 with device trees) could be a bit of a maintenance hassle, too, but maybe less than the hassle of maintaining patches. Either way would be OK, and hopefully the size of the OOT code base will shrink over time.

recipes-kernel/linux/linux-yocto_6.6.bbappend
madisongh: Once you switch this PR out of draft state, we'll find out. :)

recipes-kernel/linux/linux-yocto/defconfig
madisongh: Thanks for doing this. I think we could create our own kernel meta repo in the meantime, if that would help.

kekiefer:
I'm scratching my head a bit regarding what to do about the failing layer checks. I thought they were getting picked up by `find_sccs()` in kernel-yocto.bbclass so I moved them to a tegra-specific override folder (in addition to only being included explicitly for the `tegra` override).

recipes-kernel/linux/linux-yocto_6.10.bbappend
```diff
@@ -0,0 +1,32 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}:"
```
madisongh, line 1: Add the `:tegra` override here and that should take care of the issue being reported by `yocto-check-layer`.

recipes-kernel/linux/linux-yocto_6.6.bbappend
```diff
@@ -0,0 +1,40 @@
+FILESEXTRAPATHS:prepend := "${THISDIR}/${PN}-6.6:${THISDIR}/${PN}:"
```
madisongh, line 1: Add the `:tegra` override here and that should take care of the issue being reported by `yocto-check-layer`.

madisongh (changes requested):
The checklayer failures are due to [this function](https://github.com/openembedded/openembedded-core/blob/1d22562a89e91f19b818b8f9bc214be941303980/meta/classes-recipe/kernel-yocto.bbclass#L100-L111) referenced [here](https://github.com/openembedded/openembedded-core/blob/1d22562a89e91f19b818b8f9bc214be941303980/meta/classes-recipe/kernel-yocto.bbclass#L440). See in-line for a fix for this.

kekiefer:
Thanks Matt, that did the trick

ichergui:
@kekiefer  Could you please rebase your branch to avoid having merge commit ?

Issue #1682, opened 2024-08-30
scarthgap-l4t-r35.x kas build errors


tjhgit:
**Describe the bug**
BUG Nr1
l4t_version.bbclass not found

BUG Nr2
| install: cannot create regular file '/home/yocto/git/ext/meta-mender-community/build/tmp/work/jetson_agx_xavier_devkit-oe4t-linux/tegra-storage-layout-base/35.5.0/image/usr/share/tegraflash/flash_l4t_nvme_rootfs_ab.xml': No such file or directory
| WARNING: exit code 1 from a shell command.
ERROR: Task (/home/yocto/git/ext/meta-mender-community/build/../meta-tegra/recipes-bsp/tegra-binaries/tegra-storage-layout-base_35.5.0.bb:do_install) failed with exit code '1'

**To Reproduce**
```bash
git clone https://github.com/mendersoftware/meta-mender-community.git
git checkout scarthgap
kas build kas/jetson-agx-xavier-devkit.yml
```

**Additional context**

In order to fix the above bugs I think these changes need to be made:

**Bug Nr1**
in `kas/include/tegra-jetpack5.yml`
```
repos:
  meta-tegra:
    # refers to the scarthgap-l4t-r35.x branch
    # commit: 972488415a80afebbd652a983a46c2641eed99ad
    commit: 32d86844d857321e1ed137abbdd5a045c6f5cb3f # NOTE !!! (this commit has the global l4t_version)
```

and **Bug Nr2**
`meta-tegra/recipes-bsp/tegra-binaries/tegra-storage-layout-base_35.5.0.bb`
```
do_install() {
    install -d ${D}${datadir}/l4t-storage-layout
    install -d ${D}${datadir}/tegraflash # NOTE
    install -m 0644 ${PARTITION_FILE} ${D}${datadir}/l4t-storage-layout/${PARTITION_LAYOUT_TEMPLATE}
    install -m 0644 ${PARTITION_FILE} ${D}${datadir}/tegraflash/${PARTITION_LAYOUT_TEMPLATE} # NOTE
    [ -z "${PARTITION_LAYOUT_EXTERNAL}" ] || install -m 0644 ${PARTITION_FILE_EXTERNAL} ${D}${datadir}/tegraflash/${PARTITION_LAYOUT_EXTERNAL} # NOTE
[...]

PR #1684, opened 2024-08-30
at least typo in tegra-storage-layout-base_35.5.0.bb

tjhgit:
bug report [here](https://github.com/OE4T/meta-tegra/issues/1682)

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-storage-layout-base_35.5.0.bb
+++ b/recipes-bsp/tegra-binaries/tegra-storage-layout-base_35.5.0.bb
@@ -10,7 +10,7 @@ PARTITION_FILE_EXTERNAL ?= "${S}/tools/kernel_flash/${PARTITION_LAYOUT_EXTERNAL}
 do_install() {
     install -d ${D}${datadir}/l4t-storage-layout
     install -m 0644 ${PARTITION_FILE} ${D}${datadir}/l4t-storage-layout/${PARTITION_LAYOUT_TEMPLATE}
-    [ -z "${PARTITION_LAYOUT_EXTERNAL}" ] || install -m 0644 ${PARTITION_FILE_EXTERNAL} ${D}${datadir}/l4t-storage-layout/${PARTITION_LAYOUT_EXERNAL}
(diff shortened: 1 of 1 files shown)


dwalkes:
Thanks @tjhgit however I think rather than changing the BSP layer for this mender specific issue it would probably be better to either:
1) Modify the mender layer to use the updated location in scarthgap (preferred) or
2) Put this change in a bbappend within the mender layer.

Note that I think scarthgap support for mender is still a work in progress, you may want to sync up with the PR at https://github.com/mendersoftware/meta-mender-community/pull/402

tjhgit:
ok @dwalkes . But there is still the typo in the meta-tegra code: PARTITION_LAYOUT_EXERNAL . It is missing a T.

dwalkes:
@tjhgit thank you I see that now.  Can you modify the commit to change only this single line?

dwalkes:
You also need to sign off your commit per our [contributor guidelines](https://github.com/OE4T/meta-tegra/blob/master/CONTRIBUTING.md).

tjhgit:
@dwalkes . Regarding changing the rest, I assume this would be here, right: ./meta-mender-tegra/recipes-bsp/tegra-binaries/tegra-storage-layout-base_%.bbappend ?

Sure will modify the commit to only the single line.

dwalkes:
@tjhgit correct, although it looks like there should be handling for this already at https://github.com/mendersoftware/meta-mender-community/blob/scarthgap/meta-mender-tegra/recipes-bsp/tegra-binaries/tegra-storage-layout-base_%25.bbappend#L4

dwalkes:
@tjhgit still missing signoff and you should squash your commits.  I can just propose a PR myself now that I understand the issue if you prefer.

tjhgit:
@dwalkes . Sorry I am not so good at this ... I issued a git commit -s ... So kind to take over. Regarding: https://github.com/mendersoftware/meta-mender-community/blob/scarthgap/meta-mender-tegra/recipes-bsp/tegra-binaries/tegra-storage-layout-base_%25.bbappend#L4 , I think this function is using the wrong location: `mender_flash_layout_adjust()`.

dwalkes:
> So kind that you offer to take over

No problem, thanks for reporting.  I've opened https://github.com/OE4T/meta-tegra/pull/1685

> I think this function is using the wrong location: do_install:append:tegra194()

Makes sense, so yes this is the change you could propose in that layer in a new PR.

PR #1685, opened 2024-08-30
tegra-storage-layout-base: fix typo on external layout

dwalkes:
Fix typo and missing 'T' character in variable reference for `PARTITION_LAYOUT_EXTERNAL`.

The current typo means if the name of the file referenced by `PARTITION_FILE_EXTERNAL` was different than the name `PARTITION_LAYOUT_EXTERNAL`, the wrong filename would be used in the l4t-storage-layout dir.

Relates to  https://github.com/OE4T/meta-tegra/pull/1684


dwalkes:
Should also be backported to `scarthgap-l4t-r35.x` and `scarthgap`, let me know if you'd like PRs for these.

PR #1687, opened 2024-09-02
tegra-uefi-capsules: remove ESP image dependency for BUP

igoropaniuk:
```
Remove ESP image dependency for BUP generation. This helps
to avoid some circular dependencies in such cases as explained in [1].

The helper script (l4t_bup_gen.func) for BUP generation doesn't use it,
and the only mention of ESP partition is in the list of boot artifacts to
be included into BUP image (bup_type=bl),
where ESP image is explicitly marked to be skipped:

5. SKIP: 0 - built in BUP, 1 - not built in BUP
t23x_bl_table=(
        #PART_NAME        IMAGE_SIGNED BOARD_SPECIFIC OP_MODE_SPECIFIC SKIP
        'BCT              1            1              1                0'
        'BCT_A            1            1              1                0'
        'BCT_B            1            1              1                0'
        'BCT-boot-chain_backup 1       1              1                0'
        'A_mb1            1            1              1                0'
        'A_psc_bl1        1            0              1                0'
        'A_MB1_BCT        1            1              0                0'
        'A_MEM_BCT        1            1              0                0'
        'A_tsec-fw        1            0              0                0'
        'A_fsi-fw         1            0              0                0'
        'A_nvdec          1            0              1                0'
        'A_mb2            1            1              0                0'
        'A_bpmp-fw        1            1              1                0'
        'A_bpmp-fw-dtb    1            1              0                0'
        'A_psc-fw         1            0              1                0'
        'A_mts-mce        1            0              1                0'
        'A_sc7            1            0              1                0'
        'A_pscrf          1            0              1                0'
        'A_mb2rf          1            0              0                0'
        'A_cpu-bootloader 1            1              0                0'
        'A_secure-os      1            0              0                0'
        'A_smm-fw         1            0              0                1'
        'A_eks            1            0              0                0'
        'A_dce-fw         1            1              0                0'
        'A_spe-fw         1            0              0                0'
        'A_rce-fw         1            0              0                0'
        'A_adsp-fw        1            0              0                0'
        'uefi_variables   1            0              0                1'
        'uefi_ftw         1            0              0                1'
        'mb2-applet       0            0              0                0'
        'A_VER            0            1              0                0'
        'esp              1            0              0                1'
        'A_xusb-fw        1            0              1                0'
        'secondary_gpt    1            1              0                0'
        'secondary_gpt_backup 1        1              0                0'
)

[1] dc2d7526b1ac("bup: remove extra dependency to ESP")
Signed-off-by: Igor Opaniuk <igor.opaniuk@foundries.io>
```

### DIFF
--- a/recipes-bsp/uefi/tegra-uefi-capsules_36.3.0.bb
+++ b/recipes-bsp/uefi/tegra-uefi-capsules_36.3.0.bb
@@ -26,7 +26,6 @@ do_compile() {
     # BUP generator really wants to use 'boot.img' for the LNX
     # partition contents
     tegraflash_populate_package ${IMAGE_TEGRAFLASH_KERNEL} boot.img ${@tegra_bootcontrol_overlay_list(d, bup=True)}
-    cp "${IMAGE_TEGRAFLASH_ESPIMG}" ./esp.img
     mv generate_bup_payload.sh doflash.sh
     tegraflash_create_flash_config flash.xml.in boot.img ${STAGING_DATADIR}/tegraflash/bupgen-internal-flash.xml
     . ./flashvars
@@ -99,6 +98,5 @@ do_compile[depends] += "virtual/kernel:do_deploy tegra-flashtools-native:do_popu
 do_compile[depends] += "python3-pyyaml-native:do_populate_sysroot"
 do_compile[depends] += "tegra-bootfiles:do_populate_sysroot"
 do_compile[depends] += "coreutils-native:do_populate_sysroot virtual/secure-os:do_deploy"
-do_compile[depends] += "${@'${TEGRA_ESP_IMAGE}:do_image_complete' if d.getVar('TEGRA_ESP_IMAGE') != '' else  ''}"
 do_compile[depends] += "virtual/bootloader:do_deploy"
 do_compile[depends] += "${TEGRA_SIGNING_EXTRA_DEPS} ${DTB_EXTRA_DEPS}"
--- a/classes-recipe/image_types_tegra.bbclass
+++ b/classes-recipe/image_types_tegra.bbclass
@@ -407,6 +407,7 @@ create_tegraflash_pkg[vardepsexclude] += "DATETIME"
 
 def tegraflash_bupgen_strip_cmd(d):
     images = d.getVar('TEGRA_BUPGEN_STRIP_IMG_NAMES').split()
+    images.append("esp.img")
     if len(images) == 0:
         return 'cp flash.xml.in flash-stripped.xml.in'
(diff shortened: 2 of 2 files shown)


ichergui:
Hi @igoropaniuk 
Thanks for the PR. 
Could you please explain the issue that needs this updates ?

igoropaniuk:
Our default image recipe (`lmp-base-console-image`) has both `esp` and `ext4` fstypes in `IMAGE_FSTYPES`, so basically receipe is in charge of creation of both ext4 rootfs and ESP images. This introduces a circular dependency, as our image recipe expects t`egra-uefi-capsules:do_install()` to finish, but `tegra-uefi-capsules:do_compile` expects ESP image to be created. 

We could have followed the same approach like it's already done in meta-tegra, where ESP image is built by a separate image recipe (`recipes-bsp/uefi/tegra-espimage.bb`), but eventually I decided to investigate how the ESP image is used and if it's really included into BUP image

igoropaniuk:
@igoropaniuk @madisongh let me know if there are any additional comments/objections
I'll rebase on the latest master if there is no any and the change is ready to land

recipes-bsp/uefi/tegra-uefi-capsules_36.3.0.bb
```diff
@@ -26,9 +26,10 @@ do_compile() {
     # BUP generator really wants to use 'boot.img' for the LNX
     # partition contents
     tegraflash_populate_package ${IMAGE_TEGRAFLASH_KERNEL} boot.img ${@tegra_bootcontrol_overlay_list(d, bup=True)}
-    cp "${IMAGE_TEGRAFLASH_ESPIMG}" ./esp.img
+    # Generate dummy ESP image, as it's not included into the final BUP image
+    dd if=/dev/zero of=./esp-dummy.img bs=1 count=65536
```
madisongh, line 30: Rather than creating a dummy image, I think it would make more sense to simply remove the reference to the `esp.img` file from the flash layout XML altogether. That should eliminate the need for changing the signature of the function that creates the flash config.

madisongh (changes requested):
Definitely good to eliminate the unneeded dependency, but just removing the filename from the flash layout XML should work just as well. Just making `esp.img` the default value for the TEGRA_BUPGEN_STRIP_IMG_NAMES variable would be enough, although for backward compatibility with users that might be setting that variable themselves, it might be better to make the addition in the `tegraflash_bupgen_strip_cmd` function.

recipes-bsp/uefi/tegra-uefi-capsules_36.3.0.bb
igoropaniuk: @madisongh thanks for suggestion, pushed updated version

madisongh (approved):

madisongh:
Thanks @igoropaniuk !

PR #1699, opened 2024-09-18
arm-trusted-firmware: set DEBUG and LOG_LEVEL from bitbake

atharvanan1:
- this allows ability to modify these in other layers if needed

PR #1703, opened 2024-09-23
arm-trusted-firmware: set DEBUG and LOG_LEVEL from bitbake

atharvanan1:
- this allows ability to modify these in other layers if needed


atharvanan1:
Similar to https://github.com/OE4T/meta-tegra/pull/1699; will apply to Jetpack 5 branches: `kirkstone` and `scarthgap-l4t-r35.x`

PR #1737, opened 2024-10-25
Replace deprecated CVE_CHECK_IGNORE with CVE_STATUS

yumiweidemao:
Per [Yocto manual](https://docs.yoctoproject.org/ref-manual/variables.html#term-CVE_CHECK_IGNORE), CVE_CHECK_IGNORE has been deprecated and should be replaced by CVE_STATUS.

When building with openembedded-core, there is a fallback mechanism in [cve-check.bbclass](https://github.com/openembedded/openembedded-core/blob/master/meta/classes/cve-check.bbclass#L102) but it will print some noisy warning messages.

This pull request is to replace the deprecated CVE_CHECK_IGNORE with CVE_STATUS.


ichergui (approved):
LGTM. Thanks @yumiweidemao

ichergui (changes requested):
@yumiweidemao  Could you please add `Signed-off-by` ? So, we can merge the PR 
Thanks

yumiweidemao:
@ichergui Signoff added. Thanks!

recipes-devtools/gcc-for-nvcc/gcc-for-nvcc-10.3.inc
```diff
@@ -126,4 +126,4 @@ EXTRA_OECONF_PATHS = "\
 "
 
 # Is a binutils 2.26 issue, not gcc
-CVE_CHECK_IGNORE += "CVE-2021-37322"
+CVE_STATUS[CVE-2021-37322] = "ignored"
```
madisongh, line 129: This should specify the actual reason. Looking at the setting in OE-Core's gcc recipe, it should be `"cpe-incorrect: Is a binutils 2.26 issue, not gcc"`.

madisongh (changes requested):

recipes-devtools/gcc-for-nvcc/gcc-for-nvcc-10.3.inc
yumiweidemao: Modified the comment

madisongh: It's the setting that needs changing, and the comment should be removed.  See corresponding line in the [current gcc recipe](https://git.yoctoproject.org/poky/tree/meta/recipes-devtools/gcc/gcc-14.2.inc#n111).

yumiweidemao: Modified the line. Thanks for the clarification!

PR #1745, opened 2024-11-01
Fix device tree file check in nvidia-kernel-oot-dtb

yumiweidemao:
Currently the device tree file check in `nvidia-kernel-oot-dtb` will always return true because it checks if the filepath is non-empty instead of checking if the file exists at that path.

Fixed by replacing `-z` with `! -f` so it will check if the file actually exists. This will ensure the build session correctly exits if the desired device tree file cannot be found.

Tested by building for the off-the-shelf `jetson-agx-orin-devkit` successfully, then changing `KERNEL_DEVICETREE` to some random string that doesn't exist in the kernel source, and the build would exit instead of continuing.


kekiefer:
LGTM

PR #1747, opened 2024-11-04
linux-tegra_5.10.bb: fix sign_bootimg

Lexmark-chad:
- This matches the behavior in classes/image_types_cboot.bbclass whereby the signed kernel image is copied over the top of the original unsigned kernel image.


madisongh:
Thanks @Lexmark-chad ... give me a couple of days, as I think there are other spots that might need similar fixes, and I'd like to have those ready to go in around the same time this gets merged.

PR #1750, opened 2024-11-05
Add Upstream-Status into .patch files even for scarthgap branch

shr-project:
Merging this will allow to use meta-tegra in
those DISTROs which already added patch-status into ERROR_QA
before it was enabled by default in Styhead with:
https://git.openembedded.org/openembedded-core/commit/?h=styhead&id=b7fb91c797ab37a029b8dd1eb7277a7468bc97ed


madisongh:
The checklayer run for this one is also [reporting](https://builder.madison.systems/#/builders/5/builds/29) a patch that needs updating.

shr-project:
> The checklayer run for this one is also [reporting](https://builder.madison.systems/#/builders/5/builds/29) a patch that needs updating.

Sure, done.

I was using patchreview.py, but to detect malformed ones it needs small change I did long time ago (and already forgot about it and it was missing when I was using it yesterday).

PR #1752, opened 2024-11-05
sbsigntool: Fix Upstream-Status format

shr-project:
* drop the one which says 'Inappropriate [Backport]' as it's either Backport or Inappropriate, but cannot be both


madisongh:
Thanks

PR #1756, opened 2024-11-09
tegra-uefi-capsules: add dependency on lz4-native

shr-project:
* HOSTTOOLS is now using lz4 instead of lz4c since: https://git.openembedded.org/openembedded-core/commit/?id=fe167e082cbde1c6d186ecdda531abef610ac2ac and tegra-uefi-capsules do_compile is failing with: tegra-uefi-capsules/35.6.0/recipe-sysroot-native/usr/lib/python3.13/subprocess.py", line 1966, in _execute_child raise child_exception_type(errno_num, err_msg, err_filename) FileNotFoundError: [Errno 2] No such file or directory: 'lz4c' tegra-uefi-capsules/35.6.0/temp/run.do_compile.2229307:369 exit 1 from './doflash.sh'

* I've seen this only with scarthgap-l4t-r35.x branch modified to build for walnascar (master), but it looks like the same is needed in master branch (which has surprisingly to me older version 36.4.0 while scarthgap-l4t-r35.x has 36.6.0, but I don't really use any nvidia MACHINEs, only sometimes building world for them, so maybe it's expected)


madisongh:
Thanks.  The `lz4c` reference in the scripts only applies to Xavier modules, which aren't actually supported in the R36 series BSPs, but it might get used for the supported ones at some point in the future, so it's good to have the explicit dependency anyway.

PR #1757, opened 2024-11-09
Machine config updates and cleanup for proper CHIP_SKU handling

### LINKED ISSUE #1668: Incorrect default chip SKU for Orin Nano targets
This came up in #1655 -

The default chip SKU is being set on a per-board basis. Starting with L4T R35.5.0, the chip SKU is used to select the BPMP firmware to use. The problem is that the P3767 boards, the default chip SKU is D3, which selects the TE980M firmware, but the Orin Nano needs the TE950M firmware.

When flashing this isn't a big deal, since the actual chip SKU gets read from the target. But when signing during the build, only defaults are used, so we end up with the wrong pre-signed firmware being loaded onto the target.

We should probably address this by adding a `TEGRA_CHIP_SKU` variable that can be set in the machine configuration file as a default to be passed to the helper scripts.

madisongh:
* Fix the TEGRA_FLASHVARS setting to remove obsolete entries and adds missing ones
* Add TEGRA_FLASHVAR_CHIP_SKU settings for all machines
* Update flash helper script to use the default and complain when there isn't one

Additional cleanup:
* Refactor the Orin NX common config to an include file, to match Orin Nano and AGX Orin
* Use the common AGX Orin include for JAOI
* Add cross-check of board ID and board SKU in the helper script to help prevent mis-flashing

Fixes #1668  along with other improvements


ichergui:
Thanks @madisongh for this. LGTM

ichergui (approved):
LGTM. Thanks @madisongh

PR #1760, opened 2024-11-12
libdrm: move dependency on nvidia-drm out of weston-init

kekiefer:
The nvidia-drm kernel module and associated load config are needed by libdrm but were only getting picked up by weston-init.

There are valid configurations (headless, egl-only) that need libdrm but do not pull in weston, so the runtime recommendation is moved to the libdrm bbappend.

### DIFF
--- a/recipes-graphics/drm/libdrm_%.bbappend
+++ b/recipes-graphics/drm/libdrm_%.bbappend
@@ -1,1 +1,3 @@
+PACKAGE_ARCH:tegra = "${TEGRA_PKGARCH}"
+
 RRECOMMENDS:${PN}:tegra = "kernel-module-nvidia-drm nvidia-drm-loadconf"
--- a/recipes-kernel/nvidia-drm-loadconf/nvidia-drm-loadconf_1.0.bb
+++ b/recipes-kernel/nvidia-drm-loadconf/nvidia-drm-loadconf_1.0.bb
@@ -3,7 +3,7 @@ LICENSE = "MIT"
 LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/MIT;md5=0835ade698e0bcf8506ecda2f7b4f302"
 
 SRC_URI = "\
-    file://nvidia-drm-modprobe.conf \
+    file://nvidia-drm-modprobe.conf.in \
 "
 
 S = "${WORKDIR}/sources"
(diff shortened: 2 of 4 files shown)


madisongh:
There used to be a conflict between enabling modesetting in the nvidia-drm driver and the `nvidia_drv` X server driver module, which is why we had it tied in with weston, despite that not being the true nature of the dependency. Is that conflict still there? If so, we'll need to come up with a different scheme here.

kekiefer:
That makes sense - I didn’t check with the X server, so I’ll follow up here.

kekiefer:
Interestingly, it is fine with this option in the linux-yocto kernel (doesn't matter one way or another there), but with jammy it's old problem with the `modeset=1` parameter.

I'll change this to adjust that parameter if x11 is enabled and not xwayland.

kekiefer:
I've updated this to RCONFLICT a `${PN}-modeset` package against whatever is stated to be `${XSERVER}` if the modeset parameter is 1. The main `nvidia-drm-loadconf` package will always be present and load the module, but only RRECOMMENDs the `${PN}-modeset` part.

recipes-graphics/drm/libdrm_%.bbappend
madisongh, line 1: The weston-init recipe is already marked as machine-specific, but libdrm is not. Adding
```
```
as part of this change will help prevent weird build failures when mixing tegra and non-tegra builds.

recipes-kernel/nvidia-drm-loadconf/nvidia-drm-loadconf_1.0.bb
```diff
@@ -3,16 +3,33 @@ LICENSE = "MIT"
...
+    install -m 0644 ${B}/nvidia-drm-modprobe.conf ${D}${sysconfdir}/modprobe.d/nvidia-drm.conf
     echo "nvidia-drm" > ${D}${sysconfdir}/modules-load.d/nvidia-drm.conf
 }
+
+PACKAGES =+ "${PN}-modeset"
+FILES:${PN}-modeset = "${sysconfdir}/modprobe.d"
+RRECOMMENDS:${PN} = "${PN}-modeset"
+
+# if modesetting was enabled at build time, don't include the modprobe conf at runtime alongside the XSERVER packages
+RCONFLICTS:${PN}-modeset = "${@d.getVar('XSERVER') if bb.utils.to_boolean(d.getVar('NVIDIA_DRM_MODESET')) else ''}"
```
madisongh, line 35: Instead of all of the XSERVER variable contents, you could be more specific here and just name `xserver-xorg-video-nvidia`, which where we have the conflict.... but I'm not sure it makes that much difference.

madisongh (changes requested):
Thanks... mostly looks good, see in-line for a fix and a suggestion.

kekiefer:
Sorry for the delay, I just made the noted changes

PR #1761, opened 2024-11-12
nvidia-kernel-oot: fix a wrong variable

liuming50:
Commit 9fa8701606dffc7153e7593f817e8ef8f56d335e:
[ nvidia-kernel-oot: fix a build issue ]

intended to pass ${STAGING_KERNEL_BUILDDIR} to make but mistakenly passed ${STAGING_KERNEL_DIR} which still referring to kernel source but not kernel builddir, fix that.


liuming50:
Hi, @madisongh, my previous MR has a wrong reference to a variable, it did not really fix the problem, fixed in this MR.

PR #1765, opened 2024-11-19
external/openembedded-layer: update opencv bbappend version to 4.10

shr-project:
opencv upgrade to 4.10 was just backported and merged into styhead branch of meta-oe:

https://git.openembedded.org/meta-openembedded/commit/?h=styhead&id=c23165bb42ab4beebec336c15fd3db8df9eea751


madisongh:
I don't think a simple rename is going to work here, since there were patches I had to back-port from 4.10 onto 4.9 that should no longer apply.

shr-project:
I'm sorry I didn't get that far as to build opencv, I've backported just your first change to unblock parsing with updated meta-oe/styhead.

I guess reverting:
945bfbfd external/openembedded-layer: opencv: retrofit for opencv 4.9
would fix the rest (to make it identical with master).

madisongh:
Rather than reverting the earlier patch, I've manually updated the bbappend to just drop the now-obsolete patches, to make the history a little easier to follow. The branch should be compatible with meta-oe styhead latest now.

PR #1768, opened 2024-11-20
Scarthgap linux yocto 6.6

kekiefer:
Backport of changes from master to support linux-yocto on scarthgap.

Single functional change to include r8169 for 6.6. This should be done for master as well.


ichergui (approved):
LGTM. Thanks @kekiefer

PR #1776, opened 2024-11-28
Add SECURITY.md file

madisongh:
now required by yocto-checklayer.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,13 @@
+# Reporting vulnerabilities
+
+Security vulnerabilities, or potential vulnerabilities, in NVIDIA
+packages should be reported to NVIDIA directly through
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,13 @@
+# Reporting vulnerabilities
+
+Security vulnerabilities, or potential vulnerabilities, in NVIDIA
+packages should be reported to NVIDIA directly through
+[the NVIDIA Product Security](https://www.nvidia.com/en-us/security/) page.
```
ichergui, line 5: Maybe we should add the Nvidia forum `https://forums.developer.nvidia.com/` ?

madisongh: For security related stuff, probably better to just point people at the official page for those kinds of issues.

PR #1777, opened 2024-11-28
Update to L4T R32.7.6/JetPack 4.6.6

madisongh:
With this we can sound the death knell for JetPack 4.

Note that the `tegra-l4t-r32.7.6`  tags in the git repos on nv-tegra.nvidia.com are messed up, pointing to positively ancient branches. However, the sources in the public_sources tarball, which we use in the layer for most things, appear to be correct.

For the kernel, I derived patches (of which there were just two very small ones) by comparing the kernel sources in the tarball against our L4T R32.7.5 kernel tree.


ichergui (approved):
LGTM. Thanks @madisongh

PR #1781, opened 2024-12-06
tegra-libraries-nvdsseimeta: move gstreamer package to SRC_COMMON_DEBS

emiljt:
In the deb package repository for 36.4.0, the gstreamer package was moved back to the common pool.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-libraries-nvdsseimeta_36.4.0.bb
+++ b/recipes-bsp/tegra-binaries/tegra-libraries-nvdsseimeta_36.4.0.bb
@@ -3,11 +3,14 @@ L4T_DEB_COPYRIGHT_MD5 = "20bbd485b9b57fbc0b55d6efc08e3f4a"
 
 DEPENDS = "glib-2.0 gstreamer1.0"
 
-L4T_DEB_TRANSLATED_BPN = "nvidia-l4t-gstreamer"
-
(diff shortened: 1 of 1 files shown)


madisongh:
Thanks for this. I'm not sure how the feed change got missed when we did the last BSP update.

 It turns out we have another recipe that is also affected by this change, so I've reworked the recipes a bit to make it easier to specify that the deb package comes from the common feed, rather than the per-SoC one. And I'll try and keep a closer watch on future BSP updates to catch these changes.

eg9152:
@madisongh:  I think the reason it was missed is that this package is downloaded in two different recipes--one place it was downloaded from the correct location, and in another it was the wrong location.  So there was a race for whether you would see an error.  It depended on which recipe downloaded it first.

* Correctly downloaded it: `recipes-bsp/tegra-binaries/tegra-libraries-camera_36.4.0.bb:  SRC_COMMON_DEBS += "${@l4t_deb_pkgname(d, 'gstreamer')};subdir=${BP};name=gstreamer"`
* Failed, unless it was already in the DL cache: `recipes-bsp/tegra-binaries/tegra-libraries-nvdsseimeta_36.4.0.bb:  L4T_DEB_TRANSLATED_BPN = "nvidia-l4t-gstreamer"`

Maybe we need to make how recipes reference these packages more consistent so it's easier to catch things like this.

PR #1785, opened 2024-12-10
linux-yocto: add patch to fix compatibility with nvbootctrl

Austriker:
Solves the issue with `nvbootctrl`. 

```
# /usr/sbin/nvbootctrl dump-slots-info
Current version: 36.4.0
Capsule update status: 1
Current bootloader slot: B
Active bootloader slot: B
num_slots: 2
Error: memory map failed
# /usr/sbin/nvbootctrl verify
Info: variable BootChainFwStatus is not found.
Error: memory map failed
Error clear SR value node_name[0]
```

After adding the Nvidia SAUCE:
```
# /usr/sbin/nvbootctrl dump-slots-info
Current version: 36.4.0
Capsule update status: 0
Current bootloader slot: A
Active bootloader slot: A
num_slots: 2
slot: 0,             status: normal
slot: 1,             status: normal
```

Tested with Mender updates

### DIFF
--- a/recipes-kernel/linux/linux-yocto-6.10/0001-NVIDIA-SAUCE-soc-tegra-pmc-Add-sysfs-nodes-to-select.patch
+++ b/recipes-kernel/linux/linux-yocto-6.10/0001-NVIDIA-SAUCE-soc-tegra-pmc-Add-sysfs-nodes-to-select.patch
@@ -1,0 +1,285 @@
+From 36a9d28b3ee946e46d166985ebce16684905f3a5 Mon Sep 17 00:00:00 2001
+From: Petlozu Pravareshwar <petlozup@nvidia.com>
+Date: Fri, 8 Mar 2024 03:18:17 +0000
+Subject: [PATCH] NVIDIA: SAUCE: soc/tegra: pmc: Add sysfs nodes to select boot
+ chain
+
+Add sysfs nodes to select A/B boot chain so that BootROM can select
+the right boot path on warm boot.
+
+Bug 4510385
+
+Upstream-Status: Backport [5.15.148-1012.12]
+
+Signed-off-by: Petlozu Pravareshwar <petlozup@nvidia.com>
+Change-Id: Ife973c5310d7b640388bcc70e9c62c7351f80ce0
+Reviewed-on: https://git-master.nvidia.com/r/c/3rdparty/canonical/linux-jammy/+/3136031
+(cherry picked from commit 0cacf9c231c31a4e622fa64ea8e51cd1c5d5f7a8)
+Reviewed-on: https://git-master.nvidia.com/r/c/3rdparty/canonical/linux-jammy/+/3139547
+Reviewed-by: Bitan Biswas <bbiswas@nvidia.com>
+Reviewed-by: Russell Xiao <russellx@nvidia.com>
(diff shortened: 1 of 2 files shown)


recipes-kernel/linux/linux-yocto-6.10/0001-NVIDIA-SAUCE-soc-tegra-pmc-Add-sysfs-nodes-to-select.patch
```diff
@@ -0,0 +1,283 @@
...
+From: Petlozu Pravareshwar <petlozup@nvidia.com>
+Date: Fri, 8 Mar 2024 03:18:17 +0000
+Subject: [PATCH] NVIDIA: SAUCE: soc/tegra: pmc: Add sysfs nodes to select boot
+ chain
+
+Add sysfs nodes to select A/B boot chain so that BootROM can select
+the right boot path on warm boot.
+
+Bug 4510385
+
```
madisongh, line 11: The yocto-checklayer script doesn't flag an error for this yet, but an `Upstream-Status:`  line is required for all patches.

```
Upstream-Status: Backport [5.15.148-1012.12]
```
inserted here would be the most appropriate. This is more of a forward port than a backport, but "Backport" is the only status allowed by the Yocto tooling that's close enough to describe the situation.
See [here](https://docs.yoctoproject.org/singleindex.html#patch-upstream-status) for more info, and note that the tools are annoyingly case-sensitive in their checks.

madisongh (changes requested):
Glad this worked, thanks for submitting.  See in-line for a needed change.

I'd also appreciate it if the commit message followed the usual format here, starting with the recipe that's being changed - something like `linux-yocto: add patch to fix compatibility with nvbootctrl`

recipes-kernel/linux/linux-yocto-6.10/0001-NVIDIA-SAUCE-soc-tegra-pmc-Add-sysfs-nodes-to-select.patch
Austriker: Understood ! Thank you for the feedback I have made the changes.

Austriker:
> Glad this worked, thanks for submitting. See in-line for a needed change.
> 
> I'd also appreciate it if the commit message followed the usual format here, starting with the recipe that's being changed - something like `linux-yocto: add patch to fix compatibility with nvbootctrl`

I have updated the commit message.

akfaro:
@Austriker @madisongh
Do I see it right that this might also be an issue on linux-yocto 6.6 which is currently the default version on `scarthgap`?

The symptom I see now on scarthgap/6.6 seems to be the same as documented in the PR description ("Error: memory map failed").

I wonder though, why my own integration of linux-yocto 6.6 did not show the symptom before I updated `meta-tegra` to the PR https://github.com/OE4T/meta-tegra/pull/1768. My work was based on an early stage of the PR https://github.com/OE4T/meta-tegra/pull/1680, and also did not include the patch suggested here...

Austriker:
Hi @akfaro it might be the case I am personally running 6.10 on the scarthgap branch without any issues.
I just pulled the changes from the master branch.

Issue #1788, opened 2024-12-13
tegra-flash-helper.sh seems to not properly handle flashing in secure boot mode when only PKC is used, without SBK


bradfa:
**Describe the bug**

When attempting to flash an Orin NX which has PKC hash fused and enabled within the BootSecurityInfo fuses but SBK security is not enabled in the BootSecurityInfo fuses (regardless of if SBK is fused or not) then when flashing, the tegra-flash-helper.sh script will complain that it needs both `-u` and `-v` switches passed to it to define both the PKC key and SBK key.

This should not actually be true, as the SBK key is not required for flashing in this case as SBK is not enabled within BootSecurityInfo fuse block.

**To Reproduce**

Using meta-tegra scarthgap branch (although master appears to have the same issue) when flashing Orin NX:

```
$ sudo ./initrd-flash -u /home/andrew/downloads/pkc.key
Starting at 2024-12-13T10:21:10-05:00
Machine:       <redacted>
Rootfs device: nvme0n1p1
Found Jetson device in recovery mode at USB 3-1
== Step 1: Signing binaries at 2024-12-13T10:21:10-05:00 ==
ERR: Target is configured for secure boot (PKC); use -u and -v options to specify key files
ERR: signing failed at 2024-12-13T10:21:10-05:00
```

**Additional context**

This appears to be just a logic issue with https://github.com/OE4T/meta-tegra/blob/master/recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra-flash-helper.sh#L262

This line should be broken out to cover just PKC mode versus the combined PKCSBK mode as the key requirements for each are different.  PKC mode should only require the PKC key while SBKPKC mode requires both the PKC and SBK keys to be available.

PR #1789, opened 2024-12-16
tegra-flash-helper: Allow PKC-only secure boot flashing

bradfa:
When flashing a board fused with PKC hash but no SBK, we should not supply a SBK key as that would encrypt the flash contents which would then not be able to be read by the board when booting, as the board won't know the SBK key.

This also ideally should be backported to at least the scarthgap and styhead branches.

Fixes: https://github.com/OE4T/meta-tegra/issues/1788


bradfa:
The autobuilder failure appears to be related to linux-yocto_6.10 recipe issues and not related to my change, if I'm understanding its output correctly.

ichergui (approved):
LGTM. Thanks @bradfa

Issue #1784, opened 2024-12-09
cuda-samples is not buildable with latest master with cuda 12.6


kraj:
**Describe the bug**
cuda has been upgraded on master to 12.6 but the samples repo is still pinned at 12.5 release
This maybe the reason for non-buildable cuda-samples recipe

```
| NOTE: make -j 44 GENCODE_FLAGS=--gpu-architecture=compute_87 --gpu-code=sm_87 SMS=87 OPENMP=yes CUDA_PATH=/mnt/b/yoe/master/build/tmp/work/armv8a_tegra234-yoe-linux/cuda-samples/12.5/recipe-sysroot/
/usr/local/cuda-12.6 HOST_COMPILER=aarch64-yoe-linux-gcc  -march=armv8-a+crc -mbranch-protection=standard -fstack-protector-strong  -O2 -D_FORTIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-se
curity --sysroot=/mnt/b/yoe/master/build/tmp/work/armv8a_tegra234-yoe-linux/cuda-samples/12.5/recipe-sysroot CCFLAGS=-march=armv8-a+crc -mbranch-protection=standard -fstack-protector-strong -O2 -D_FOR
TIFY_SOURCE=2 -Wformat -Wformat-security -Werror=format-security --sysroot=/mnt/b/yoe/master/build/tmp/work/armv8a_tegra234-yoe-linux/cuda-samples/12.5/recipe-sysroot  -O2 -g   -fmacro-prefix-map=/mnt
/b/yoe/master/build/tmp/work/armv8a_tegra234-yoe-linux/cuda-samples/12.5/git=/usr/src/debug/cuda-samples/12.5  -fdebug-prefix-map=/mnt/b/yoe/master/build/tmp/work/armv8a_tegra234-yoe-linux/cuda-sample
s/12.5/git=/usr/src/debug/cuda-samples/12.5  -fmacro-prefix-map=/mnt/b/yoe/master/build/tmp/work/armv8a_tegra234-yoe-linux/cuda-samples/12.5/git=/usr/src/debug/cuda-samples/12.5  -fdebug-prefix-map=/m
[...]

PR #1790, opened 2024-12-16
cuda-compatibility-workarounds: Fix build with upcoming glibc 2.41

kraj:
sinpi/cospi/sinpif/cospif are now defined in glibc as part of C23 extentions but they conflict with CUDA declrations, therefore use the glibc definitions when its provided by glibc

Fixes https://github.com/OE4T/meta-tegra/issues/1784


madisongh:
Thanks @kraj .  If you could rebase, the checklayer run should pass, and I'll merge this.

PR #1792, opened 2024-12-19
tegra-flash-helper: Allow PKC-only secure boot flashing

bradfa:
Commit c523c2f93873d8d255c2e84ec390ccb4d3701c40 on master.

When flashing a board fused with PKC hash but no SBK, we should not supply a SBK key as that would encrypt the flash contents which would then not be able to be read by the board when booting, as the board won't know the SBK key.


madisongh:
I've gone ahead and cherry-picked the commit from master into both styhead and scarthgap. Sorry for not taking care of this sooner.

PR #1796, opened 2025-01-07
Fix DEVSUM x86-64 for cuda-cupti

minicarpet:
nativesdk-cuda-cupti had wrong DEVSUM
This branch should also be merged to master


madisongh:
Thanks! I'll take care of the cherry-pick to other branches.

PR #1807, opened 2025-01-16
tegra-espimage: allow custom filesystem size

loblik:


loblik:
Present comment in code says ESP must be 64M. Is it really true?

Can we default to this value but still allow to change it? Seems to work fine on Orin at least.

madisongh:
It's been so long that I don't remember why I was so fixated on 64MB for the size... the only real limitation I see is that the storage layouts that come from NVIDIA have the ESP size set to that value. If you're using a custom layout, you could set a different size.

Feel free to remove the comment as part of this PR.

PR #1808, opened 2025-01-16
optee-os: make build work in devtool

loblik:
Make sure we have proper path to MM even when building from devtool workspace.


ichergui (approved):
LGTM. Thanks @loblik

PR #1810, opened 2025-01-16
edk2-firmware-tegra: fix agx xavier peristent vars

dwalkes:
As discussed in [1], [2], and [3], AGX Xavier targets are no longer able to write any persistent variables from edk boot, and are no longer to perform capsule updates after 355 reboots.

Pull in the patch in [4] which is confirmed to resolve this issue as tested on AGX Xavier with 500 reboots using the script at [5] plus a capsule update test.

1: https://forums.developer.nvidia.com/t/possible-uefi-memory-leak-and-partition-full/308540/43
2: https://forums.developer.nvidia.com/t/redundant-a-b-rootfs-not-switching-with-set-active-boot-slot-but-working-with-set-sr-br/306475/27
3: https://github.com/NVIDIA/edk2-nvidia/issues/114
4: https://github.com/NVIDIA/edk2-nvidia/commit/c101ba515b2737fb78d8929c2852f5c8f9607330
5: https://github.com/Trellis-Logic/linux-test-scripts/blob/main/tegra_uefi_reboot_test.py


dwalkes:
kirkstone version of https://github.com/OE4T/meta-tegra/pull/1804

PR #1818, opened 2025-01-21
Support zstd compressed tegraflash images

AustinSchuh:
zstd is a lot faster when compressing and decompressing, which helps with iteration speed.


ichergui (approved):
LGTM
Thanks @AustinSchuh

madisongh (commented):
Hmm.  Rather than building the additional compression method in here, it would probably be better to remove the direct `gzip` use, just generate an uncompressed tarball, and then let people specify `tegraflash.XXXX` where XXXX is the suffix for their preferred compression method.

And while we're at it, we could probably drop direct creation of ZIP packages here, too... particularly since ZIP doesn't play well with the EXT4 images that usually end up in the package.

AustinSchuh:
Sounds interesting.  Mind being a bit more specific about how that would work?  I'm not great with bitbake, and want to make sure I understand what you are thinking.  Something like just passing the extension to tar, and letting it pick what to do?

madisongh:
> Something like just passing the extension to tar, and letting it pick what to do?

Sort of.  The output for the `tegraflash` IMAGE_TYPE would be a plain `.tar` - not compressed at all. A user can then choose to set IMAGE_FSTYPES to be `"tegraflash.zst"`, and the CONVERSIONTYPES machinery in `image_types.bbclass` should pass the tarball through zst for them.

AustinSchuh:
OK, that worked.  More lines deleted than added!

madisongh:
There was one minor fix needed, which I've taken care of, and I've merged this manually.  Thanks, @AustinSchuh !

AustinSchuh:
Thanks!  I tested on scarthgap, master must have been a bit different.

PR #1830, opened 2025-02-05
opencv: Adjust bbappend for 4.11 upgrade from meta-oe

kraj:
Drop backport patch present in 4.11 already


ichergui (approved):
Thanks @kraj

PR #1836, opened 2025-02-12
linux-yocto-6.6: Delete bbappend

kraj:
6.6 recipes are gone from core layer


kraj:
see https://git.yoctoproject.org/poky/commit/?id=5346f1fecbf9d90a98eae9f99b57ad9491a862b5

PR #1861, opened 2025-03-14
tegra-uefi-capsules: fix signing with external server

Lexmark-chad:
- 'mv generate_bup_payload.sh doflash.sh' should be conditional on TEGRA_SIGNING_EXCLUDE_TOOLS.  Those who use external signing servers will set this to "1" which skips the generation of generate_bup_payload.sh.  See invocations of tegraflash_generate_bupgen_script in image_types_tegra.bbclass.

### DIFF
--- a/recipes-bsp/uefi/tegra-uefi-capsules_36.4.0.bb
+++ b/recipes-bsp/uefi/tegra-uefi-capsules_36.4.0.bb
@@ -26,7 +26,9 @@ do_compile() {
     # BUP generator really wants to use 'boot.img' for the LNX
     # partition contents
     tegraflash_populate_package ${IMAGE_TEGRAFLASH_KERNEL} boot.img ${@tegra_bootcontrol_overlay_list(d, bup=True)}
-    mv generate_bup_payload.sh doflash.sh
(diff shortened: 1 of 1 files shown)


ichergui (approved):
LGTM. Thanks @Lexmark-chad

madisongh (approved):
I think I'm understanding the need for this... it looks like we could simplify this a bit by moving some of the BUP/capsule-specific logic out of the bbclass and into the recipe at some point, so there's less ping-ponging between the two files when trying to sort out what should be happening.  That logic was in the bbclass due to BUPs being their own thing, getting generated from multiple places in multiple recipes. That's not the case these days.

Lexmark-chad:
> it looks like we could simplify this a bit by moving some of the BUP/capsule-specific logic out of the bbclass and into the recipe at some point, so there's less ping-ponging between the two files when trying to sort out what should be happening.

I'll circle back and take a look at that.  Will be next week at the earliest.

madisongh:
Thanks @Lexmark-chad ...  no big rush, just wanted to note this for the next time we make changes in this area

Issue #638, opened 2021-03-22
dunfell-l4t-r32.5.0: nv_update_engine doesn't work if TNSPEC is too long


nielsavonds:
In our setup, the TNSPEC was 64 bytes long. Apparently, nv_update_engine is unable to cope with a TNSPEC longer than 63 bytes, probably due to some buffer overflow. This translates in failed updates showing the following log:

```
Device TN Spec: 3668-300-0000-B.0-1-2-jetson-xavier-nx-nobi-revision-a-mmcblk0p1
Can't find matching TN Spec in OTA Blob!
```

**To Reproduce**
1. Create a MACHINE named jetson-xavier-nx-nobi-revision-a (or anything with similar or longer length)
2. Create a build based on the meta-tegra dunfell-l4t-r32.5.0 branch
3. Try to run nv_update_engine -i --no-reboot
4. See error "Can't find matching TN SPEC in OTA Blob!"

**Suggestion**
I assume this is something out of your control, so I would suggest including a warning or error when building using a machine name that is too long, to avoid anyone else spending a lot of time trying to figure out what is going on. I would either create a separate function for this or put it in tegraflash_generate_bupgen_script. Something along the lines of:

```
    tnspeclen=$(expr length ${TNSPEC})
    if [ $tnspeclen -ge 64 ];
    then
        bbfatal "TNSPEC too long: ${TNSPEC} ($tnspeclen characters - max 63 characters)"
    fi
```

ichergui:
Hey @nielsavonds 
Thanks for sharing this.
I think we can add this check in the build time for safety, however, we are using the same/similar machine names as Nvidia have in Jetpack.

madisongh:
We could issue a warning on any MACHINE name longer than 31 characters, since the rest of the fields+separators shouldn't be more than 32 characters; an earlier warning would be better, I think.  But also including a check on the total length of the spec during BUP generation is also a good idea, just in case.

madisongh:
I've also added a note about machine name length to the [wiki page about creating a custom machine definition](https://github.com/OE4T/meta-tegra/wiki/Creating-a-custom-MACHINE).

PR #1865, opened 2025-03-23
setup-nv-boot-control: add COMPAT_TARGET

dwalkes:
Found when testing https://github.com/OE4T/tegra-demo-distro/pull/320 with the test procedure there on AGX Xavier devices.  I believe it's necessary for this commit or something like it on any Jetpack 5 AGX Xavier based targets to be applied and implemented before slot switching through nvbootctrl to work properly.

I considered defining `TNSPEC_COMPAT_MACHINE` in the include file for agx-xavier but not sure we want to make that change, especially on a stable branch, at this point, because I'm not sure what side effect changing the compatible spec will have for already deployed devices.

Also based on the discussion at https://forums.developer.nvidia.com/t/nvbootctrl-hardcoded-to-only-handle-agx-xavier/270411/4 I suspect this probably only applies to AGX Xavier.

I think I may just make a note about this in https://github.com/OE4T/meta-tegra/wiki/Creating-a-custom-MACHINE.

I've only tested on kirkstone thus far but if this is the right change it should probably go to `scarthgap-l4t-r35.x` as well.

# Commit Comment
From discussion at [1], NVIDIA hard-codes a check for COMPATIBLE_SPEC into nvbootctrl in order to determine whether to use EFI variables in emmc.

For custom MACHINEs which use a custom TARGET that doesn't match the `unify_name` value in the
`nv-l4t-bootloader-config.sh` script, this means any attempt to set bootloader slot from nvbootctrl, for instance the attempt to set slots as an alternative to capsule update with [2], fail with
```
Error: open fail errno = 30 reason = Read-only file system
Error: write variable BootChainFwNext failed!
```

Add a COMPAT_TARGET to the setup-nv-boot-control script which by default is set to TARGET.  However, if using a custom MACHINE/TARGET which is based on
`jetson_agx_xavier_devkit` you can define:
```
TNSPEC_COMPAT_MACHINE = "jetson-agx-xavier-devkit"
```
in your machine configuration and nvbootctrl will
work as expected, where `jetson-agx-xavier-devkit` is the value of `unify_name` in `nv-l4t-bootloader-config.sh` in the Jetpack release sources from nvidia for
`jetson_agx_xavier_devkit` target.

1: https://forums.developer.nvidia.com/t/nvbootctrl-hardcoded-to-only-handle-agx-xavier/270411/4
2: https://github.com/OE4T/tegra-demo-distro/pull/320

### DIFF
--- a/recipes-bsp/tools/setup-nv-boot-control/setup-nv-boot-control.sh.in
+++ b/recipes-bsp/tools/setup-nv-boot-control/setup-nv-boot-control.sh.in
@@ -1,5 +1,6 @@
 #!/bin/sh
 TARGET="@TARGET@"
+COMPAT_TARGET="@COMPAT_TARGET@"
 BOOTDEV="@BOOTDEV@"
 sysconfdir="@sysconfdir@"
 controlfile="${sysconfdir}/nv_boot_control.conf"
@@ -24,8 +25,8 @@ compatspec=$(echo "$boardspec" | gen_compat_spec)
 if [ -z "$compatspec" ]; then
     compatsed="-e/^COMPATIBLE_SPEC/d"
 else
-    compatsed="-es,@COMPATIBLE_SPEC@,${compatspec}-${TARGET}-,"
-    set_efi_var "TegraPlatformCompatSpec" "781e084c-a330-417c-b678-38e696380cb9" "${compatspec}-${TARGET}-" "write-once" || rc=1
+    compatsed="-es,@COMPATIBLE_SPEC@,${compatspec}-${COMPAT_TARGET}-,"
+    set_efi_var "TegraPlatformCompatSpec" "781e084c-a330-417c-b678-38e696380cb9" "${compatspec}-${COMPAT_TARGET}-" "write-once" || rc=1
 fi
 sed -e"s,@TNSPEC@,${boardspec}-${TARGET}-${BOOTDEV}," $compatsed \
     "${sysconfdir}/nv_boot_control.template" > "$controlfile" || rc=1
--- a/recipes-bsp/tools/setup-nv-boot-control_1.0.bb
+++ b/recipes-bsp/tools/setup-nv-boot-control_1.0.bb
@@ -18,6 +18,7 @@ ESPMOUNT ?= "/boot/efi"
 NVIDIA_ESPMOUNT ?= "/opt/nvidia/esp"
 ESPMOUNTUNIT ?= "${@'-'.join(d.getVar('ESPMOUNT').split('/')[1:])}.mount"
 ESPVARDIR ?= "${ESPMOUNT}/EFI/NVDA/Variables"
+TNSPEC_COMPAT_MACHINE ??= "${TNSPEC_MACHINE}"
 
 S = "${WORKDIR}"
 B = "${WORKDIR}/build"
@@ -26,6 +27,7 @@ inherit systemd update-rc.d
 
 do_compile() {
     sed -e's,@TARGET@,${TNSPEC_MACHINE},g' \
+        -e's,@COMPAT_TARGET@,${TNSPEC_COMPAT_MACHINE},g' \
         -e's,@BOOTDEV@,${TNSPEC_BOOTDEV},g' \
         -e's,@sysconfdir@,${sysconfdir},g' \
         ${S}/setup-nv-boot-control.sh.in >${B}/setup-nv-boot-control.sh


kekiefer:
I’m traveling this week but would like to take the time to properly review this. 

My first impression based on the description is that the motivation here is the same as the existing TNSPEC_MACHINE variable. It was added for runtime bup compatibility for custom machines but the challenge sounds familiar. It worked because NVIDIAs logic only did a partial string comparison to check if an asset was compatible, so a devkit machine name via TNSPEC_MACHINE could get glued to the custom machine name used in bup assets and it would be accepted as valid. At least IIRC, this was a while ago. 

Can you comment on whether using TNSPEC_MACHINE already does what you want, or at least if it is possible to reuse this variable rather than adding a new one? At least in our case we would have this variable set already for our custom machine.

dwalkes:
> Can you comment on whether using `TNSPEC_MACHINE` already does what you want, or at least if it is possible to reuse this variable rather than adding a new one? At least in our case we would have this variable set already for our custom machine.

Meaning you have TNSPEC_MACHINE set to `jetson-agx-xavier-devkit` on your custom MACHINE instead of the [default value of MACHINE](https://github.com/OE4T/meta-tegra/blob/34c6b88928da7996bb99f7f04dd06bcd60cb64b5/conf/machine/include/tegra-common.inc#L22).

If so I think that would work to resolve this case, however I think this would also mean any existing capsule spec checks would fail, so I'm not sure about changing this as a resolution for any existing deployment which wants to add support for slot switches.  I'd need to test what happens when both TegraPlatformSpec and TegraPlatformCompatSpec change but I assumed this would mean any capsules previously generated wouldn't match anymore.

kekiefer:
For example I set it to ‘TNSPEC_MACHINE:append = "-jetson-xavier-maxn"’ for a custom MACHINE name so it would get a concatenated name but still be valid.

dwalkes:
>  I'd need to test what happens when both TegraPlatformSpec and TegraPlatformCompatSpec change but I assumed this would mean any capsules previously generated wouldn't match anymore.

I verified this is the case... if I start with a spec which looks like this:

```
root@VC1078:~# cat /opt/nvidia/esp/EFI/NVDA/Variables/TegraPlatformSpec-781e084c-a330-417c-b678-38e696380cb9
2888-400-0004-P.0-1-2-mymachine-
root@VC1078:~# cat /opt/nvidia/esp/EFI/NVDA/Variables/TegraPlatformCompatSpec-781e084c-a330-417c-b678-38e696380cb9
2888-400-0004--1-2-mymachine-
```
and then attempt to upgrade to a capsule built with `TNSPEC_MACHINE="${MACHINE}-jetson-agx-xavier-devkit"`

The capsule update fails.

However if I update with this commit and a change in my layer to set
```
TNSPEC_COMPAT_MACHINE = "jetson-agx-xavier-devkit"
```
The capsule update succeeds and I get 
```
root@VC1078:~# cat /opt/nvidia/esp/EFI/NVDA/Variables/TegraPlatformSpec-781e084c-a330-417c-b678-38e696380cb9
2888-400-0004-P.0-1-2-mymachine-
root@VC1078:~# cat /opt/nvidia/esp/EFI/NVDA/Variables/TegraPlatformCompatSpec-781e084c-a330-417c-b678-38e696380cb9
2888-400-0004--1-2-jetson-agx-xavier-devkit-
```

So I think there are 3 reasons to use this change over a direct modification of TNSPEC_MACHINE.
1. Supports an OTA upgrade scenario if you didn’t have TNSPEC_MACHINE set to include the compatible machine initially
2. Matches what NVIDIA recommends [in the forum](https://forums.developer.nvidia.com/t/nvbootctrl-hardcoded-to-only-handle-agx-xavier/270411/4) and for their distro.
3. Avoids any possible issues with tnspec length which we’ve seen previously at https://github.com/OE4T/meta-tegra/issues/638 for machine names longer than 20ish characters.  My tnspec ended up being 62 bytes so I didn’t have any issues but I think a machine closer to 31 chars might be over the 64 char limit.

I think the changes are low risk since if you don’t define TNSPEC_COMPAT_MACHINE everything should be unchanged in your specs.  I could probably make an even smaller change which didn’t modify the rootfs content at all unless TNSPEC_COMPAT_MACHINE was set if that’s important.

That said, if you aren’t comfortable merging this in I’ll put in my local layer.  I suspect it probably only matters for r35.x branches and agx-xavier so perhaps nobody else is going to find this one or if they do they can use this as a reference to put in their own layer.

kekiefer:
Thanks for looking into this, and helping me understand the motivation and scope of changes here. I think all 3 of your points are good and valid, and I don't see a downside here considering 

>  if you don’t define TNSPEC_COMPAT_MACHINE everything should be unchanged in your specs.

That should cover my concern, that we set the TNSPEC_MACHINE up front, to specifically address this problem. But that wasn't the recommended solution, as you've pointed out.

> I could probably make an even smaller change which didn’t modify the rootfs content at all unless TNSPEC_COMPAT_MACHINE was set if that’s important.

I can't think of a reason why this would be a problem.

Thanks for making these changes. Looks like we should have them in place for kirkstone and scarthgap-r35.

kekiefer (approved):

dwalkes:
@madisongh already backported to scarthgap-l4t-r35.x with https://github.com/OE4T/meta-tegra/commit/e836736de0a439c8893ede4cef6973bd6347affa - thanks!

I've updated https://github.com/OE4T/meta-tegra/wiki/Creating-a-custom-MACHINE#jetson-agx-xavier to cross-link here and add a note about defining `TNSPEC_COMPAT_MACHINE`.

PR #1882, opened 2025-04-11
nvidia-kernel-oot: support external device-tree recipes

SJLC:
When device-tree builds moved to nvidia-kernel-oot, the dtb files built from the platform dts files were getting stashed into the sysroot area for future use during image build, but the dts files themselves were *not* saved anywhere. This broke external device-tree recipes that wanted to include the platform dts and add modifications.

Add export of dts files into nvidia-kernel-oot sysroot area, under usr/src/device-tree subdir.


kekiefer:
I tried this change with my custom external devicetree build, instead of pulling the whole L4T archive into that recipe. It is working well for me.

kekiefer (approved):

PR #1837, opened 2025-02-13
L4T R36.4.3

madisongh:

PR #1883, opened 2025-04-19
[scarthgap] Update to L4T R36.4.3

madisongh:


dwalkes:
Basic smoke test of swupdate, docker, and deepstream all look good on demo-image-full and jetson-orin-nano-devkit-nvme, thanks!  Updated https://docs.google.com/spreadsheets/d/1EWG5QZvtWdk74gz48D4xCbegK-bsqMFJ8XOsyGdUUeI/edit?gid=1760631514#gid=1760631514 and added a scarthgap column.  Deepstream python issues are gone on scarthgap, I put a note about these at https://github.com/OE4T/meta-tegra/pull/1837#issuecomment-2816941216

ichergui (approved):
LGTM. Thanks @madisongh

recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot_git.bb
truls, line 1: In my opinion, it would be nice to keep this recipe. It's useful for folks running scarthgap but with a more recent kernel than 6.6.

madisongh: I was concerned about maintaining it, since it won't get much, if any, testing in this branch, so I left it out. Plus the recipe, as it stands, gets preference over the stock L4T version, which isn't suitable for a stable branch.

I'd be willing to consider it as a separate PR later, though.

PR #1891, opened 2025-05-03
meta: update L4T BSP from 35.6.0 to 35.6.1

dchvs:
meta: update L4T BSP from 35.6.0 to 35.6.1
* Refreshed patches as needed
* Return use of ${L4T_VERSION} from names of 'driver'
  libraries


madisongh:
Thanks @dchvs for getting this started!

I've used this as the basis for a new [WIP branch](https://github.com/OE4T/meta-tegra/tree/wip-kirkstone-l4t-r35.6.1) for gathering all the changes we'll need for this update, and have added a couple of things there.

We'll need to add in the 'super' mode changes to the machine configurations, and it looks like there are some tweaks to the flashing process, at least for the Xavier platforms, that might need to be made, too.

madisongh:
Will switch this over to the WIP branch that has needed additions.

PR #1893, opened 2025-05-05
nvidia-kernel-oot: support module symbol reuse

alonbl:
The module media and other modules are reusable for other modules, the package
should expose symbols to other packages.
    
The `module.bbclass` expects `kernel-module-` prefix package name and matching
include directory to trigger the dependency mechanism.
    
```
        python __anonymous () {
            depends = d.getVar('DEPENDS')
            extra_symbols = []
            for dep in depends.split():
                if dep.startswith("kernel-module-"):
                    extra_symbols.append("${STAGING_INCDIR}/" + dep + "/Module.symvers")
            d.setVar('KBUILD_EXTRA_SYMBOLS', " ".join(extra_symbols))
        }
```
 
Add:
1. `kernel-module-nvidia-kernel-oot` `PROVIDES` alias.
2. Symlink of `/usr/include/kernel-module-nvidia-kernel-oot -> /usr/include/nvidia-kernel-oot`
    
A module that requires these modules can use the standard dependency such as:

```    
      DEPENDS += "kernel-module-nvidia-kernel-oot"
```

`module.bbclass` [fix](https://github.com/yoctoproject/poky/pull/21) is also required.

### DIFF
--- a/recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot.inc
+++ b/recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot.inc
@@ -69,6 +69,7 @@ do_install() {
     install -d ${D}/boot/devicetree
     install -m 0644 ${B}/kernel-devicetree/generic-dts/dtbs/* ${D}/boot/devicetree/
     install -d ${D}${includedir}/${BPN}
+    ln -s ${BPN} ${D}${includedir}/kernel-module-${BPN}
     find ${B} -name Module.symvers -type f | xargs sed -e's:${B}/::g' >${D}${includedir}/${BPN}/Module.symvers
 
     cp -R ${S}/nvidia-oot/include/* ${D}/${includedir}/${BPN}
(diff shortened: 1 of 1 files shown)


madisongh:
Is the patch to `modules.bbclass`  you mention required for the nvidia-kernel-oot recipe to install properly, or is it needed for other recipes to reference the symbols exported by the OOT drivers?

alonbl:
> Is the patch to `module.bbclass` you mention required for the nvidia-kernel-oot recipe to install properly, or is it needed for other recipes to reference the symbols exported by the OOT drivers?

Hi,

It is required for any other module that contain more than one module in their tree with hierarchy, as the `KBUILD_EXTRA_SYMBOLS` should be populated to install as well.

Unrelated to this specific issue, I added it here for completeness for end-to-end solution with camera drivers complex drivers.

Thanks,
Alon

ichergui:
Hi @alonbl 
Did you create an issue with some details so we can understand better the purpose of this PR ?

alonbl:
> Hi @alonbl Did you create an issue with some details so we can understand better the purpose of this PR ?

Hello @ichergui.

I anything missing in the description of this issue? I will be happy to explain more.
The root issue is that the nvidia-kernel-oot naming does not comply with the yocto convention so the `module.bbclass` cannot leverage the feature of symbol dependency to other modules.

The solution is to add an alias and a symlink for this to work and keep backward compatibility.

Thanks,
Alon

PR #1897, opened 2025-05-13
nvidia-kernel-oot: fix linux-yocto build

ohporter:
of_property_for_each_u32() only takes three arguments in newer kernels.

### DIFF
--- a/recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot/0001-nvidia-kernel-oot-handle-of_property_for_each_u32-ap.patch
+++ b/recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot/0001-nvidia-kernel-oot-handle-of_property_for_each_u32-ap.patch
@@ -1,0 +1,82 @@
+From c833d387665922f619cb4c6a551a4d993e38bbb6 Mon Sep 17 00:00:00 2001
+From: Matt Porter <mporter@konsulko.com>
+Date: Wed, 14 May 2025 11:32:21 -0400
+Subject: [PATCH] nvidia-kernel-oot: handle of_property_for_each_u32() api
+ changes
+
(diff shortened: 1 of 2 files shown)


madisongh (changes requested):
Ah, I see the API change was back-ported to 6.6.75, so this is fairly recent breakage.

It would be better to implement this by porting the parts of the [upstream 575.51.02 changes](https://github.com/NVIDIA/open-gpu-kernel-modules/commit/41595798880394d9c6fa5e76b37ffc1bc428cfb2) that added this check, extracting just the specific check in the `conftest.sh` script and the corresponding conditional in the driver code, rather than conditionalizing the patch inclusion in the recipe.

ohporter:
> It would be better to implement this by porting the parts of the [upstream 575.51.02 changes](https://github.com/NVIDIA/open-gpu-kernel-modules/commit/41595798880394d9c6fa5e76b37ffc1bc428cfb2) that added this check, extracting just the specific check in the `conftest.sh` script and the corresponding conditional in the driver code, rather than conditionalizing the patch inclusion in the recipe.

Thanks for the quick response. Agreed this is a better approach. Thanks for the pointer, I'll update accordingly.

madisongh (changes requested):
Thanks for this... if you could update the patch description with an `Upstream-Status:` line -- ref [here](https://docs.yoctoproject.org/5.0.9/singleindex.html#patch-upstream-status) -- we should be god to go.

madisongh:
@ohporter Sorry for not being clear, but the Upstream-Status: line goes into the patch file you're adding here, not the message for the commit.

madisongh (approved):

ohporter:
> @ohporter Sorry for not being clear, but the Upstream-Status: line goes into the patch file you're adding here, not the message for the commit.

Probably more due to me updating this while sitting in a conference...bad at multi-tasking. 😄

PR #1937, opened 2025-07-12
make-sdcard: Improve script to create filesystems if needed

minicarpet:
If the `filesystem_type` field in the NVIDIA XML layout is set to a value other than `'basic'`, the script will prepare the partition using the specified type.

According to the [official documentation](https://docs.nvidia.com/jetson/archives/r36.4.4/DeveloperGuide/AR/BootArchitecture/PartitionConfiguration.html#partition-child-elements), only 'basic' is expected, and support for ext2 is deprecated. However, with this modification, other filesystem types remain supported.

This change can be particularly useful for users who need to define custom partition types beyond what the default layout allows.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
@@ -148,7 +148,7 @@ create_filesystems() {
     for pline in "${PARTS[@]}"; do
 	eval "$pline"
 	if [ -z "$partfile" ] && [ -n "$fstype" ] && [ "$fstype" != "basic" ]; then
-	    printf "Creating $fstype filesystem to /dev/$DEVNAME$PARTSEP$partnumber"
+	    printf "Creating $fstype filesystem to /dev/$DEVNAME$PARTSEP$partnumber\n"
 	    mke2fscmd="mkfs.$fstype /dev/$DEVNAME$PARTSEP$partnumber"
 	    if ! eval "$mke2fscmd" >/dev/null 2>"$errlog"; then
 		    echo "ERR: filesystem failed" >&2
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
@@ -141,6 +141,27 @@ make_partitions() {
     return 0
 }
 
+create_filesystems() {
+    local partnumber
+    local pline mke2fscmd fstype
+		local errlog=$(mktemp)
+    for pline in "${PARTS[@]}"; do
+			eval "$pline"
+	    if [ -n "$fstype" ] && [ "$fstype" != "basic" ]; then
+				printf "Creating $fstype filesystem to /dev/$DEVNAME$PARTSEP$partnumber"
+				$mke2fscmd="mke2fs -t $fstype /dev/$DEVNAME$PARTSEP$partnumber"
+				if ! eval "$mke2fscmd" >/dev/null 2>"$errlog"; then
+					echo "ERR: filesystem failed" >&2
+					cat "$errlog" >&2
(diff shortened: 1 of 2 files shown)


recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
```diff
@@ -451,6 +472,7 @@ if ! sgdisk "$output" --verify >/dev/null 2>&1; then
     echo "ERR: verification failed for $output" >&2
     exit 1
 fi
+create_filesystems || exit 1
```
madisongh, line 475: This should go after, rather than before, the `partprobe` below, and probably also belongs in the `if`/`then` section, since it only applies if writing to a block device.

recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
```diff
@@ -141,6 +141,27 @@ make_partitions() {
...
 
+create_filesystems() {
+    local partnumber
+    local pline mke2fscmd fstype
+		local errlog=$(mktemp)
+    for pline in "${PARTS[@]}"; do
+			eval "$pline"
+	    if [ -n "$fstype" ] && [ "$fstype" != "basic" ]; then
+				printf "Creating $fstype filesystem to /dev/$DEVNAME$PARTSEP$partnumber"
+				$mke2fscmd="mke2fs -t $fstype /dev/$DEVNAME$PARTSEP$partnumber"
```
madisongh, line 152: It might be better to use `mkfs.$fstype` here, to allow for other filesystem types than ext2/3/4.

recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
```diff
@@ -141,6 +141,27 @@ make_partitions() {
     return 0
 }
 
+create_filesystems() {
```
madisongh, line 144: The whitespace in these added lines should match up with what's in the rest of the file. Admittedly, the existing whitespace is somewhat peculiar (thanks, emacs!), with 4-space indentation but hard tabs on an 8-space boundary.

madisongh (changes requested):
This addition is OK, with some changes (see in-line).
There some caveats about using this, though:
* NVIDIA's flashing tools could parse the filesystem type and generate errors on values they don't support - if not now, at some point in the future.
* Your host system's `e2fsprogs` may be a different version than what is on the target device, so the target might not be able to mount the filesystem you created.

recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
```diff
@@ -141,6 +141,27 @@ make_partitions() {
...
+    local partnumber
```
madisongh, line 145: You should declare all of the variables that get set via the `eval "$pline"` as local here, as is done in the `write_partitions_to_{device,image}`  functions, and those functions should be updated to add the new `fstype` to their declarations (and it looks like they're missing at least `start_location` from theirs, too, oops).

recipes-bsp/tegra-binaries/tegra-helper-scripts/make-sdcard.sh
```diff
@@ -141,6 +141,27 @@ make_partitions() {
...
+    local pline mke2fscmd fstype
+    local errlog=$(mktemp)
+    for pline in "${PARTS[@]}"; do
+	eval "$pline"
+	if [ -n "$fstype" ] && [ "$fstype" != "basic" ]; then
```
madisongh, line 150: You should probably also check for `[ -z "$partfile" ]` here, since it doesn't make sense to format a partition if it's just going to get written over with a filesystem image later.

madisongh (changes requested):
Just about there, see in-line for a couple of further changes

PR #1950, opened 2025-07-30
perf: add libtraceevent

newlnxbd:
Add libtraceevent to enable tracepoint support.

This fixes the following error when using perf commands:
 perf probe -x /lib/libc.so.6 -a malloc
   Added new events:
     probe_libc:malloc    (on malloc in /usr/lib/libc.so.6)
     probe_libc:malloc    (on malloc in /usr/lib/libc.so.6)
   perf is not linked with libtraceevent, to use the new probe you can use tracefs:
        cd /sys/kernel/tracing/
        echo 1 > events/probe_libc/malloc/enable
        echo 1 > tracing_on
        cat trace_pipe
        Before removing the probe, echo 0 > events/probe_libc/malloc/enable
 perf record -e probe_libc:malloc -a ls
   event syntax error: 'probe_libc:malloc'
                       \___ unsupported tracepoint

### DIFF
--- a/recipes-kernel/perf/perf.bbappend
+++ b/recipes-kernel/perf/perf.bbappend
@@ -1,4 +1,0 @@
-PACKAGECONFIG:tegra ?= "tui libunwind"
-# Need to enable -fcommon on C compilations and disable POSIX yacc
-# mode in bison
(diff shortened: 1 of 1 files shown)


kekiefer:
> It might make sense to just remove it altogether 

Yes, that sounds good to me

PR #1966, opened 2025-08-15
l4t_sign_image.sh patch: fix missing 'echo' in logging line

eg9152:
The patch for l4t_sign_image.sh seems to have removed "echo" from the logging line. I've had a patch for this patch file in my BSP layer for a while and I decided I would get it submitted.


madisongh:
Thanks!

PR #1967, opened 2025-08-15
scripts and recipes: fix indentation spaces/tabs mix to be just spaces

eg9152:
These files all have a mix of tabs and space for indentation, which causes issues viewing depending on your tabs setting.  This changes all the tabs to either 4 spaces or 8 spaces, depending on whether the person that put them in had theirs set to 4 or 8 spaces.  I could tell from the surrounding lines that are using spaces as I went through each file.

### DIFF
--- a/classes-recipe/image_types_tegra.bbclass
+++ b/classes-recipe/image_types_tegra.bbclass
@@ -187,10 +187,10 @@ tegraflash_create_flash_config() {
         -e"s,APPSIZE,${ROOTFSPART_SIZE}," \
         -e"s,RECROOTFSSIZE,${RECROOTFSSIZE}," \
         -e"s,APPUUID_b,," -e"s,APPUUID,," \
-	-e"s,ESP_FILE,${ESP_FILE}," -e"/VARSTORE_FILE/d" \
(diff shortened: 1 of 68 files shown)


eg9152:
@madisongh: Let me know if there is a different way you want to handle this.  I know there are a lot of files.  And there will be a few conflicts to cherry-pick it into the other branches.  And there are a few other files that need to be updated in the other branches that aren't in master.  (I'm interested in getting this into scarthgap.)  But I'm happy to do that.  I got into a good flow with searching and updating files.

madisongh:
@eg9152 Sorry, I'm still thinking about this. This topic has come up before, and while I understand it can be difficult to configure some text editors to handle the mixed tab/space indentation arrangement that comes naturally to emacs (although TBH even it messes up sometimes), I'm also not a big fan of essentially gratuitous whitespace changes, particularly on this scale.  Besides the cherry-picking issue you already mentioned, it also makes it harder to trace the history of substantive changes.

So I've been resistant to making such changes, but I'm not entirely opposed to the idea - being consistent in the indentation is appealing, too. I'm not sure I would be in favor of back-porting such changes to any of the stable branches, though.

PR #1979, opened 2025-09-01
uefi signing: tracks UEFI DB files with file-checksums 

gabrielssanches:
This makes it possible for the build to track key/cert files defined by TEGRA_UEFI_DB_KEY/TEGRA_UEFI_DB_CERT.

Changes made to those files are kept up to date in the build.

For this to work, instead of specifying the full path for the key/cert files in TEGRA_UEFI_DB_KEY/TEGRA_UEFI_DB_CERT user must specify the file names.
The path for where key/cert are located must be provided in the variable TEGRA_UEFI_DB_PATH

### DIFF
--- a/classes-recipe/tegra-uefi-signing.bbclass
+++ b/classes-recipe/tegra-uefi-signing.bbclass
@@ -6,11 +6,18 @@ def tegra_uefi_signing_deps(d, tasks=False):
         return ' '.join([d + ':do_populate_sysroot' for d in deps])
     return ' '.join(deps)
 
+def tegra_uefi_signing_filechecksums(d):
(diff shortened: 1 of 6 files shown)


madisongh (changes requested):
Rather than introducing a new variable and modifying SRC_URI, it would be more backward compatible to directly add references to the key and cert files to the `[file-checksums]` flag on the tasks that use them. That should still get you the desired end result of causing task hash changes when the key/cert files change.

gabrielssanches:
changes made

gabrielssanches:
a bunch of warnings about '.' being a directory.

gonna try your suggestion

ichergui (approved):
LGTM. Thanks @gabrielssanches

madisongh (approved):

PR #1997, opened 2025-09-12
tegra-eks-image: adds custom EKS image (EKB binary) support

gabrielssanches:
tegra-bootfiles: adds option to use custom eks image
    
    users can use the eks image built by tegra-eks-image-custom
    by setting USE_PREBUILT_EKS_IMAGE = "1"
    
    tegra-eks-image-custom: Generates custom EKS image
    
    Builds a custom eks image (EKB binary) usign gen_ekb.py
    tool provided by NVIDIA.
    
    user needs to set these variables with corresponding key files
    for it to work:
    
    TEGRA_EKB_OEM_K1: The OEM_K1 key stored in the OEM_K1 fuse. If secure boot fuse is not set, this can be 0000000000000000000000000000000000000000000000000000000000000000.
    TEGRA_EKB_SYM2:   (optional, but recommended) The disk encryption key. hwkey-agent and luks-srv use it.
    TEGRA_EKB_AUTH:   The UEFI variable authentication key.


Lexmark-chad:
I agree that it would be helpful for others if `meta-tegra` could handle generating the ekb if keys are provided.  But, that can be done in the existing `tegra-eks-image` and not have to require a new recipe.  Just my $0.02.

conf/machine/include/tegra-common.inc
```diff
@@ -49,6 +49,8 @@ PREFERRED_PROVIDER_tensorrt-plugins ?= "tensorrt-plugins-prebuilt"
 PREFERRED_PROVIDER_tensorrt-trtexec ?= "tensorrt-trtexec-prebuilt"
 PREFERRED_RPROVIDER_tegra-gbm-backend ?= "tegra-libraries-gbm-backend"
 
+USE_PREBUILT_EKS_IMAGE ?= "1"
```
Lexmark-chad, line 52: Consider moving this to the recipe and perhaps deriving it from one or more non-empty keys.

gabrielssanches:
> Looks good. An update to your commit message is in order (`if USE_PREBUILT_EKS_IMAGE is '0'`), but I think it's ready for @madisongh to take a look now.

commit re-worded, thanks for pointing it out

PR #2001, opened 2025-09-16
optee: adds custom OP-TEE TA signing support

gabrielssanches:
By default, a development/test key from the upstream OP-TEE source is compiled in.

This enables for a custom key by setting TEGRA_OPTEE_TA_SIGN_KEY to either a RSA 2048 or 4096 key file.

The change is done in optee-l4t.inc, which is included by all recipes that are related to OP-TEE. That means the custom key is also used to sign optee-client, optee-nvsamples and so on.


gabrielssanches:
as in the comment, it is not enough.
e.g.:
nvhwkey fails because it has been compiled with the default.pem instead of TA_SIGN_KEY

ichergui:
I don't understand your reply `as in the comment, it is not enough.` 
I used the bbappend strategy and the TA signing works as expected.

PR #2005, opened 2025-09-23
tegra-uefi-keys-dtb: automates UEFI device tree file generation

gabrielssanches:
Generates UEFI device tree files instead of doing outside of yocto.
This covers generation of UefiDefaultSecurityKeys.dts/dtbo
The minimal set of keys and certs is:
    
    TEGRA_UEFI_PK_CERT
    TEGRA_UEFI_KEK_1_CERT
    TEGRA_UEFI_DB_1_KEY
    TEGRA_UEFI_DB_1_CERT
    
tegra-flashtools-native: installs gen_uefi_keys_dts.sh


madisongh:
I'd prefer it if the existing recipe retains its current behavior by default, so as not to break builds that use the bbappend approach.

While automating the DTS file generation isn't a bad idea, I'm a bit worried that having a recipe do that generation would make it too easy to expose secrets, since users are apt to just check in the `db_1.key` and `db_1.cert` files into git as part of a layer.

As for the missing tool, we could add a recipe to build just the needed tool for this purpose.

ichergui:
Hi @gabrielssanches 
Any update ? 
Or we should close this PR.

ichergui:
Closing this PR since no updates from October 2025. Please re-submit new PR if needed

PR #2015, opened 2025-09-29
Remove obsolete fuse programming support

madisongh:
Fuse programming is more complex with the more recent Jetsons, so we should remove the mechanism left over from the earlier generations (and R32.x L4T) to prevent misunderstanding (and possible irreversible errors).


ichergui (approved):
Thanks @madisongh

PR #2025, opened 2025-10-02
Update recipes for standalone MM built from source for tegra264

ichergui:


madisongh (changes requested):
The `BUILD_GUID=` setting in  `TEGRA_BUILD_GUID` also needs an update, to match the GUID returned by `Platform/NVIDIA/StandaloneMmJetson/PlatformBuild.py`. Otherwise, this looks good

ichergui:
Thanks @madisongh. I will push my changes shortly.

PR #2047, opened 2025-10-17
python3-tensorrt: uprev to version 10.13.3

pberginkonsult:
Current python3-tensorrt version failed to build. Uprev to new version and get a successful build.

One new define 'TENSORRT_PLUGIN_DISABLED' is introduced and is hardcoded to "0" in recipe as it was before.


madisongh (changes requested):
I'm not sure upgrading is the right solution for the build failure here. Just adding `-DCMAKE_POLICY_VERSION_MINIMUM=3.5` to `EXTRA_OECMAKE` should be sufficient.
 
I don't know how closely coupled the Python bindings are, version-wise, to the TensorRT libraries, but we've usually tried to keep the versions in sync.

Issue #2018, opened 2025-09-29
[R38.2.0] Fix clunky flashing procedure


madisongh:
With the transition to NVIDIA's "unified" flashing scripts, we currently generate the tegraflash tarball, from which you run `initrd-flash` to generate the workspace for those scripts, and that in turn generates a wrapper script for running the unified flash.

This needs to be streamlined, eliminating the intermediate step. It would be nice if we could generate the unified-flash workspace directly as part of the tegraflash package generation during the build, but it's unclear whether we need to talk to the target device to get the artifacts in the correct form.

We also should get rid of the `doflash.sh` and `dosdcard.sh` scripts. The `initrd-flash` should probably be renamed to something more appropriate.

madisongh:
With R38.2 and Thor targets, the flashing process in Jetson Linux has changed to use NVIDIA's "unified" flashing tools.

The bad news is that this involves even more scripts, and more layers of scripts, for the new process. The good news is that the steps in the flashing flow are a bit more regularized, and we can take advantage of that change.

The new flow is:
1. Binaries are prepared for flashing -- generating/signing boot firmware, creating partition tables for the external boot device, and generating the RCM blobs for reading the chip info/CVM EEPROM and for booting the flashing initrd. Since this applies to both secured and non-secured devices, all of this can now be done during the build.
2. The files generated in step 1 are staged and converted into a form for consumption by the unified flashing tools. The same conversion script used by the Jetson Linux flash tools is used here, so we don't have to try to duplicate that ourselves.
3. The unified flashing tools are called on to do the flashing.

Part of this work is cleaning out support for older modules that required additional processing, or that had a different flashing flow.

In addition, we should also simplify the mechanism by which we pass information into the flashing scripts, at least somewhat.

PR #2051, opened 2025-10-22
Flashing process rework

madisongh:
For #2018 

I think I now have enough of this done to have it reviewed and tested. There's still some further work to come, but with what's here now, you should be able to generate a tegraflash package that you can just run `./initrd-flash` to do the flashing, instead of the two-step hack I had earlier.

The new script supports `--external-only` (aka `--skip-bootloader`) for flashing just the external storage drive, and `--qspi-only` for flashing just the QSPI boot firmware.  (Note that the script has a `-k` / `--partition` option that it accepts for updating just a single partition, but that's not yet hooked up.)

In addition, there's more processing done during the build that should speed up execution of the `initrd-flash` script.

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
+++ b/recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
@@ -103,13 +103,14 @@ while true; do
                 echo "ERR: specify only one of --external-only, --qspi-only, --partition" >&2
                 exit 1
             fi
+            uniflash_flags="$uniflash_flags -u $partition_name"
             ;;
         -h|--help)
             usage
             exit 0
             ;;
         -D|--debug)
-            uniflash_flags="-D"
+            uniflash_flags="$uniflash_flags -D"
             shift
             ;;
         --)
@@ -319,6 +320,9 @@ mkdir out
 mkdir -p out/flash_workspace/flash-images out/flash_workspace/rcm-boot
 ./create_l4t_bsp_images.py $convargs --info --dest $PWD/out
 ./create_l4t_bsp_images.py $convargs --dest $PWD/out/flash_workspace/flash-images
+if [ -n "$partition_name" ]; then
+    ./create_l4t_bsp_images.py $convargs -k $partition_name --dest $PWD/out/flash_workspace/flash-images
+fi
 ./create_l4t_bsp_images.py $convargs --dest $PWD/out/flash_workspace/rcm-boot --rcm-boot
 cp -R out/flash_workspace/rcm-boot out/flash_workspace/rcm-flash
 cat > out/doflash.sh <<EOF


recipes-bsp/tegra-binaries/tegra-helper-scripts/tegra-flash-helper.sh
```diff
@@ -24,24 +24,10 @@ blocksize=4096
...
-}
-
-process_l4t_conf_dtbo() {
-    :
-}
-
 partition_exists_in_PT_table() {
-    [ "$1" = "BCT-boot-chain_backup" ]
+    local partname="$1"
+    local layoutfile="$2"
```
kraj: while we demand /bin/bash as shell, I would suggest to avoid bashism when possible, here we can avoid using `local` keyword and pass $1 and $2 directly to script below

madisongh: Unfortunately I long ago had to give up on unbashifying the script, since I couldn't easily come up with ways of translating some of the bashisms I had to borrow from NVIDIA's script into POSIX shell that would produce the exact same results.

classes-recipe/image_types_tegra.bbclass
```diff
@@ -164,29 +158,40 @@ EOF
...
+    if [ -e external-flash.xml.in ]; then
+        ${TEGRA_SIGNING_ENV} MACHINE=${TNSPEC_MACHINE} ./tegra-flash-helper.sh --sign --no-flash --external-device ${TEGRA_SIGNING_ARGS} external-flash.xml.in ${LNXFILE} ${IMAGE_BASENAME}.${IMAGE_TEGRAFLASH_FS_TYPE}
+	mv secureflash.xml external-secureflash.xml
+	mv flash.idx external-flash.idx
     fi
+    if [ -e rcmboot-flash.xml.in -a -n "${IMAGE_TEGRAFLASH_INITRD_FLASHER}" ]; then
+        ${TEGRA_SIGNING_ENV} MACHINE=${TNSPEC_MACHINE} ./tegra-flash-helper.sh --no-flash --rcm-boot ${TEGRA_SIGNING_ARGS} rcmboot-flash.xml.in initrd-flash.img ${IMAGE_BASENAME}.${IMAGE_TEGRAFLASH_FS_TYPE}
+	[ -d rcmboot_blob ] || bberror "RCM boot blob was not created"
+    fi
+    find . -type d -name __pycache__ -exec rm -rf {} +
```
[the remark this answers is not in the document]
madisongh: The existing `find` command has been in there for a long time, it's just been moved as part of these changes. Check the man page for `find` for an explanation of the `+` command terminator.

pberginkonsult:
LGTM - have tested this on a dev-kit and on a custom carrier board.

PR #2057, opened 2025-10-24
Backport from master, fixes for nvidia-kernel-oot and gcc-for-nvcc

pberginkonsult:
Bring over three commits from master to r38 branch.


madisongh:
I've rebased the WIP branch on top of master latest, rather than pulling in these changes separately.

PR #2062, opened 2025-11-06
[Scarthgap] cudnn-samples: Update runtime dependencies

leon-anavi:
Add runtime dependencies to compile and run the samples:

```    
cd /usr/src/cudnn_samples_v9
mkdir -p build && cd build
cmake ../CMakeLists.txt  \
-DCMAKE_CUDA_COMPILER=/usr/local/cuda-12.6/bin/nvcc \
-DCMAKE_CXX_FLAGS="-I/usr/local/cuda-12.6/include" \
-DCMAKE_EXE_LINKER_FLAGS="-L/usr/local/cuda-12.6/lib64 -lcudart " \
-DCMAKE_CUDA_ARCHITECTURES=87
make -j$(nproc)
    
./conv_sample/conv_sample
    
./multiHeadAttention/multiHeadAttention -attnTrain1
    
cd /usr/src/cudnn_samples_v9/mnistCUDNN/
../build/mnistCUDNN/mnistCUDNN
```

Starting with CUDA 11.2, FreeImage is no longer distributed with the CUDA Samples so it was added as a runtime dependency: https://docs.nvidia.com/cuda/archive//11.5.0/cuda-samples/index.html#new-features-in-cuda-toolkit-11-2

### DIFF
--- a/recipes-devtools/cudnn/cudnn-samples/0001-mnistCUDNN-CMakeLists.txt-Fix-FreeImage.patch
+++ b/recipes-devtools/cudnn/cudnn-samples/0001-mnistCUDNN-CMakeLists.txt-Fix-FreeImage.patch
@@ -1,0 +1,40 @@
+From 99a40ca87adb8c15cb9e454ae8e5a9c0a8e7172c Mon Sep 17 00:00:00 2001
+From: Leon Anavi <leon.anavi@konsulko.com>
+Date: Thu, 6 Nov 2025 17:07:04 +0000
+Subject: [PATCH] mnistCUDNN/CMakeLists.txt: Fix FreeImage
+
+Remove the private FreeImage and use the global freeimage.
+
+Starting with CUDA 11.2, FreeImage is no longer distributed with
+the CUDA Samples:
+https://docs.nvidia.com/cuda/archive//11.5.0/cuda-samples/index.html#new-features-in-cuda-toolkit-11-2
+
+Upstream-Status: Inappropriate [OE-Specific]
+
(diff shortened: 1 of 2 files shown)


madisongh (changes requested):
Thanks @leon-anavi , but I'm not sure I want to carry the recipe for FreeImage in the BSP layer, since it appears to be unmaintained (at least at the old SourceForge location), and has some known security issues.  It's unfortunate NVIDIA used it for one of the samples programs.

@ichergui handled this recently as part of the Thor/JetPack 7 work by moving the cudnn-samples recipe to the [meta-tegra-community](https://github.com/OE4T/meta-tegra-community/blob/wip-l4t-r38.2.0/recipes-devtools/cudnn/cudnn-samples_9.12.0.46-1.bb) repo.  I'm still not crazy about carrying the FreeImage recipe there, but at this point I think it would be a less bad option.

Another option would be to use a PACKAGECONFIG setting to conditionalize the build of the one sample that uses FreeImage, with the default being off.  That would probably require a patch to the CMake files, but we could leave the recipe here and let users decide to pull in a FreeImage recipe from one of the other layers (I see a couple of different recipes in the layer index).

leon-anavi:
> Hey @leon-anavi Are you able to try with the recipes from meta-tegra-community ? Let us know so we can close this PR Thanks

Sorry, not yet but it is on my TODO list after FOSDEM.

Should we close this GitHub pull request and open a new one in meta-tegra-community?

Also should we drop the whole cudnn-samples recipe from meta-tegra and just stick to the one from meta-tegra-community?

PR #2090, opened 2025-12-13
tegra-flash-init:init-flash.sh: start udevd

liuming50:
It was observed when the external device getting flashed on the host side, sometimes the partition layout changes are not in /dev neither in /proc, they do not reflect the real partition layout on the device.

How to reproduce:

Put the board to recovery mode and run the l4t_initrd_flash.sh from Linux_for_Tegra:

```
$ sudo ./tools/kernel_flash/l4t_initrd_flash.sh --external-device nvme0n1p1 \
-c tools/kernel_flash/flash_l4t_external.xml -p "-c bootloader/t186ref/cfg/flash_t234_qspi.xml \
--no-systemimg" --network usb0 p3509-a02+p3767-0000 externa
```

then flash it again with tegraflash generated by yocto:

```
sudo ./initrd-flash --erase-nvme
```

Starting udevd in init can fix this issue per test.

### DIFF
--- a/recipes-core/initrdscripts/tegra-flash-init/init-flash.sh
+++ b/recipes-core/initrdscripts/tegra-flash-init/init-flash.sh
@@ -22,6 +22,24 @@ for m in $MODULES_TO_LOAD; do
     modprobe -v "$m"
 done
 
+UDEVD=
+UDEVD_PATHS="/sbin/udev/udevd /sbin/udevd /lib/udev/udevd /lib/systemd/systemd-udevd"
+for path in $UDEVD_PATHS; do
+    if [ -x "$path" ]; then
+        UDEVD=$path
+        break
+    fi
+done
+
+if [ -x "$UDEVD" ]; then
+    # Workaround if console=null, systemd-udevd needs valid stdin, stdout and stderr to work
+    sh -c "exec 4< /dev/console" || { exec 0> /dev/null; exec 1> /dev/null; exec 2> /dev/null; }
+
+    $UDEVD --daemon
+    udevadm trigger --action=add
+    udevadm settle
+fi
+
 mount -t configfs configfs -o nosuid,nodev,noexec /sys/kernel/config
 
 [ ! -e /usr/sbin/wd_keepalive ] || /usr/sbin/wd_keepalive &


liuming50:
> I'm not really understanding the problem this is solving. Can you provide more information about it... what isn't showing up where, exactly, and how is that causing a problem?
> 
> I was hoping to avoid having to start services in the flashing initrd, if at all possible. Could a `partprobe` invocation be used instead, if it's just a matter of the partition table not getting loaded?

Hi, @madisongh:

The problem is sort of strange, we are building a mender enabled image with yocto, sometimes we need to flash the device with both Jetpack flasher and yocto tegraflasher, and when we flash the device firstly with jetpack and then after we flash the device with yocto tegraflash, then it's observed the partition layout is not correct.

It could be reproduced by the following steps:

1 Firstly flash the device with Jetpack:

```
$ sudo ./tools/kernel_flash/l4t_initrd_flash.sh --external-device nvme0n1p1 \
-c tools/kernel_flash/flash_l4t_external.xml -p "-c bootloader/t186ref/cfg/flash_t234_qspi.xml \
--no-systemimg" --network usb0 p3509-a02+p3767-0000 external
``` 

2 Patch meta-tegra to get some output:

```
$ git diff recipes-core/initrdscripts/tegra-flash-init/init-flash.sh
diff --git a/recipes-core/initrdscripts/tegra-flash-init/init-flash.sh b/recipes-core/initrdscripts/tegra-flash-init/init-flash.sh
index b3e5c43..2a7c2ed 100644
--- a/recipes-core/initrdscripts/tegra-flash-init/nit-flash.sh
+++ b/recipes-core/initrdscripts/tegra-flash-init/init-flash.sh
@@ -222,6 +222,8 @@ else
                            if wait_for_disconnect 2>&1 >> /tmp/flashpkg/flashpkg/logs/export-$dev.log; then
                                sgdisk /dev/$dev --verify 2>&1 >> /tmp/flashpkg/flashpkg/logs/export-$dev.log
                                sgdisk /dev/$dev --print 2>&1 >> /tmp/flashpkg/flashpkg/logs/export-$dev.log
+                               cat /proc/partitions >> /tmp/flashpkg/flashpkg/logs/export-$dev.log
                                continue
                            fi
                        fi
```

the output of sgdisk and "cat /proc/partitions" is like:

```
Number  Start (sector)    End (sector)  Size       Code  Name
   1         2238464        30943231   13.7 GiB    0700  APP
   2        30943232        59647999   13.7 GiB    0700  APP_b
   3            2048          264191   128.0 MiB   0700  A_kernel
   4          264192          265727   768.0 KiB   0700  A_kernel-dtb
   5          266240          331007   31.6 MiB    0700  A_reserved_on_user
   6          331776          593919   128.0 MiB   0700  B_kernel
   7          593920          595455   768.0 KiB   0700  B_kernel-dtb
   8          595968          660735   31.6 MiB    0700  B_reserved_on_user
   9          661504          825343   80.0 MiB    0700  recovery
  10          825344          826367   512.0 KiB   0700  recovery-dtb
  11          827392          958463   64.0 MiB    EF00  esp
  12          958464         1122303   80.0 MiB    0700  recovery_alt
  13         1122304         1123327   512.0 KiB   0700  recovery-dtb_alt
  14         1124352         1255423   64.0 MiB    0700  esp_alt
  15        59648000        60467199   400.0 MiB   0700  UDA
  16         1255424         2237439   479.5 MiB   0700  reserved
major minor  #blocks  name
```

```
   7        0    3358720 loop0
  31        0      65536 mtdblock0
 259        0  250059096 nvme0n1
 259        1   57671680 nvme0n1p1
 259        2      65536 nvme0n1p2
 259        3        448 nvme0n1p3
 259        4      32768 nvme0n1p4
 259        5      65536 nvme0n1p5
 259        6        448 nvme0n1p6
 259        7      32768 nvme0n1p7
 259        8      81920 nvme0n1p8
 259        9        512 nvme0n1p9
 259       10     307200 nvme0n1p10
 259       11      65536 nvme0n1p11
 259       12      81920 nvme0n1p12
 259       13        512 nvme0n1p13
 259       14      65536 nvme0n1p14
```

see the sgdisk showing the device is already flashed on partition 1  and 2 with 13.7 GiB images, but the device nodes of partition 1 and 2 are still remaining the old layouts flashed from Jetpack.

Only if I start udevd, then the layout would be updated with no problems.

madisongh:
Thanks for clarifying. Yes, the kernel's copy of the partition table needs refreshing after the device is exported. Can you try with #2091 ? That's the approach I was referring to in my earlier comment, and, in my testing, also fixes the underlying issue.

liuming50:
> Thanks for clarifying. Yes, the kernel's copy of the partition table needs refreshing after the device is exported. Can you try with #2091 ? That's the approach I was referring to in my earlier comment, and, in my testing, also fixes the underlying issue.

Hi, @madisongh that fix works! Thanks, will close this one.

PR #2094, opened 2025-12-16
tegra-flash-init: add partprobe during device export

liuming50:
If a storage device has a partition table prior to being exported via USB, the kernel retains that original partition table after the USB export, regardless of what the host system writes to the device.

So once the export has been released by the host, run partprobe on the device to re-sync the kernel with the actual partition table, to avoid later confusion.


liuming50:
@madisongh can we backport this fix to LTS scarthgap?

PR #2135, opened 2026-01-30
cuda-cccl: refresh patch

ichergui:
Avoid compile errors from math function like macros
```
error: '::isnan' has not been declared; did you mean 'std::isnan'? [-Wtemplate-body]
   43 |     return ::isnan(__x);
      |              ^~~~~
      |              std::isnan
```
Those libc functions might not be available or defined as macros. 
We can work around this issue by just dropping the global qualification `::`

This will work for macros and functions alike.

This PR fixes https://github.com/OE4T/meta-tegra-community/issues/167


madisongh (changes requested):
Please update the commit message with an explanation as to why this patch change is needed. Did the old one stop working for some reason, or was it wrong, or, ... ?

Please verify that the `Upstream-Status` line in the patch itself is accurate. Is this really OE-specific, or is this another partial-sort-of-backport like the last one was (though it looks like the PR it was derived from was never merged)?

ichergui:
Hi @madisongh 
@miscco and I rework the CCCL patch based on OE build. 
Please me know if you any question.

madisongh (approved):

PR #2139, opened 2026-02-05
Upgrade r35.x

joekale-pp:
Upgrades recipes from 35.6.2/1 to 35.6.4

Current hold up is waiting to make sure EDK2 should really be 35.6.3 instead of 35.6.4 for this release.
Build tested with demo-image-full for jetson-orin-nano-devkit machine. Needs testing on device.


joekale-pp:
Just pushed an update to relabel those branches and recipes as 35.6.4. `demo-image-full` builds successfully, but I have not been able to locate an actual devkit to test the resulting image. Earliest I can get to the test is probably Friday.

madisongh:
I got in some Xavier testing today, all looks good.  Thanks, @joekale-pp !

PR #2140, opened 2026-02-08
docs: refactor to use mdbook instead of wiki [Work in Progress]

dwalkes:
This PR is not yet ready to merge but creating here as a starting point for discussion and likely place for a discussion in the upcoming meeting

* For working site demo, see https://oe4t.github.io/master/
* For any pull requests before initial merge, especially for remaining tasks 1-5 below, you can target corresponding branches at https://github.com/trellis-logic/meta-tegra

## Remaining tasks:

1. Get the master branch [[SUMMARY.md](http://summary.md/)](https://github.com/Trellis-Logic/meta-tegra/blob/master/docs/SUMMARY.md) and any other content matching the preferred layout for all branches.
2. Update commit series on https://github.com/Trellis-Logic/meta-tegra/tree/scarthgap to include content from master branch.
3. Cherry commit(s) from scarthgap branches to any other branches we want documentation files deployed to (probably at least scarthgap-l4t-r35.x, maybe also kirkstone, and kirkstone-l4t-r32.7.x)
4. Update version file at https://github.com/OE4T/oe4t.github.io/blob/main/versions.json to list all branches supported.
5. Remove content from branches which are not applicable and make the docs branch content diverge from each other (can also be done later after merge to oe4t)
6. Update the meta-tegra repo to include actions repository secret `OE4T_GITHUB_DEPLOY_KEY` - either get from Dan or Generate yourself.
  * If generating yourself, add the public deploy key at https://github.com/OE4T/oe4t.github.io/settings/keys
7. Remove the deploy key from Dan's repo to avoid accidental pushes from his meta-tegra repo.
8. Merge changes from each of the branches in https://github.com/OE4T/oe4t.github.io/blob/main/versions.json into the meta-tegra repo.
9. Replace each of the wiki files content at https://github.com/OE4T/meta-tegra/wiki which are listed in the summary file at https://github.com/Trellis-Logic/meta-tegra/blob/master/docs/SUMMARY.md with links to the master version of the documentation site 
* Exclude OE4T-Meeting*
* Exclude ZZZ*
> The wiki content for this page has been replaced with the content at https://oe4t.github.io/master/link-to-page Please refer to the content of the master branch as older branches may be out of date

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,1 +1,1 @@
-../README.md
+Home.md
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -1,9 +1,9 @@
 # Summary
 
-[Introduction](Home.md)
+[Introduction](README.md)
 
 # Getting Started
-- [meta-tegra README](README.md)
+- [meta-tegra README](meta-tegra-README.md)
 - [Which branch should I use](Which-branch-should-I-use.md)
 - [Flashing the Jetson Dev Kit](Flashing-the-Jetson-Dev-Kit.md)
 - [initrd Flashing Support](initrd-flashing-support.md)
--- a/docs/meta-tegra-README.md
+++ b/docs/meta-tegra-README.md
@@ -1,0 +1,1 @@
+../README.md


madisongh:
How about we update the "Which branch should I use" page so that it has a link on it to the copy that is generated off the master branch (along with some text to say "click here for the latest version" or some such)? Then we can just do copies for back-ports, the copy of that page would be a snapshot, but there would always be a link to latest available.

BTW, I took a quick look at https://oe4t.github.io  - looks great!

dwalkes:
OK that direction makes sense to me, essentially we prioritize keeping backports of documentation pain-free at the expense of possibly having out of date and duplicate information on old branches. I think it would not be too difficult to add a warning banner on all pages other than master to state that documentation may be out of date, similar to what openembedded and other projects use. I haven't tried to implement that yet.

I did pull in all the release notes, which branch should I use and updated all the content to reflect the latest on the wiki as of today, including some changes for link resolution.  I think this is pretty close for both these two branches.

When we are ready to pull the trigger on this I think I can replace every wiki page (other than meeting notes and archived) with some text like this:

> The wiki content for this page has been replaced with the content at [https://oe4t.github.io/master/link-to-page](https://oe4t.github.io/master/link-to-page)  Please refer to the content of the master branch as older branches may be out of date

Should we organize the docs into subdirectories instead of just a flat tree?  Might be good for release notes at least.

The symlinks back to README and CONTRIBUTING files are a little odd and mean that we can't link to other documentation pages from these easily.  Would probably be better to move the main copy into the docs dir and make the symlink exist in the base repo but I'm not sure you want to do that, especially for the README.  Let me know if you'd like changes beyond this.

I've updated the summary to list the remaining steps.  Let me know if you'd like me to do anything different with commit history on the master or scarthgap branches.  Otherwise I think those are both ready to go, pending the questions above and comment in the summary of this ticket.  I'll backport to all other branches and create PRs for whatever else you want too, just let me know.

dwalkes:
Oh one other change I could make could be to try to make the docs github actions only run when the content in the docs directory changed, that might be a good thing to do too.  Need to also include README.md and CONTRIBUTING.md files pending the discussion about symlinks above.

PR #2141, opened 2026-02-09
Update L4T R36.4.4 -> R36.5.0

madisongh:
In draft while waiting for the UEFI changes for R36.5.0 to get published.

Built and booted base, weston, and full images from the demo distro, and ran a few tests on each. Looks OK.
Built and booted demo-image-base with linux-yocto 6.12, looks OK. Patches we had back-ported to make the 36.4.4  OOT drivers compatible with the 6.12 kernel could all be dropped.


madisongh:
The only remaining pending item is to update the linux-yocto support to use the 6.18 kernel.

kekiefer:
Modules will build against 6.18 for me now. Testing and patch cleanup underway.

PR #2150, opened 2026-02-17
[scarthgap] Update L4T R36.4.4 -> R36.5.0

madisongh:
This is equivalent to #2141 but targeting the scarthgap branch.

Tested with jetson-orin-nano-devkit-nvme target, demo-image-{base,sato,weston,full}.

Also ran a few tests on a sato image with the linux-yocto 6.6 kernel.


ichergui:
Hi @madisongh 
Anything I can help with here ?

PR #2157, opened 2026-02-27
tegra-target-flash-scripts: fix UDC discovery and QSPI path for t264

mihai-chiorean:
Replace hardcoded known_udc_dev checks in nv_enable_remote.sh with
dynamic /sys/class/udc/* enumeration, matching how init-flash.sh
already discovers the UDC.

Add QSPI device symlink (/dev/810c5b0000.spi -> /dev/mtd0) for t264
initrd flash, gated on tegra264 compatible.

### UDC issue

During initrd flash on AGX Thor, nv_enable_remote.sh fails to find
the USB Device Controller because it only checks hardcoded device
names (known_udc_dev1="3550000.usb", known_udc_dev2="..."). Thor's
UDC device name doesn't match either, so the script falls through
with "no UDC found" and the ADB connection never comes up.

init-flash.sh already handles this correctly with
`ls -1 /sys/class/udc | head -n 1` — this patch applies the same
approach to nv_enable_remote.sh.

### QSPI issue

During initrd flash on AGX Thor, nvdd fails to write QSPI NOR because
it looks for /dev/block/810c5b0000.spi (the t264 SPI5 MMIO address).
The Yocto kernel exposes QSPI NOR via Linux MTD as /dev/mtd0 instead.
Since /dev/block -> /dev in the initramfs, a symlink fixes the path
resolution. Gated on tegra264 to avoid affecting other platforms.

### DIFF
--- a/recipes-bsp/uefi/edk2-firmware-tegra-38.4.0.inc
+++ b/recipes-bsp/uefi/edk2-firmware-tegra-38.4.0.inc
@@ -38,6 +38,7 @@ SRC_URI += "\
     file://0003-L4TLauncher-allow-for-empty-missing-APPEND-line-in-e.patch;patchdir=../edk2-nvidia \
     file://0004-XusbControllerDxe-use-BaseMemoryLib-functions.patch;patchdir=../edk2-nvidia \
     file://0005-PrePi-use-GCC5-instead-of-GCC-in-build-options.patch;patchdir=../edk2-nvidia \
+    file://0007-BaseTools-Pccts-set-C-standard.patch \
 "
 
 S = "${UNPACKDIR}/edk2-tegra/edk2"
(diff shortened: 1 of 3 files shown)


madisongh:
@mihai-chiorean The [master-l4t-r38.4.x](https://github.com/OE4T/meta-tegra/tree/master-l4t-r38.4.x) is the current active branch for R38.4, can you retarget your PR there?

It would also help if you could provide a bit more background on the failure you're seeing that requires these changes.
Thanks!

madisongh:
@mihai-chiorean Thanks for the updated description... I've never had an issue with flashing my AGX Thor development kit with the way the script is currently. Are you using a development kit, or some kind of custom hardware setup? And is this with the stock NVIDIA kernel (linux-noble-nvidia-tegra recipe), or are you using something else?

ichergui:
Hi @mihai-chiorean 
Thanks for the PR. I agree with @madisongh 
I didn't have an issue with the flashing process. Please test it again without updates then let us know.

mihai-chiorean:
ok, jsut got a new thor and confirmed we don't need this.

PR #2182, opened 2026-03-27
linux-yocto: patches to add tegra-spidev compatible string to spidev

asterix:
UBUNTU: SAUCE: spi: add "tegra-spidev" compatible string.
BugLink: https://bugs.launchpad.net/bugs/2019240

Add "tegra-spidev" compatible string. This is required to use SPI peripherals as spidev devices. The original Nvidia patch seems denied by the linux upstream maintainers.

### DIFF
--- a/recipes-kernel/linux/linux-yocto-6.18/0012-UBUNTU-SAUCE-add-tegra-spidev-compatible-string.patch
+++ b/recipes-kernel/linux/linux-yocto-6.18/0012-UBUNTU-SAUCE-add-tegra-spidev-compatible-string.patch
@@ -1,0 +1,30 @@
+UBUNTU: SAUCE: spi: add "tegra-spidev" compatible string.
+BugLink: https://bugs.launchpad.net/bugs/2019240
+
+Add "tegra-spidev" compatible string.
+
+Relevant L4T commits:
+https://github.com/OE4T/linux-jammy-nvidia-tegra/commit/eb5afccf5a83a74c8e3a4d97850c8157080fcd10
+https://github.com/OE4T/linux-jammy-nvidia-tegra/commit/b5482920612a2161b49e5b3abdd5d94d5ba297fd
+
+Upstream-Status: Backport
+ 
+--- a/drivers/spi/spidev.c
++++ b/drivers/spi/spidev.c
+@@ -716,6 +716,7 @@
+ 	{ .name = "spi-authenta" },
+ 	{ .name = "em3581" },
+ 	{ .name = "si3210" },
++	{ .name = "tegra-spidev" },
+ 	{},
+ };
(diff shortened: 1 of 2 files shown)


madisongh (changes requested):
1) You need to update the linux-yocto bbappend to add this patch to the SRC_URI for it to have any effect.
2) I'd prefer matching the structure of the patches in the Ubuntu/NV kernel, and retaining the commit messages from the commits in their repo - so there would be two patches here. You can set `Upstream-Status:` to `Backport` and reference the commit out of the linux-jammy-nvidia-tegra kernel for each patch.

But, if you intend to use the upstream kernel, it might be better to configure your SPI device in the way that the upstream maintainers want spidev to be used. It's hard to say from the e-mail thread you referenced whether that's possible, though, given the way the NVIDIA drivers currently work.

asterix:
@madisongh yes, i missed the append, I'll fix up. I can also update to Backport, but this is technically wrong because this is not on kernel mainline. What the maintainers suggest very unclear and the conversation ends with no conclusion. The suggestion to use an existing compat string is hacky. A DT overlay is not going to to solve this problem either. So either nvidia changes their string to something that's existing or we use this patch.

Btw, 'linux,spidev' and "spidev" don't work in compat strings anymore and should not be suggested as alternatives.

madisongh:
> I can also update to Backport, but this is technically wrong because this is not on kernel mainline

Yeah, I know, but there isn't a good tag to cover the actual use case here.  I'm just considering it a "backport" from the NVIDIA kernel, which is sort of upstream from us, also.

asterix:
One more question - I want this patched on scarthgap too. The main branch has 6.18 specific append. Do you want me to create a generic append or remove the 6.18 versioning? May be it won't be directly mergeable into scarthgap, so may be just another PR for it? How do you handle this?

madisongh:
> May be it won't be directly mergeable into scarthgap, so may be just another PR for it? 

The added patches may need rebasing due to the different kernel versions between the two branches, so a separate PR is probably a good a idea.

For simpler situations, I usually handle the cherry-picking between branches without needing separate PRs.

asterix:
will open another PR of scarthgap, thx.

madisongh:
@asterix The update wasn't quite what I had asked for, so I've made the needed changes myself. Verified using the loopback method described in discussion #1604.

asterix:
Sounds good, but what is it that I missed here?

madisongh:
> Sounds good, but what is it that I missed here?

I asked for the patches to be ported over from the NVIDIA kernel, and maybe I wasn't clear enough on what form they should take... if you look at 1a1af90 you'll see that I preserved the original commit messages. I do that where possible to track what should/should not be carried forward in future updates.

PR #2187, opened 2026-04-08
optee: add option to build fTPM components

dolwup:
NVIDIA provides the ability to build Firmware TPM Support into OP-TEE by specifying -t to their build script in the l4t optee sources. This is a replication of what this flag does.

Adds OPTEE_ENABLE_FTPM_BUILD as a variable (default: 1) to enable/disable fTPM support. Depending on this variable build and install the ftpm-helper and ms-tpm TAs as well as the nvftpm-helper-app CA (for provisioning).

### DIFF
--- a/recipes-security/optee/optee-client/tee-ftpm-modprobe.service
+++ b/recipes-security/optee/optee-client/tee-ftpm-modprobe.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Load fTPM Module
+After=tee-supplicant.service
+BindsTo=tee-supplicant.service
+ConditionPathExists=/dev/teepriv0
+
+[Service]
+Type=oneshot
+RemainAfterExit=true
+ExecStart=modprobe --ignore-install tpm_ftpm_tee
+ExecStop=modprobe -r tpm_ftpm_tee
+
+[Install]
+WantedBy=tee-supplicant.service
--- a/recipes-security/optee/optee-client/tee-supplicant.service.in
+++ b/recipes-security/optee/optee-client/tee-supplicant.service.in
@@ -2,6 +2,7 @@
 Description=TEE Supplicant
 
 [Service]
+Type=exec
 User=root
 EnvironmentFile=-@sysconfdir@/default/tee-supplicant
 ExecStart=@sbindir@/tee-supplicant $OPTARGS
--- a/recipes-security/optee/optee-client/tee-supplicant.sh.in
+++ b/recipes-security/optee/optee-client/tee-supplicant.sh.in
@@ -6,6 +6,7 @@ DESC="OP-TEE Supplicant"
 
 DAEMON=@sbindir@/$NAME
 OPTARGS=
+FTPM_ENABLED="@OPTEE_ENABLE_FTPM@"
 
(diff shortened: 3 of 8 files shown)


dolwup:
Opened as draft because I haven't been able to retest it after rebasing yet

ichergui:
Hi @dolwup 
Thanks for your PR. Much appreciated.
Did you create an issue where you described the need of those changes ? 
Any issue at the runtime that depends on enabling fTPM ?

dolwup:
> Hi @dolwup Thanks for your PR. Much appreciated. Did you create an issue where you described the need of those changes ? Any issue at the runtime that depends on enabling fTPM ?

I didn't create an issue yet. I can do that if you want and link this pr.
We wanted to use the fTPM so I tried to set it up according to [nvidias documentation](https://docs.nvidia.com/jetson/archives/r36.4.4/DeveloperGuide/SD/Security/FirmwareTPM.html) but I noticed that support for it was missing in OE4T.
Afaik fTPM is not required for anything yet. This just adds a feature that's available on stock l4t ^^

ichergui:
Thanks for the clarifications.

madisongh:
Thanks for this, generally looks good. I'm not crazy about the new variable, though. Any chance you could use `PACKAGECONFIG` instead?

Also, I'm not sure what happens if someone who has been building a product without the fTPM support updates to a build that includes it by default. I'd rather have the default be off, rather than on, at least in the scarthgap branch, just so there's no chance of breaking someone's upgrades. We could default it to on in master and future release branches though.

dolwup:
I made some changes to make it work on 36.5. 
Nvidia updated their modprobe deny-list to also block loading the `tpm_ftpm_tee` manually like explained in their [documentation](https://docs.nvidia.com/jetson/archives/r36.5/DeveloperGuide/SD/Security/FirmwareTPM.html#running-the-ftpm-provisioning-script). I checked how they handle this on stock l4t and appearently they load it through a systemd service on startup (after tee-supplicant). I added that to the PR.


> Any chance you could use `PACKAGECONFIG` instead?

I'm not sure especially with the extra changes I added to load the ftpm kernel module on startup. With PACKAGECONFIG the user would need to override that variable for 3 different recipes.

> I'd rather have the default be off, rather than on

I changed it to be off by default. That's the default behavior for L4T anyway. 
I also turn off autoloading of the ftpm kernel module depending on `OPTEE_ENABLE_FTPM` (renamed, since it's not only affecting the optee build). NVIDIA autoloads it regardless in L4T but I don't see why we should load the module when we didn't build support for it into optee.

dolwup:
Another note: I haven't tested this without systemd yet (specifically the automatic module loading) but it _should_ work. 
I'll try to build `tegra-demo-distro`  with sysvinit and check if it works

dolwup:
I tested the changes with `tegra-demo-distro` with sysvinit instead of systemd and it works. I had to add some depends/recommends though specifically the kernel module and the `tegra-configs-udev` which provides the denylist modprobe config which blocks the kernel module from loading too early.
Should be ready now

madisongh:
Thanks!

PR #2192, opened 2026-04-21
image_types_tegra: rename tegraflash.tar to tegraflash-tar

### LINKED ISSUE #2191: tegraflash.tar fails to parse with latest oe-core
Using ".tar" suffix causes tar to be parsed as image conversion.

After:
https://git.openembedded.org/openembedded-core/commit/?id=76316c765d3183f7ecbf8e83c514c392d5a86eed

it causes parsing failure:
```
No CONVERSION_CMD defined for subtype "tar" - possibly invalid conversion type name or missing support class
```

It's discussed on ML in:
https://lists.openembedded.org/g/openembedded-core/topic/118885182#msg235513
and suggested fix it to rename the IMAGE_FSTYPE. I can send PR to do that, but first wanted to see if you agree with renaming and what naming you prefer.

CC: @kraj @bhstalel

shr-project:
Using ".tar" suffix causes tar to be parsed as image conversion.

After:
https://git.openembedded.org/openembedded-core/commit/?id=76316c765d3183f7ecbf8e83c514c392d5a86eed

it causes parsing failure:

No CONVERSION_CMD defined for subtype "tar" - possibly invalid conversion type name or missing support class

It's discussed on ML in:
https://lists.openembedded.org/g/openembedded-core/topic/118885182#msg235513

Fixes: https://github.com/OE4T/meta-tegra/issues/2191

### DIFF
--- a/classes-recipe/image_types_tegra.bbclass
+++ b/classes-recipe/image_types_tegra.bbclass
@@ -4,7 +4,7 @@ TEGRA_UEFI_SIGNING_CLASS ??= "tegra-uefi-signing"
 inherit ${TEGRA_UEFI_SIGNING_CLASS}
 TEGRA_UEFI_USE_SIGNED_FILES ??= "false"
 
-IMAGE_TYPES += "tegraflash.tar"
+IMAGE_TYPES += "tegraflash-tar"
 
 IMAGE_ROOTFS_ALIGNMENT ?= "4"
 
@@ -195,8 +195,8 @@ tegraflash_custom_post() {
 }
 
 tegraflash_finalize_pkg() {
-    rm -f ${IMGDEPLOYDIR}/${IMAGE_NAME}.tegraflash.tar
-    ${IMAGE_CMD_TAR} --sparse --numeric-owner --transform="s,^\./,," -cf ${IMGDEPLOYDIR}/${IMAGE_NAME}.tegraflash.tar .
+    rm -f ${IMGDEPLOYDIR}/${IMAGE_NAME}.tegraflash-tar
+    ${IMAGE_CMD_TAR} --sparse --numeric-owner --transform="s,^\./,," -cf ${IMGDEPLOYDIR}/${IMAGE_NAME}.tegraflash-tar .
 }
 
 tegraflash_create_flash_config() {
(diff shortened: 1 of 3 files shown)


kraj (approved):
LGTM

madisongh:
You'll need to change the output file name to end with `.tegraflash-tar` also - I don't see that in the diff - or subsequent conversions will fail.

There is a built-in assumption in the way image and conversion types are handled that the suffix on the file name exactly matches the image type or conversion type.

We can rename the output file, too, but that defeats shell auto-completion when it comes time to unpack the tarball, which is damned inconvenient.

shr-project:
> n in the way image and conversion types are handled that the suffix on the file name exactly matches the image type or conversion type.

My build is still running, so I haven't seen it failing based on the assumption it matches the filename (I thought it doesn't), that's why the PR was still in the draft state, before at least finishing the build test.

shr-project:
You're right it just failed now:
```
zstd: can't stat image-20260421143755.tegraflash-tar : No such file or directory -- ignored
```

So what is your preferred fix?

madisongh:
@shr-project We're kind of between a rock and a hard place now, I guess, so I'll take the file name change. It's a shame that the prior flexibility that allowed image types to contain `.` was accidental rather than intentional. And I guess we're the only ones that needed it.

I'm not sure what the approved way of handling our needs is supposed to be, I'm open to suggestions on a redesign of the layer to fit whatever that is, though.

shr-project:
OK, I've updated the PR to rename the file to -tar as well and it passed my minimal build test locally. Lets see if it passes your CI as well. If it works, the same will be needed in https://github.com/OE4T/meta-tegra/commits/master-l4t-r38.4.x/ and master-l4t-r38.2.x if these branches are maintained (seems they are, but currently only for Thor) - sorry I'm not very familiar with tegra MACHINEs :).

madisongh:
The longer-term solution will be, I think, to return to the way things were before 0c07ff6ac8434afc391695bd3687355f7b4436f2 went in, making `tegraflash` a terminal image type with no compression/conversion additions allowed to be appended, and having a separate, private setting for compression type of the resulting tarball.

shr-project:
Isn't it possible to use multiple CONVERSION/COMPRESSION TYPES and use CONVERSION_CMD instead of calling tar from tegraflash_finalize_pkg? Then it might be possible to support different combinations people might configure with the standard mechanism. But it's been a while since I was digging in this area.

PR #2193, opened 2026-04-21
image_types_tegra: rename tegraflash.tar to tegraflash-tar

shr-project:
Using ".tar" suffix causes tar to be parsed as image conversion.

After:
https://git.openembedded.org/openembedded-core/commit/?id=76316c765d3183f7ecbf8e83c514c392d5a86eed

it causes parsing failure:

No CONVERSION_CMD defined for subtype "tar" - possibly invalid conversion type name or missing support class

It's discussed on ML in:
https://lists.openembedded.org/g/openembedded-core/topic/118885182#msg235513

Fixes: https://github.com/OE4T/meta-tegra/issues/2191


madisongh:
Thanks, but I'll take care of cherry-picking the changes over and resolving the conflicts.

PR #2197, opened 2026-05-01
nvidia-kernel-oot display fixes for linux-yocto kernels

kekiefer:
Here are a few patches to address a couple issues with displays on linux-yocto.

The first patch backports missing support for fbdev in 6.11+ kernels.

The second two patches address some minor inconsistencies with displayport, particularly involving hotplug.


madisongh:
LGTM, and thanks for submitting that fix upstream, hopefully it will get some traction

kekiefer:
For the purposes of a cross-reference on github (since it didn't get picked up in the commit message) the upstream PR is here: https://github.com/NVIDIA/open-gpu-kernel-modules/pull/1127

PR #2199, opened 2026-05-01
Exclude paths to key/cert files from checksum calculation

mityada:
The following variables contain paths to the key/cert files used for signing:
- TEGRA_SIGNING_PKC
- TEGRA_SIGNING_SBK
- TEGRA_UEFI_DB_CERT
- TEGRA_UEFI_DB_KEY

These paths are specific to a build enviroment and should not be used in task checksum calculation.

The contents of these files are already accounted for using file-checksums. See TEGRA_SIGNING_FILECHECKSUMS and TEGRA_UEFI_SIGNING_FILECHECKSUMS.


madisongh:
Thanks!

PR #2200, opened 2026-05-02
Backport #1988 and #2199 to scarthgap

mityada:
Backport #1988 and #2199 to scarthgap


madisongh:
@mityada Could you go ahead and add your fix from #2199 to this PR as well? I expect it will be needed.

PR #2204, opened 2026-05-05
tegra-helper-scripts: add support for additional files in flash packa…

minicarpet:
This modification allows to add files to the flash package.
This can be useful for production for example when using init-extra.sh script which can uses those files.

can be merged into whinlater, wrynose and master branches.

### DIFF
--- a/docs/initrd-flashing-support.md
+++ b/docs/initrd-flashing-support.md
@@ -78,6 +78,169 @@ The script:
 
 **Note:** add the current Linux user to the `disk` group to avoid the usage of `sudo` to run `initrd-flash` script.
 
+## Production-Time Customization
+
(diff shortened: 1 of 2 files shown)


madisongh:
Could you include some information about this addition in `doc/initrd-flashing-support.md`  also? That whole document probably needs an update, but it's not abundantly clear, from what you have here, what this might be used for, or how the additional files get populated at build time.

minicarpet:
> Could you include some information about this addition in `doc/initrd-flashing-support.md` also? That whole document probably needs an update, but it's not abundantly clear, from what you have here, what this might be used for, or how the additional files get populated at build time.

Done, I tried to be as explicit as possible

PR #2206, opened 2026-05-07
linux-yocto: fix do_kernel_configcheck warning

ohporter:
linux-yocto's do_kernel_configcheck flags an m/y mismatch on CONFIG_CRC_ITU_T:

  WARNING: linux-yocto-6.18.24+git-r0 do_kernel_configcheck:
  [kernel config]: specified values did not make it into the
  kernel's final configuration:

    [NOTE]: 'CONFIG_CRC_ITU_T' last val (m) and .config val (y)
            do not match
    [INFO]: CONFIG_CRC_ITU_T : y ## .config: 11863 :
            configs/v6.18/standard/genericarm64/features/nfc/
   	    nfc-generic.cfg (m)

genericarm64.cfg sets AQUANTIA_PHY=m and nfc-generic.cfg also sets CRC_ITU_T=m causing the warning when tegra.cfg conflicts with AQUANTIA_PHY=y. Remove the conflict by matching genericarm64's module setting.


madisongh:
I would call this a workaround, rather than a fix, since it's the hard-coding of the `CONFIG_CRC_ITU_T=m` value in `nfc-generic.cfg` which is the problem here. That item and `CONFIG_CRC_CCITT` are `select`ed where needed by other drivers/subsystems. They shouldn't be manually configured that way (although those lines date back to 2017, when things may have been different).

It looks like this should be OK... as far as I can tell, the Ethernet interface that uses the Aquantia PHY also uses the `nvethernet` out-of-tree driver for the MAC anyway, so no one would be expecting support for it to be entirely built into the kernel.

PR #2208, opened 2026-05-11
cuda: add nativesdk-cuda-driver-stubs to runtime dependencies for cud…

minicarpet:
When populateing sdk with:
`TOOLCHAIN_HOST_TASK:append = " nativesdk-cuda-nvprof nativesdk-cuda-gdb "`
an error occirs during `package_qa `step with  `file-rdeps` missing dependencies.

This PR fixes these QA issues for both recipes.

**NOTE** : not sure if the sip is the best option for `cuda-gdb`

### DIFF
--- a/recipes-devtools/cuda/cuda-gdb_12.6.68-1.bb
+++ b/recipes-devtools/cuda/cuda-gdb_12.6.68-1.bb
@@ -15,4 +15,9 @@ FILES:${PN}-dev += "${prefix}/local/cuda-${CUDA_VERSION}/share/gdb"
 RDEPENDS:${PN} += "gmp"
 RDEPENDS:${PN}-dev += "python3"
 INSANE_SKIP:${PN}-dev += "staticdev"
+# cuda-gdb ships pre-built TUI binaries for multiple Python versions (3.8-3.12)
+# linked against libpython3.x, libncurses and libcrypt from the host OS.
+# There are no nativesdk providers for these system libraries — they are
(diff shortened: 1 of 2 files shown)


madisongh:
Why are you adding those tools to the SDK in the first place? SDKs are for cross-building, while the debugger and profiler are used for directly running applications... or am I misunderstanding, and somehow those tools have some kind of host/target split like regular gdb does?

Maybe a better option would be to remove `nativesdk` from the BBLCASSEXTEND setting on those two recipes?

minicarpet:
Well to be honest I am not sure they are really needed, especially nvprof...
About gdb does it not necessary to have it to debug cuda apps from the sdk ?

If not necessary, then yes, nativesdk should be removed.

ichergui:
Hi @minicarpet 
Any update ? 
Or we should close it.

minicarpet:
Hi @ichergui,
I was waiting for an answer from my last comment.

madisongh:
Removing `natviesdk` from the BBCLASSEXTEND setting is probably the best answer here. 

NVIDIA does have the source code for cuda-gdb in a Github repo, so if we were to replace the cuda-gdb recipe with one that builds from source, we'd have more control over the dependencies. That might be a big job, though.

And it looks like the cuda-nvprof package was dropped with CUDA 13, so that one becomes a non-issue anyway.

minicarpet:
Fine to me,
Closed

PR #2226, opened 2026-06-06
uefi: make lowest supported version customizable

mityada:
This can be used to allow rollback to previous L4T version, i.e. 36.5.0 -> 36.4.3


madisongh (changes requested):
I get the basic concept, I think. But rather than creating yet another bbclass, could the new `TEGRA_UEFI_xxx ?=` settings go into `conf/machine/include/tegra-common.inc`, and the version string conversion function get moved to a module under `meta-tegra/lib`?

I would also like to see the relevant page(s) in `docs/` get updated to add in these new variables and how they would be used.

mityada:
> I get the basic concept, I think. But rather than creating yet another bbclass, could the new `TEGRA_UEFI_xxx ?=` settings go into `conf/machine/include/tegra-common.inc`, and the version string conversion function get moved to a module under `meta-tegra/lib`?
> 
> I would also like to see the relevant page(s) in `docs/` get updated to add in these new variables and how they would be used.

I moved the function and the variables as you requested.

As for the documentation, I failed to find any relevant pages to add new variables to. It probably should have been https://github.com/OE4T/meta-tegra/blob/master/docs/Over-the-air-reflashing-process.md, but it seems hopelessly outdated.

madisongh:
> I moved the function and the variables as you requested.

Thanks, and hopefully it still works as you need it to this way.

> As for the documentation, I failed to find any relevant pages to add new variables to. It probably should have been https://github.com/OE4T/meta-tegra/blob/master/docs/Over-the-air-reflashing-process.md, but it seems hopelessly outdated.

Yep, there's  a bunch of catch-up work still to do... having the documentation in the wiki led to it getting stale more quickly.f

If you could add a section to `docs/Secure-Boot-Support-in-L4T-R35.2.1-and-later.md` to mention how rollback protection works and how you would set the variables to control downgrades, even if it doesn't flow well with the rest of the somewhat stale other sections, I'd appreciate it.

mityada:
> If you could add a section to `docs/Secure-Boot-Support-in-L4T-R35.2.1-and-later.md` to mention how rollback protection works and how you would set the variables to control downgrades, even if it doesn't flow well with the rest of the somewhat stale other sections, I'd appreciate it.

These variables are not technically related to Secure Boot, so I added a new page for UEFI Capsule Updates.

madisongh:
Thanks!

PR #2227, opened 2026-06-08
conf/layer.conf: update LAYERSERIESCOMPAT with wrynose

pberginkonsult:
When Nvidia is releasing Jetpack 39.2.0 with Yocto support quite
close in time with the release of Yocto 6.0, that is a LTS release,
I think it will be quite common to use them together. Let us aim for
compatibility of the two until we need to branch them out.

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -12,7 +12,7 @@ LAYERVERSION_tegra = "1"
 
 LAYERDEPENDS_tegra = "core"
 
-LAYERSERIES_COMPAT_tegra = "blacksail"
+LAYERSERIES_COMPAT_tegra = "wrynose blacksail"
 
 # Recipe modifications for other layers that may be included in the build
 BBFILES += "${@' '.join('${LAYERDIR}/external/%s/recipes*/*/*.bb' % layer \


maximdeclercq:
Normally this branch is meant to merge up into master, which carries the full Orin support. Adding wrynose makes sense, but bumping LAYERSERIESCOMPAT here diverges from master and could break the merge. It would be cleaner to have the changes merged onto master first, and then have the OE4T maintainers cherry-pick it onto the existing wrynose branch.

madisongh:
I'd prefer not to make this change, since the focus is just on master, and I do no test builds against wrynose, so I don't want to advertise the compatibility that this change implies. Once we merge into master, we will back-port to wrynose. 

But yes, the two bases are pretty close; for now you can always override the LAYERSERIESCOMPAT setting in your local build setup to work around this, if you need to.

pberginkonsult:
No worries. Then I keep the patch locally. I will initially work against the released wrynose release. You are free to close this and the other one in meta-tegra-community.

My intention was to ease the work and instead of backporting we develop this as both wrynose and blacksail and split them in later stage.

madisongh:
> My intention was to ease the work and instead of backporting we develop this as both wrynose and blacksail and split them in later stage.

Yep, understood. I just didn't want to confuse less-experienced folks that might see breakage using the untested combination. And folks with experience know how to coerce the configuration they need...

madisongh:
Will target wrynose later.

PR #2228, opened 2026-06-08
Emit missing Thor flash variables

maximdeclercq:
The T264 flash helper reads `$DTB_FILE`, `$BCTFILE`, and `$BPMP_MEM_CONFIG` from the generated `flashvars` file, but none are listed in `TEGRA_FLASHVARS`, so `tegra-flashvars` never emits them and the helper receives empty values.


conf/machine/include/tegra264.inc
```diff
@@ -26,6 +26,7 @@ TEGRA_CUDA_ARCHITECTURE ?= "110"
 
 TEGRA_FLASHVAR_BR_CMD_CONFIG ?= ""
 TEGRA_FLASHVAR_FSIFW ?= ""
+TEGRA_FLASHVARS:append = " BCTFILE BPMP_MEM_CONFIG DTB_FILE"
```
madisongh: You can just add these entries to the setting in `tegra-common.inc`, rather than appending here.

madisongh (changes requested):
Thanks, please see in-line

pberginkonsult:
Please add a better git commit message to understand why this change is needed (that you have written in a comment in Github). In my opinion git commits shall have enough info to understand the change and just additional info about really detailed stuff shall go in to github comments.

maximdeclercq:
@madisongh @pberginkonsult Thanks for the feedback. Changes should be implemented.

madisongh (approved):

PR #2232, opened 2026-06-08
[wrynose] uefi: make lowest supported version customizable

mityada:
Cherry-pick #2226 to wrynose


madisongh:
I took care of this manually, and will do the same for the scarthgap branch.

PR #2235, opened 2026-06-08
L4T R39.2.0: CUDA 13.2 build fixes (nvcc -Xcompiler handling, cub proclaims specializations)

johannesschrimpf:
When compiling onnxruntime with the wip-l4t-r39.2.0 branch, I got two different compilation errors regarding cuda. Both patches were drafted with AI assistance, reviewed and tested by me.

The cuda-cccl patch, I can easily also add in my own meta layer as a bbappend, so if you'd rather not take that one, feel free to drop it. As far as I understand, the cuda.bbclass change is not as easy to fix downstream.

**1) Error message due to dangling -Xcompiler (cuda.bbclass)**

```
aarch64-poky-linux-g++: error: unrecognized command-line option '-Xcompiler'
```

The problem seems to be that the OE-core distro defaults inject `-Werror=format-security`, which is then forwarded by cuda.bbclass as `-Xcompiler -Werror=format-security`. When cmake probes the CUDA compiler it strips `-Werror` parameters, leaving `-Xcompiler` without an argument, which then fails.

Apparently `-Xcompiler` with a space and with `=` are treated the same, so adding the `=` prevents the parameter from being stripped and builds cleanly.

**2) namespace-scope problems (cuda-cccl)**

```
| .../work/aarch64_tegra234-poky-linux/onnxruntime/1.26.0/recipe-sysroot/usr/local/cuda-13.2/include/cccl/cub/device/device_transform.cuh:44:111: error: global qualification of class name is invalid before ':' token
|    44 | struct ::cuda::proclaims_copyable_arguments<CUB_NS_QUALIFIER::detail::__return_constant<T>> : ::cuda::std::true_type
|       |                                                                                                               ^
| .../work/aarch64_tegra234-poky-linux/onnxruntime/1.26.0/recipe-sysroot/usr/local/cuda-13.2/include/cccl/cub/device/device_transform.cuh:44:111: error: expected '{' before ':' token
```

The problem seems to be that cub/device/device_transform.cuh and dispatch/tuning/tuning_transform.cuh declare two
cuda::proclaims_copyable_arguments specializations with a leading global qualifier (struct ::cuda::...<...> : base), which gcc does not accept here.

Re-declaring them inside namespace cuda { ... } builds cleanly. As far as I can tell it is behaviour-preserving: same
specialization of the same template for the same types, just declared from namespace cuda instead of with a leading global qualifier.

### DIFF
--- a/classes-recipe/cuda.bbclass
+++ b/classes-recipe/cuda.bbclass
@@ -16,7 +16,7 @@ LDFLAGS:prepend:cuda = "${TOOLCHAIN_OPTIONS} "
 LDFLAGS:append:cuda = " ${CUDA_LDFLAGS}"
 DEBUG_PREFIX_MAP:remove:cuda = "-fcanon-prefix-map"
 
-def cuda_extract_compiler(compiler, d, prefix='-Xcompiler '):
+def cuda_extract_compiler(compiler, d, prefix='-Xcompiler='):
     args = d.getVar(compiler).split()
     if args[0] == "ccache":
         return args[1], ' '.join([prefix + arg for arg in args[2:]])
--- a/recipes-devtools/cuda/cuda-cccl/0001-Fix-invalid-C-syntax-in-proclaims_copyable_arguments.patch
+++ b/recipes-devtools/cuda/cuda-cccl/0001-Fix-invalid-C-syntax-in-proclaims_copyable_arguments.patch
@@ -1,0 +1,50 @@
+Fix invalid C++ syntax in proclaims_copyable_arguments specializations
+
+cccl/CUB declares its two cuda::proclaims_copyable_arguments specializations in
+cub/device/device_transform.cuh and tuning_transform.cuh with a leading global
+qualifier (struct ::cuda::...<...> : base). gcc rejects this in some translation
+units (e.g. onnxruntime CUDA kernels pulling in cub::DeviceTransform via cutlass
++ libcu++) with:
+
+  error: global qualification of class name is invalid before ':' token
+  error: expected '{' before ':' token
+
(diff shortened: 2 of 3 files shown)


ichergui (approved):

ichergui:
@johannesschrimpf  Thanks for your PR. I will share your changes to the CCCL maintainers so hopefully it will be upstreamed.

madisongh:
Is that CCCL patch the same as https://github.com/NVIDIA/cccl/pull/8771 ?

johannesschrimpf:
> Is that CCCL patch the same as [NVIDIA/cccl#8771](https://github.com/NVIDIA/cccl/pull/8771) ?

Yes, looks like it. Good find. So I guess we rather take the upstream patch in directly and mark it as submitted?

ichergui:
> > Is that CCCL patch the same as [NVIDIA/cccl#8771](https://github.com/NVIDIA/cccl/pull/8771) ?
> 
> Yes, looks like it. Good find. So I guess we rather take the upstream patch in directly and mark it as submitted?

Yes Please refresh your patch

johannesschrimpf:
> > > Is that CCCL patch the same as [NVIDIA/cccl#8771](https://github.com/NVIDIA/cccl/pull/8771) ?
> > 
> > 
> > Yes, looks like it. Good find. So I guess we rather take the upstream patch in directly and mark it as submitted?
> 
> Yes Please refresh your patch

I now backported the changes from the original PR, added the reference, and updated the commit message. Changes were different formatting. Could you have a look if this looks ok?

ichergui (approved):
LGTM

PR #2238, opened 2026-06-09
perf,linux-noble-nvidia-tegra: fix build failure for perf

pberginkonsult:
Fix to be able to build perf with Nvidia kernel.
The NVIDIA 6.8 kernel cherry-picked b0ae641a9e1e ("perf python: Avoid duplicated code in get_tracepoint_field") which replaces direct struct field access with a call to evsel__tp_format(). However, the function definition was introduced by a later commit that was not included in this kernel, causing a build failure:

  tools/perf/util/python.c:528: error: implicit declaration of function
  'evsel__tp_format'

Revert the python.c change from b0ae641a9e1e, falling back to the existing trace_event__tp_format_id() with lazy-loading guard that was present before the cherry-pick. The PyUnicode_AsUTF8 improvement from 60f55efa822c is preserved.

### DIFF
--- a/recipes-kernel/linux/linux-noble-nvidia-tegra-6.8/0002-Revert-perf-python-Avoid-duplicated-code-in-get-trac.patch
+++ b/recipes-kernel/linux/linux-noble-nvidia-tegra-6.8/0002-Revert-perf-python-Avoid-duplicated-code-in-get-trac.patch
@@ -1,0 +1,65 @@
+From 0000000000000000000000000000000000000000 Mon Sep 17 00:00:00 2001
+From: Peter Bergin <peter@berginkonsult.se>
+Date: Sun, 08 Jun 2026 00:00:00 +0000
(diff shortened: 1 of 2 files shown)


pberginkonsult:
@ichergui @kekiefer can you please direct this to your kernel team?

Looks like around 70 patches are backported to `tools/perf` in the new kernel version for 39.2.0. But building perf failed. Tried to bisect and find culprint and ended up in this revert patch. Good if the person(s) implementing this could take a look.
https://gitlab.com/nvidia/nv-tegra/3rdparty/canonical/linux-noble/-/commit/b0ae641a9e1e5e912cb371005e9f99a07f3a7444
The commit reference a "Bug 5985551" that I'm concerned about.

kekiefer:
Thanks @pberginkonsult for the change. I'll reach out to the kernel team and see if I can shed a little more light on this.

kekiefer:
So the perf series was pulled in and tested on the noble kernel, but not against the python bindings, so this detail was missed. This requirement has now been identified and it's being tracked for the next releases.

PR #2239, opened 2026-06-09
L4T r39.2.0 + linux yocto (t234)

kekiefer:
A handful of changes to make linux-yocto compile, build, and boot cleanly for Orin. Upstream is not complete with Thor at 6.18, and the gap is as of yet unexplored, so I figured it would be best to just get this going in the meantime.

One bit of re-imagining I did here is to rework the DTB build. Rather than try to finagle L4T makefiles to get things working with linux-yocto, I decided to build the DTBs with the devicetree.bbclass. The include ordering is still critical, so there's still some special handling for that, but at least it's a little more Bitbakothic(?) now.


ichergui (approved):
Thanks @kekiefer. LGTM

PR #2241, opened 2026-06-09
nvidia-kernel-oot: fix loading ordering between nvidia-drm and nvidia…

ichergui:
…-uvm


kekiefer:
@ichergui can you expand on the need for this just a little more in the commit message? The change looks good (in line what we've had to do in the past), and I understand the need, but let's make sure it's clear to other people looking back at the change.

ichergui:
> @ichergui can you expand on the need for this just a little more in the commit message? The change looks good (in line what we've had to do in the past), and I understand the need, but let's make sure it's clear to other people looking back at the change.

CC: @aparna-nvd

Done :heavy_check_mark:

kekiefer (approved):

kekiefer (changes requested):
After looking closer at this, I'm convinced that this is working around a bigger packaging problem in the oot modules. I want to put in a pause on this change and get to actual root cause here.

ichergui:
Closing this PR. I will submit new one.

PR #2244, opened 2026-06-10
nvidia-kernel-oot: fix build nvidia display driver for tegra264 vs te…

ichergui:
…gra234

- the module target should be nvidia-gpu-display for Thor but nvidia-display for Orin


kekiefer (approved):
LGTM. Working now on T4000 with noble-rt (headless) and AGX Orin with linux-yocto and weston.

PR #2249, opened 2026-06-12
machine: add APE audio graph card support for devkit boards

kekiefer:
Add kernel modules for the Tegra APE (Audio Processing Engine) sound card to boards whose carriers include an on-board RT5640 codec (P3737 for AGX Orin variants, P4071 for AGX Thor variants).

The common AHUB/APE module list, codec drivers, and aconnect bus driver are collected in a new devkit-audio.inc, following the same pattern as devkit-wifi.inc. Boards on the P3768/P3509 carrier (Orin NX/Nano) have no on-board codec and do not include devkit-audio.inc.

The stale PREFERRED_RPROVIDER entries for OOT sound packages that are no longer built by the nvidia-kernel-oot recipe are also removed from tegra-common.inc


madisongh:
Thanks @kekiefer ... do we need to update the TEGRA_AUDIO_DEVICE setting for the dev kits, too, so `recipes-bsp/tegra-binaries/tegra-configs-alsa_39.2.0.bb` installs the ALSA config for the AHUB/APE setup instead of the HDMI configuration?

kekiefer:
> Thanks @kekiefer ... do we need to update the TEGRA_AUDIO_DEVICE setting for the dev kits, too, so `recipes-bsp/tegra-binaries/tegra-configs-alsa_39.2.0.bb` installs the ALSA config for the AHUB/APE setup instead of the HDMI configuration?

I don't think so. The default devkit configurations don't include jacks for APE audio, instead exposing signals on the 40-pin header, so leaving the HDMI audio out there makes sense. If someone has the pins wired up they can change this if they need, or just handle directing to the APE audio device in their application.

PR #2250, opened 2026-06-12
r39.2.0 nv oot dtb updates

kekiefer:
Two changes to clean up the oot dtb build a little more. The first to go back to totally nailing down the include paths for consistency. The second to break the out-of-tree dtb concerns into a bbclass so we can pick it up e.g. in tegrademo oot dtb examples. I'll send an update for the tegrademo devicetrees subsequently.


kekiefer:
Let me know if you'd rather see the bbclass as just an include file. I like the bbclass because then consumers don't have to worry about paths. But the downside is then some of the logic is a bit off to the side.

classes/nv-oot-dtb.bbclass
```diff
@@ -0,0 +1,64 @@
+TEGRA_UEFI_SIGNING_CLASS ??= "tegra-uefi-signing"
+
+inherit devicetree
+inherit ${TEGRA_UEFI_SIGNING_CLASS}
+
+DEPENDS += "nvidia-kernel-oot"
+
+# Overlay fragments combined into single DTBs contain some duplicate
+# labels. DTC 1.7.0 treats this as a fatal error by default.
+DTC_FLAGS:append = " -E no-duplicate_label"
```
madisongh: If you move this to before the `inherit devicetree`, you could make this `DTC_FLAGS ?= "..."`, to override the `?=` definition in that class.  For bbclass and base recipe implementations, it's a good idea to avoid use of `:append` and `:prepend` where possible, since those operations are impossible to override.

It would also be OK to keep this here but switch to using the `+=`  operator.

classes/nv-oot-dtb.bbclass
```diff
@@ -0,0 +1,64 @@
...
+
+# From nvidia-kernel-oot build/nvidia-public/devicetree/Makefile.generic
+DTC_PPFLAGS:append = " -DLINUX_VERSION=600 -DTEGRA_HOST1X_DT_VERSION=2 -DOS_LINUX"
```
madisongh: Same comment here as for line 10

classes/nv-oot-dtb.bbclass
```diff
@@ -0,0 +1,64 @@
...
+
+DT_NV_BASE = "${STAGING_DIR_HOST}/usr/src/device-tree/nvidia"
+
+# expand_includes() uses a Python set(), randomizing -I order. nv-public/ and
+# nv-platform/ have same-named dtsi files; nv-public/ versions define the mmc
+# aliases needed for correct block device enumeration. List paths explicitly in
+# DTC_PPFLAGS (before DT_INCLUDE) without nv-platform/, which resolves via
+# gcc's current-file-directory search per Makefile.generic.
+DTC_PPFLAGS:prepend:tegra234 = " \
```
madisongh: Is there a reason you didn't just copy over the replacement implementation for the `expand_includes` function that you did for the recipe in tegra-demo-distro?  I think that's the cleaner approach, and we could try upstreaming that change to devicetree.bbclass in OE-Core, since its use of `set()` is probably not very good for build reproducibility, either.

classes/nv-oot-dtb.bbclass
```diff
@@ -0,0 +1,64 @@
...
+        tegra_uefi_attach_sign "$dtbf"
+    done
+}
+do_sign_dtbs[dirs] = "${B}"
+do_sign_dtbs[depends] += "${TEGRA_UEFI_SIGNING_TASKDEPS}"
+do_sign_dtbs[file-checksums] += "${TEGRA_UEFI_SIGNING_FILECHECKSUMS}"
+
+addtask sign_dtbs after do_compile before do_install
+
+do_install:append() {
```
madisongh: This would be better implemented as
```
do_install() {
    devicetree_do_install
    ... extra commands here...
}
```
rather than using `:append`, to make it easier for users to change the functionality in their bbappends.

classes/nv-oot-dtb.bbclass
```diff
@@ -0,0 +1,64 @@
...
+    for dtbf in ${B}/*.dtb.signed ${B}/*.dtbo.signed; do
+        [ -e "$dtbf" ] || continue
+        install -m 0644 "$dtbf" "${D}/boot/devicetree/"
+    done
+}
+
+FILES:${PN} += "/boot/devicetree/*.dtb.signed /boot/devicetree/*.dtbo.signed"
+
+do_deploy:append() {
```
madisongh: Similar comment here as for the install step above

recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot-dtb_39.2.0.bb
madisongh, line 19: Same comment here as for the do_install step below, and even though this is a python function, it should work similarly  - implement this as `do_compile()` and call `devicetree_do_compile`

madisongh (changes requested):
Thanks for this.... I think this is appropriate for implementation as a bbclass, since it's essentially a drop-in replacement for `devicetree.bbclass` but specific to NVIDIA targets. Might be better to spell out `devicetree` in the file name, though. Maybe even make it `devicetree-tegra.bbclass`  or `tegra-devicetree.bbclass`?

See in-line for some recommendations for implementation changes.

classes/nv-oot-dtb.bbclass
kekiefer: I chose this approach specifically because it avoids the reimplementation of that function. I grew tired of carrying it around. But to your point, that’s a reason to move the change upstream. Thanks for the feedback I’ll revert this to old ways.

kekiefer: Oh and I’ll try to move the update upstream

recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot-dtb_39.2.0.bb
kekiefer: This isn't part of the bbclass, so the same arguments don't hold for appendages on the default tasks. Here we're just having to handle the building of overlays that weren't in the same DT_FILES_PATH, which is a special addition for this recipe to handle the NV_OVERLAY_PATH.

madisongh: OK, if users aren't going to add a bbappend for this recipe that might need to do something differently than what's implemented here, then that's acceptable. The no-append/prepend guideline is mainly to accommodate bbappends from our users.

classes/nv-oot-dtb.bbclass
kekiefer: I've opted to use += here, in order to preserve the default functionality set forth by devicetree.bbclass, rather than to have that all copied here. It adds some padding and reserved entries that are tunable for individual use cases, and I added some entries in the dtb build to zero these out, with a comment that this isn't essential and merely done to match the L4T makefile.

classes/nv-oot-dtb.bbclass
kekiefer: Same rationale as above. The defaults are sane, we just need more.

classes/nv-oot-dtb.bbclass
kekiefer: Done

classes/nv-oot-dtb.bbclass
kekiefer: Done

classes/nv-oot-dtb.bbclass
kekiefer: Done

kekiefer:
Thanks for the feedback. I thought tegra-devicetree sounded nice, so I went with that. I've made most of the changes but left one of your comments open because I think it was just misidentified as being part of the bbclass implementation.

recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot-dtb_39.2.0.bb
kekiefer: Ah, ok. Indeed the intention with the bbclass is that other devicetree recipes will be able to just pick up the class and do their own thing with their own structure, which is the way it is now with the example in tegrademo. The needs of this recipe are specific to the structure of the oot modules sources.

madisongh (approved):

PR #2256, opened 2026-06-15
Flashing updates

### LINKED ISSUE #2248: Update documentation pages for JP 7.2

### LINKED ISSUE #2253: Re-unite flashing-related scripts
The t264/Thor firmware assembly process is different from t234/Orin, and in R38.x NVIDIA also introduced a new "unified" flashing process to be used for Thor.

When adapting R38.x for the layer, it wasn't clear whether flash scripting changes in L4T were due to the new Thor firmware, or due to the new "unified" flashing support. Plus, it wasn't clear if NVIDIA planned to move Orin to the new "unified" base.

This meant that there was no good way to retain Orin support in the flashing-related functions and scripts in the layer when adapting to R38.x.

Now that Orin is back in R39, even though NVIDIA (unsurprisingly) just tacked the old Orin flashing process back into the mix, we should not have just hacked back in the old Orin support into the layer, with a big pile of `:tegra234` / `:tegra264` overrides, plus separate flash helpers. We should try to go back to having single functions and scripts that support both platforms/ways of doing things as much as we can.

madisongh:
Not quite ready for merging, pending further documentation updates, but the scripting and build side changes should be done.

Fixes #2253


ichergui:
Thanks so much @madisongh.

ichergui:
@madisongh shouldn't we drop the `doflash`  scripts ? 
I think we should support only the `initrd-flash` process. Correct me if I'm wrong ?

madisongh:
> @madisongh shouldn't we drop the doflash scripts ?
I think we should support only the initrd-flash process. Correct me if I'm wrong ?

There should be no `doflash` (please point to where it is created, if I missed something). I kept the `dosdcard` script and the `doexternal` script for writing just the SDcard-based (for Orin Nano dev kits) or NVMe/USB-based storage directly attached to the host system. The latter probably doesn't happen very often, but writing SDcards that way is definitely faster, and less error-prone, than initrd-flash.

recipes-bsp/tegra-binaries/tegra-helper-scripts/initrd-flash.sh
```diff
@@ -230,17 +225,15 @@ sign_binaries_t234() {
...
+        # apply its sed rewrites to the the xxxFILE tokens.
+        . ./boardvars.sh
+        if MACHINE=$MACHINE BOARDID=$BOARDID FAB=$FAB BOARDSKU=$BOARDSKU BOARDREV=$BOARDREV CHIPREV=$CHIPREV CHIP_SKU=$CHIP_SKU \
                             "$here/$FLASH_HELPER" --no-flash --sign --external-device -u "$keyfile" -v "$sbk_keyfile" $instance_args \
-                            external-flash.xml.in $DTBFILE $EMC_BCT $ODMDATA $LNXFILE $ROOTFS_IMAGE; then
-                mv secureflash.xml external-secureflash.xml
-            else
-                return 1
-            fi
+                            external-flash.xml.in $LNXFILE $ROOTFS_IMAGE; then
```
markusta1, line 233: @madisongh could we add --datafile $DATAFILE here?
Otherwise flashing ORIN fails for me with a custom data img.

madisongh: Thanks for pointing this out.  Should be fixed since #2266 went in.

PR #2257, opened 2026-06-16
machine: add missing tegra210-adma to devkit-audio.inc

kekiefer:
The APE audio commit omitted the tegra210-adma DMA controller module, which was overlooked since other components are ASoC drivers. Without it, the admaif driver fails at open time with ENODEV, since there's nothing to open the dma channel for.

Add the omitted component to the devkit-audio MACHINE_EXTRA_RRECOMMENDS list.


ichergui (approved):
LGTM. Thanks @kekiefer

PR #2258, opened 2026-06-16
python3-jetson-io: add recipe for Jetson expansion header configurati…

ichergui:
…on tool

- Stage the DTB and DTBO files into /usr/share/jetson-io at image build time.
- Patch board.py to use that directory at runtime, fix block device detection on standard Linux, and search DTBOs alongside DTBs.


ichergui:
* Add new recipe for Jetson IO for sensor/camera configuration. 
  * Kindly refer to the documentation below:
  * https://docs.nvidia.com/jetson/archives/r39.2/DeveloperGuide/HR/ConfiguringTheJetsonExpansionHeaders.html#running-jetson-io

kekiefer:
I think we should work on breaking up and reworking the patches so they can be upstreamed, which will satisfy the need for maintainability of recipes here. First, there's two distinct changes that should be split apart. Then, they can be reworked in a way that we can at least submit them for comment they can be merged or reworked, and dropped from the patch list here in the future.

kekiefer (changes requested):
Need to rework the patches for upstreaming

madisongh:
I'd need to look over the code again, but based on my memory of it, and the patch you're applying here, the tool is pretty dependent on some specific configuration/layout features of Jetson Linux/Ubuntu. Coaxing it, by patches applied here, to support any custom OS setup a user might dream up, would be difficult, if not impossible. Besides, what use is this for a Yocto-style environment?
* You have to run the tool on the target.
* I don't think it spits out a dts file that you can then copy back into your build environment for reproducing the overlay across all your builds, does it?

I've never added a recipe for jetson-io to the layer for the above reasons. Now I can see folks migrating from Jetson Linux might think that they would need this tool, at least for a while.  I'm not sure I buy it, though.

I guess I'll need some convincing that there's a compelling reason to add this recipe at all. If I am convinced, the addition needs to come with a documentation page that talks about how the OS environment needs to be configured to make it work, limitations, etc.

kekiefer:
It works by modifying the overlays applied in extlinux. It will add a new entry to the file to have extlinux load an overlay, and if needed, creates an overlay dynamically to do the configuration (if the intended configuration isn’t a pre-canned one). For development, I can see this as being pretty handy, the obvious limitation being that it’s specific to the devkits. The thing that was done here, that didn’t make sense in retrospect, was placing the canned overlays in datadir, because the tool points extlinux there. It happens to work because the rootfs is all one partition, but in a case where it’s not it wouldn’t work as worst, and at best it’s a little messy. 

I’m working with Ilies on a modification here that doesn’t involve patching jetson-io and instead leverages the existing extlinux recipes in a way that doesnt preclude the use of this tool, but doesn’t change anything if it’s not included.

ichergui:
> I'd need to look over the code again, but based on my memory of it, and the patch you're applying here, the tool is pretty dependent on some specific configuration/layout features of Jetson Linux/Ubuntu. Coaxing it, by patches applied here, to support any custom OS setup a user might dream up, would be difficult, if not impossible. Besides, what use is this for a Yocto-style environment?
> 
> * You have to run the tool on the target.
> * I don't think it spits out a dts file that you can then copy back into your build environment for reproducing the overlay across all your builds, does it?
> 
> I've never added a recipe for jetson-io to the layer for the above reasons. Now I can see folks migrating from Jetson Linux might think that they would need this tool, at least for a while. I'm not sure I buy it, though.
> 
> I guess I'll need some convincing that there's a compelling reason to add this recipe at all. If I am convinced, the addition needs to come with a documentation page that talks about how the OS environment needs to be configured to make it work, limitations, etc.

Thanks @madisongh for the feedback. As @kekiefer mentioned we are working on improving this implementation. Definitely the docs should be updated.

dwalkes:
> Definitely the docs should be updated.

@ichergui I haven't tried this yet, it seems like there should probably be some updates to https://github.com/OE4T/meta-tegra/blob/wip-l4t-r39.2.0/docs/extlinux.conf-support.md - we should probably mention any support for jetson-io-tool here.

I just realized there are references to DTBFILE in this documentation page which are likely broken now and should be replaced with something similar to https://github.com/OE4T/tegra-demo-distro/commit/92e72cab9ee6f6bfcdbcc3e0b5cc139bb2e67224

kekiefer:
I’ll take a look at updating those today

ichergui:
@dwalkes I just pushed updates. Please check 
Thanks @kekiefer

docs/extlinux.conf-support.md
```diff
@@ -1,46 +1,100 @@
...
 
-`efivar -p --name 781e084c-a330-417c-b678-38e696380cb9-L4TDefaultBootMode` should return a value of `1` when using this feature. For additional context see [this thread](https://matrix.to/#/!YBfWVpJwNVtkmqVCPS:gitter.im/$x-m4h9rIYnwtMOaYtEkHg0a5HFzM4-mcpjoALOGkP4Y?via=gitter.im&via=matrix.org&via=3dvisionlabs.com) in element.
+When set, adds an `FDT` directive to `extlinux.conf` pointing to the specified device tree file. The file is installed to `/boot/dtb/` on the rootfs.
+
+This is useful for testing devicetree changes in the kernel and for supporting devicetree transitions on slot switch without a capsule update. When not set, the devicetree for the kernel is taken from the `kernel-dtb` partition, which is only updated via tegraflash or capsule update.
+
+Set it to the DTB filename (no path prefix):
+
+```
+UBOOT_EXTLINUX_FDT = "tegra234-p3701-0000-p3737-0000.dtb"
```
dwalkes: Should we use 
```
"${@os.path.basename(d.getVar('KERNEL_DEVICETREE').split()[0])}"
```
for this so it's not platform specific and uses the default dtb?  I'm wondering if we should keep this logic in a shared class so we can continue to reference it as we did before.

docs/extlinux.conf-support.md
```diff
@@ -1,46 +1,100 @@
...
 
-To use, specify
+`UBOOT_EXTLINUX_FDT` must be set for overlays to be applied. L4TLauncher only processes the `OVERLAYS` directive when it has loaded a DTB from the rootfs via the `FDT` directive. Without `FDT`, overlays are silently ignored.
 
 ```
-UBOOT_EXTLINUX_FDT = "${DTBFILE}"
+UBOOT_EXTLINUX_FDT = "tegra234-p3701-0000-p3737-0000.dtb"
 UBOOT_EXTLINUX_FDTOVERLAYS = "my-overlay.dtbo"
 ```
-Where `"my-overlay.dtbo"` is an overlay built using the mechanisms specific to your branch implementation (or potentially one provided by NVIDIA.  See [Using-device-tree-overlays](Using-device-tree-overlays.md) for more details.  Note that since the overlay only happens to the kernel DTB this mechanism cannot be used to make any changes to the UEFI DTB.
```
dwalkes, line 28: I think we still need a description of what my-overlay.dtbo is?  Is this UBOOT_EXTLINUX_FDTOVERLAYS what jetson-io is writing?  If so a reference to that here would probably be good.

kekiefer: You're right.. the details of how to get that do belong in using-device-tree-overlays and I shouldn't have dropped that reference. It's not jetson-io, which is purely for runtime configuration.

docs/extlinux.conf-support.md
kekiefer: This seems like a good default. This recipe does not specify UBOOT_EXTLINUX_FDT by default, and in minimizing change, that behavior was not altered. It might, however, be less surprising to do it the way you suggest as a default. This documentation section should then just explain the customization of this variable should that default not be desirable.

kekiefer:
@dwalkes the two points you've raised have been addressed with the two new commits here

kekiefer (approved):

dwalkes (approved):

PR #2259, opened 2026-06-18
tegraflash: support spi-only doflash packages without rootfs

anischali:
This PR updates the tegraflash package generation flow to handle targets where no rootfs image is present.

Instead of failing when `${IMAGE_TEGRAFLASH_ROOTFS}` is missing, the generated `doflash.sh` script now warns the user and forces `--spi-only`. This keeps `doflash.sh` usable for SPI-only flashing targets while still making it clear that `initrd-flash` remains the recommended flashing path for normal use cases.

Feedback and review would be appreciated, especially on whether this is the right place to handle the no-rootfs case, and whether we should introduce an explicit variable to request SPI-only flashing instead of inferring it from the missing rootfs image.

If this PR is accepted, it may also be useful to propagate or backport this change to other maintained branches where the same tegraflash rootfs handling logic exists.

### DIFF
--- a/classes-recipe/image_types_tegra.bbclass
+++ b/classes-recipe/image_types_tegra.bbclass
@@ -342,6 +342,7 @@ tegraflash_populate_package() {
 create_tegraflash_pkg() {
     local oldwd="$PWD"
     local has_sdcard="no"
+    local has_rootfs="yes"
 
     rm -rf ${WORKDIR}/tegraflash
     mkdir -p ${WORKDIR}/tegraflash
@@ -354,7 +355,15 @@ create_tegraflash_pkg() {
         cp "${IMAGE_TEGRAFLASH_INITRD_FLASHER}" ./initrd-flash.img
     fi
     tegraflash_custom_pre
-    cp "${IMAGE_TEGRAFLASH_ROOTFS}" ./${IMAGE_BASENAME}.${IMAGE_TEGRAFLASH_FS_TYPE}
+    if [ -f "${IMAGE_TEGRAFLASH_ROOTFS}" ]; then
+        cp "${IMAGE_TEGRAFLASH_ROOTFS}" ./${IMAGE_BASENAME}.${IMAGE_TEGRAFLASH_FS_TYPE}
+    else
+        bbwarn "No rootfs detected, Are you sure you know what you are doing?
+            We recommend using initrd-flash for flashing, or only use doflash.sh
+            for spi only flashing. Forcing --spi-only for this target"
+        
+        has_rootfs="no"
+    fi
     tegraflash_create_flash_config flash.xml.in ${LNXFILE}
     if grep -Eq '^[[:space:]]+<device type="sdcard"' flash.xml.in; then
         has_sdcard="yes"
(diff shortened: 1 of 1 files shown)


ichergui:
Hi @anischali 
Thanks for the PR. Please check our WIP branch `wip-l4t-r39.2.0` and see if those changes are needed. 
We are planing to merge the WIP branch to master.

anischali:
> Hi @anischali Thanks for the PR. Please check our WIP branch `wip-l4t-r39.2.0` and see if those changes are needed. We are planing to merge the WIP branch to master.


Hi @ichergui,
Looking at [line 448](https://github.com/OE4T/meta-tegra/blob/30879170eed1bfdf025aa9d127624c4f66bbfbd5/classes-recipe/image_types_tegra.bbclass#L448) of `image_types_tegra.bbclass` on the `wip-l4t-r39.2.0` branch, I believe this change is still needed — at minimum, we should check whether the rootfs file exists before copying it, rather than letting the build fail automatically.

Regarding the `doflash` part of this PR: it likely needs more work, since `doflash` no longer exists. One option would be to add SPI-only flashing support within the `doexternal` section. I'd really like to see SPI-only flashing supported, as it would save users from having to implement that logic themselves.

madisongh:
I wanted to think about this some more before commenting, but  @ichergui has forced the issue, and I don't want anyone to waste their time going down the wrong path.

There are two issues here somewhat mixed together.

First is about the flashing mechanism itself. Going forward, you should be using `initrd-flash` for all flashing -- that's the direction NVIDIA has been advising for a while now. And that's why the `doflash` script creation got removed. I did add a `--qspi-only`  option flag to `initrd-flash` to do just the QSPI programming, but it probably needs some testing on Orin, I'm not sure that's been covered as yet.

The second, and more interesting question is around how to have a recipe that generates essentially a tegraflash package with _just_ the boot firmware in it, for programming the QSPI flash.  Since tegraflash is implemented as an image type, it's expecting to be processing a rootfs image. We do have the capsule recipe that calls on some functions in `image_types_tegra.bbclass` for generating a BUP, but the normal tegraflash functions pretty much assume they're being called for image generation, hence the expectation of the rootfs.

It makes more sense, at least with the current hardware that more cleanly separates the boot firmware from the kernel+rootfs combination, to have a cleaner recipe-level separation, too. This is what I've been thinking about. If there's a recipe for building the boot firmware, we need it to deploy some artifacts that can be used for flashing, as well as artifacts that could be incorporated into the full tegraflash bundle with a rootfs.  I think that's possible, I'm just not quite sure how to do it cleanly.

anischali:
Totally agree with this, separating the low-level firmware from bootloader (EDK2 or others) and also from the rest of the disk — kernel, rootfs, etc. — and making that boundary fully modular. I think this could open up a lot of possibilities for the project, and could make it easier down the line to support other bootloaders, different disk layouts, or even other update technologies. Happy to help work through what that might look like if there's interest!

In the meantime, as a smaller first step, maybe we could just go with rootfs copy protection on its own until we land on a sweet, fully modular solution — that would already help unblock things on my end.

anischali:
Hi! On my end I went with a small workaround — truncating the rootfs size to 0 — which keeps my CI storage manageable for now.

Given that this doesn't seem to be a widespread need at the moment, closing this PR feels reasonable. I just wanted to leave the thought that supporting SPI-only flashing might eventually call for some separation in how artifacts are structured — no pressure at all, just something to keep in the back of the mind whenever the time feels right.

Happy to contribute or revisit this whenever it makes sense. Thanks for everything! 🙏

PR #2260, opened 2026-06-18
Add IGX Orin and IGX Thor developer kits

maximdeclercq:
## Summary

Adds machine support for the NVIDIA IGX Orin and IGX Thor developer kits on L4T R39.2.0 / JetPack 7.2, built on the existing `tegra234` and `tegra264` SoC includes. Both kits seem to use the p3740-0002 carrier.

| Machine | Module | Boot | FSI firmware | UEFI |
|---|---|---|---|---|
| `jetson-igx-orin-devkit` | p3701-0008 | eMMC (safety layout) | `fsi-lk.bin` | `general_igx` |
| `jetson-igx-thor-devkit` | p3834-0008 | NVMe | `fsi-fw-t264.bin` | `general_igx` |

## AGX/IGX differences

The IGX parts ship with the Functional Safety Island populated and run the IGX UEFI build. They are close to their AGX counterparts but not identical: IGX Thor is the same p3834-0008 module, while IGX Orin uses the p3701-0008 module (the AGX Orin devkit uses p3701-0000). The deltas versus the AGX siblings are:

- `TEGRA_FLASHVAR_FSIFW` set, so the FSI firmware is written to the fsi-fw partition (`tegra-storage-layout` already plumbs this when FSIFW is non-empty).
- `TEGRA_EDK2_CONFIGURATION = "general_igx"`, selecting the IGX platform UEFI (`uefi_t23x_general_igx` / `uefi_t26x_general_igx`), set before the SoC `require` so it overrides the `general` default.
- p3740-0002 / p3971-0000 carrier BCTs in place of the AGX devkit carrier.
- IGX-specific `nvpmodel` / `nvfancontrol` profiles (p3701-0008 for Orin, p3834-0008-p3740-0002 for Thor), and for Orin the safety partition layout and `pci=pcie_bus_safe`.

## Verification

I only have IGX Thor T5000 modules on hand, not the developer kits themselves, so I have not flashed or booted these machines on real dev-kit hardware yet. The values are taken directly from the L4T R39.2.0 board configs and packages and cross-checked against them value-by-value, so I am fairly confident they are correct, but a build and on-hardware boot pass on real dev-kit hardware is still wanted before relying on them.

### DIFF
--- a/conf/machine/include/igx-orin.inc
+++ b/conf/machine/include/igx-orin.inc
@@ -1,0 +1,45 @@
+# Common settings for IGX Orin modules
+
+NVPMODEL ?= "nvpmodel_p3701_0008"
+NVFANCONTROL ?= "nvfancontrol_p3701_0008"
+TEGRA_EDK2_CONFIGURATION ?= "general_igx"
+
+require conf/machine/include/tegra234.inc
+
+KERNEL_ARGS ?= "mminit_loglevel=4 console=tty0 console=ttyTCU0,115200 firmware_class.path=/etc/firmware fbcon=map:0 video=efifb:off console=tty0 efi=runtime nvme.use_threaded_interrupts=1 pci=pcie_bus_safe"
+
+EMMC_DEVSECT_SIZE ?= "512"
+BOOTPART_SIZE ?= "8388608"
+BOOTPART_LIMIT ?= "10485760"
+ROOTFSPART_SIZE_DEFAULT ?= "55834574848"
+TEGRA_FLASHVAR_ODMDATA ?= "gbe-uphy-config-8,hsio-uphy-config-0,nvhs-uphy-config-0"
+EMC_BCT ?= "tegra234-p3701-0008-sdram-l4t.dts"
+PARTITION_LAYOUT_TEMPLATE_DEFAULT ?= "flash_t234_qspi_sdmmc_safety.xml"
+TEGRA_AUDIO_DEVICE ?= "tegra-hda-jetson-agx"
+
+OTABOOTDEV ?= "/dev/mtdblock0"
+OTAGPTDEV ?= "/dev/mtdblock0"
+
+TEGRA_PLUGIN_MANAGER_OVERLAYS ?= "tegra234-carveouts.dtbo tegra-optee.dtbo"
+
+TEGRA_FLASHVAR_BPFDTB_FILE ?= "tegra234-bpmp-3701-0008-3740-0002-c00.dtb"
+TEGRA_FLASHVAR_BPF_FILE ?= "bpmp_t234-TA990SA-A1_prod.bin"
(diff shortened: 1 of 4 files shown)


madisongh:
I'll need @ichergui to look at this... the [meta-tegra-holoscan](https://github.com/nvidia-holoscan/meta-tegra-holoscan) layer has a machine configuration for some kind of IGX dev kit, and we don't want conflicts between the layers.

Also, I've been hesitant to add machine configurations for hardware that I can't be sure are at least getting build-and-boot tested on a regular basis. Some of that his carryover from the times when there were enough behavior differences between module models/families that it really required such testing. That might be less of an issue now, though, and perhaps this is an area where having NVIDIA on board with the project will help.

ichergui:
Hi @maximdeclercq 
Thanks for the PR. the L4T R39.2.0 is not the appropriate BSP for IGX.
[IGX SW 2.0](https://docs.nvidia.com/igx/user-guide/latest/software-releases/software-release-2-0-thor-notes-pr.html) supports `Thor/tegra264` only via `L4T R38.5.0`. 
[IGX SW 1.1.01](https://docs.nvidia.com/igx/user-guide/1.1.1/software-releases/software-releases.html) supports `Orin/tegra234` only via `L4T R36.4.5`.
`IGX Orin` and `IGX thor` didn't merged yet to the same BSP as the jetson.    
IGX uses different BSP than the jetson. We should avoid using IGX in `meta-tegra` for now. 
[meta-tegra-holoscan](https://github.com/nvidia-holoscan/meta-tegra-holoscan) supports only IGX Orin. I'm working on getting `L4T 38.5.0` ready for IGX Thor devkit and IGX thor mini. 

Please let me know if you are interested.

**Notes:**
- IGX Orin is a bit complicated because we need to build 2 different Linux images to align with the 2 modes (iGPU mode and dGPU).  So, nvGPU for Orin iGPU and OpenRM for the dGPU.
  - The compute stack are different  iGPU mode vs dGPU mode.
- IGX Thor needs only one build since the GPU driver can manage both iGPU and dGPU. 
  - The compute stack is the same. 2 GPU devices are visible to the userland.
- IGX uses different Linux kernel than Jetson.
- The next IGX SW release should support both Orin and Thor IGX devkits. 

Hopefully at some point we will have both IGX and Jetson uses the same L4T BSP. At that time, we can differently bring IGX support to `meta-tegra`

@madisongh @maximdeclercq  We should decline the PR for now.

maximdeclercq:
Thanks @ichergui, makes sense. R39.2 looked like it had the right pieces given the shared Thor configs, but I didn't realise IGX is still on its own BSP line, even outside of IGX-SW 2.0.

I did successfully boot AGX builds on my IGX T5000 module already, which is also what made me believe that an AGX-derived config could be possible, but I'll keep these in my own layer and revisit if they ever converge. Closing this.

PR #2265, opened 2026-06-20
L4T r39.2.0 dtb/dtbo build fixups

### LINKED ISSUE #2264: Device tree recipe refactoring issues
1. The bbclass file should be moved to `classes-recipe`
2. When trying to rework the `tegrademo-devicetree` recipe to use the bbclass, I discovered that because the overlays are only compiled/installed/deployed by the `nvidia-kernel-oot-dtb` recipe, they don't get deployed when `PREFERRED_PROVIDER_virtual/dtb` points to a custom device tree recipe.

For that second item, maybe moving the overlays to a separate recipe would be a reasonable approach? And maybe have the bbclass automatically add the dependency on the overlay recipe, perhaps via a variable that could be nulled out in the overlay recipe itself (or by users that decide they don't need those overlays)?

kekiefer:
The changes here fix a few problems with the new format of devicetree builds. External builds weren't building overlays, so we need new handling for that, and the external builds were missing an include path that was taken for granted in the build here. Also fix up a QA tmpdir path check failure along the way.

* tegra-devicetree: move to classes-recipe and improve search path
* nvidia-kernel-oot: skip .git dirs staging dts for _git version of recipe
* plugin_manager_overlays: depend on virtual/dtbo and split a recipe

Addresses #2264

### DIFF
--- a/classes-recipe/image_types_tegra.bbclass
+++ b/classes-recipe/image_types_tegra.bbclass
@@ -28,6 +28,8 @@ def tegra_dtb_extra_deps(d):
     deps = []
     if d.getVar('PREFERRED_PROVIDER_virtual/dtb'):
         deps.append('virtual/dtb:do_populate_sysroot')
+    if d.getVar('PREFERRED_PROVIDER_virtual/dtbo'):
+        deps.append('virtual/dtbo:do_populate_sysroot')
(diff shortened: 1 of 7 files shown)


ichergui (approved):
Excellent refactor @kekiefer. LGTM

PR #2272, opened 2026-06-23
39.2.0 CUDA sdk fixes

johannesschrimpf:
This PR includes three fixes for SDK generation and CUDA cross-compilation. 

To pull the CUDA host tooling into the SDK I added this to my local layer:

```
TOOLCHAIN_HOST_TASK:append:tegra = " nativesdk-packagegroup-cuda-sdk-host"
```

Tested on aarch64 SDK host, and I confirmed that the SDK for x86_64 also builds.

Please look at the commit messages for detailed descriptions.

### DIFF
--- a/recipes-devtools/cuda/cuda-shared-binaries.inc
+++ b/recipes-devtools/cuda/cuda-shared-binaries.inc
@@ -30,7 +30,20 @@ CUDA_VERSION_DASHED = "${@d.getVar('CUDA_VERSION').replace('.','-')}"
 
 CUDA_PKGNAMES ?= "${@' '.join(['%s-${CUDA_VERSION_DASHED}_${PV}_${CUDA_DEB_PKGARCH}.deb' % pkg for pkg in d.getVar('CUDA_PKG').split()])}"
 
-CUDA_INSTALL_ARCH ?= "${HOST_ARCH}"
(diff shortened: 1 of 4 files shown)


recipes-devtools/cuda/cuda-shared-binaries.inc
```diff
@@ -30,7 +30,20 @@ CUDA_VERSION_DASHED = "${@d.getVar('CUDA_VERSION').replace('.','-')}"
...
+# explicit class-native assignment below overrides the recipes' "${HOST_ARCH}"
+# so that native builds on aarch64 hosts also resolve to sbsa.
+def cuda_install_arch(d):
+    arch = d.getVar('HOST_ARCH')
+    if arch == 'aarch64':
+        return 'sbsa'
+    return arch
+
+CUDA_INSTALL_ARCH ?= "${@cuda_install_arch(d)}"
+CUDA_INSTALL_ARCH:class-native = "${@cuda_install_arch(d)}"
```
madisongh, line 46: Good catch on this, the CUDA_INSTALL_ARCH settings were a bad implementation - unnecessary duplication across all those recipes, for one thing, and not-well-thought-out overrides.
Our next step should be to get rid of the CUDA_INSTALL_ARCH settings in all of the recipes (including the overrides), and have just the one setting here. Then we won't need the `:class-native` override here, either.

PR #2274, opened 2026-06-23
cuda: clean up CUDA_INSTALL_ARCH usage

madisongh:
Replace the multiple settings of CUDA_INSTALL_ARCH (with and without overrides) with the one correct setting in the common include file.


madisongh:
Follow-up to #2272

madisongh:
Added a commit related to #2208

ichergui (approved):

Issue #2270, opened 2026-06-22
host1x_fence: Empty module_exit causes sysfs leak, reload failure, and kernel panic


leogao:
**Describe the bug**

nvidia-oot/drivers/gpu/host1x-fence/dev.c

Symptom: Three cascading failures:

modprobe host1x_fence after initial load fails with sysfs: cannot create duplicate filename '/class/host1x-fence' / kobject_add_internal failed for host1x-fence with -EEXIST
modprobe -r host1x_fence succeeds (module code freed) but /sys/class/host1x-fence persists
Any access to /sys/devices/virtual/host1x-fence/host1x-fence/uevent (e.g. by cat or udev) triggers kernel panic:
Internal error: Oops - Undefined instruction: 0000000002000000 [#1] SMP
pc : 0xffff80007bd001c8
lr : device_get_devnode+0x80/0xf8
Call trace:
  0xffff80007bd001c8 (P)
  dev_uevent+0x19c/0x248
  uevent_show+0x90/0x128
  dev_attr_show+0x28/0x80
  sysfs_kf_seq_show+0x90/0x118
  kernfs_seq_show+0x4c/0x70
  seq_read_iter+0x10c/0x480
Code: 00000000 00000000 00000000 00000000 (00000000)
Kernel panic - not syncing: Oops - Undefined instruction: Fatal exception

**To Reproduce**

$ modprobe -r

**Additional context**

Root cause: The module's exit function is completely empty:

static void __exit tegra_host1x_exit(void)
{
}
On modprobe -r:

Module code pages are freed/unmapped
But the sysfs class (/sys/class/host1x-fence), char device, and device node remain registered
The devnode callback (host1x_fence_devnode) still points to the now-freed module address 0xffff80007bd001c8
When any process reads the uevent sysfs file, the kernel calls device_get_devnode() → jumps to freed memory (all zeros) → Undefined instruction → kernel panic
The init function (host1x_uapi_init) registers: alloc_chrdev_region, class_create, cdev_add, device_create — none are ever torn down.

Note: The in-kernel host1x module does NOT register this class. The conflict is purely with the OOT module's own stale registration from a previous load.

leogao:
Fix (two patches required):

Patch 1: NULL-guard the mode pointer in host1x_fence_devnode(). The devtmpfs_delete_node path calls device_get_devnode with mode=NULL, so the callback must not unconditionally dereference it.

<pre lang="c">
[...]

PR #2277, opened 2026-06-24
nvidia-kernel-oot: fix host1x_fence crash on module unload and tegra_camera kmemleak false positive

leogao:
Fix two nvidia-kernel-oot issues:

### 1. host1x_fence: fix NULL deref and add module exit cleanup

`host1x_fence_devnode()` unconditionally dereferences the `mode` pointer, but
`device_destroy` calls `device_get_devnode` with `mode=NULL`, causing a kernel
panic on module unload.

Additionally, `tegra_host1x_exit()` was empty — no cleanup of the sysfs class,
cdev, or device. This leaves dangling sysfs entries whose devnode callback points
to freed module memory, triggering an undefined instruction panic on any subsequent
access to `/sys/class/host1x-fence/host1x-fence/uevent`. Reloading the module
fails with `-EEXIST`.

Fix: add a NULL guard for `mode`, and implement proper reverse-order teardown
(device_destroy → cdev_del → class_destroy → unregister_chrdev_region).

Related: https://github.com/OE4T/meta-tegra/issues/2270

### 2. tegra_camera: suppress kmemleak false positive

kmemleak reports unreferenced `video_device` objects allocated in
`tegra_channel_init_video()`. These are long-lived V4L2 structures referenced
through OOT module data that kmemleak cannot scan. They are properly freed in
`tegra_channel_cleanup_video()` on module unload.

Fix: mark with `kmemleak_not_leak()` to suppress false positives.

Signed-off-by: Wei Gao <wei.gao@windriver.com>

### DIFF
--- a/recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot.inc
+++ b/recipes-kernel/nvidia-kernel-oot/nvidia-kernel-oot.inc
@@ -80,6 +80,7 @@ COMMON_PATCHES = "\
     file://0064-nvdisplay-nvidia-drm-support-fbdev-on-Linux-6.11-ker.patch;patchdir=nvdisplay \
     file://0065-nvdisplay-nvidia-drm-trigger-connector-detect-on-hot.patch;patchdir=nvdisplay \
     file://0066-nvdisplay-nvidia-drm-update-legacy-modeset-state-on-.patch;patchdir=nvdisplay \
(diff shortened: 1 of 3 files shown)


madisongh:
As was pointed out in the discussion at #2270, for the L4T R39.2 release, NVIDIA opted to prevent the host1x-fence driver from being unloadable to address the first issue, and the second issue is just silencing a warning that kernel developers might see. So I'm not inclined to take these patches.

madisongh:
Closing because these changes diverge from upstream.

PR #2284, opened 2026-07-02
gcc-for-nvcc: isolate sstate swspec namespace

paletteguy:
The gcc-source-for-nvcc recipe currently uses the generic 'gcc' SSTATE_SWSPEC namespace. That overlaps with oe-core gcc-source and can cause unrelated cache matches when the two recipes diverge in patch content.

Switch the namespace token to 'gcc-for-nvcc' so shared-state objects are keyed independently for this recipe family. This avoids noisy cross-matches and reduces unnecessary hash churn in incremental builds nwhen either side updates its patch set.


madisongh:
Thanks, this change got missed way back when the recipes got renamed to `gcc-for-nvcc`.

While `whinlatter` isn't actively maintained any longer, I'll merge this and cherry-pick it to the active branches where it would also be needed.

PR #2286, opened 2026-07-03
Fix OPTEE test suite (xtest) for Thor devkit

### LINKED ISSUE #2285: OPTEE test suite (xtest) fails on Thor devkit
Running the OP-TEE test suite (`xtest -t regression`) on a Jetson AGX Thor
devkit (tegra264, OP-TEE 4.6, L4T r39.2.0) exposed two independent
problems:

1. **`xtest` crashed (SIGSEGV) at `regression_1024`**, aborting the entire
   regression suite.

```text
$ xtest -t regression 1024
Test ID: 1024
Run test suite with level=0

TEE test application started over default TEE instance
######################################################
#
# regression
#
######################################################

* regression_1024 Test PTA_SYSTEM_GET_TPM_EVENT_LOG Service
o regression_1024.1 TPM test service invocation
<no further output — process terminated by SIGSEGV>
$ echo $?
139
```

2. **`regression_1041` ("fTPM sanity") failed** because Linux never created
   `/dev/tpm0`, even though the firmware TPM works in the secure world.

```
* regression_1041 Test fTPM sanity
Waiting for TPM device 0 / 20
...
Waiting for TPM device 19 / 20
.../regression_1000.c:3400: found has an unexpected value: 0x0 = false, expected 0x1 = true
stat("/dev/tpm0"): No such file or directory
Perhaps fTPM hasn't finished probing
  regression_1041 FAILED
```

**To Reproduce**
1. Build `meta-tegra` branch `master` with `MACHINE` based on `jetson-agx-thor-devkit`
2. Build with bitbake argument `demo-image-full`
3. Deploy to hardware with method `initrd-flash`
4. run `xtest -t regression 1024` then `xtest -t regression 1041`

ichergui:
Refer to the issue #2285 to get more details


conf/machine/include/tegra264.inc
```diff
@@ -76,3 +76,5 @@ PREFERRED_VERSION_hafnium ?= "2.11-l4t%"
 
 TEGRA_EDK2_PLATFORM = "t26x"
 MACHINE_EXTRA_RDEPENDS += "nvidia-kernel-oot-compute nv-kernel-module-nvpmodel-clk-cap"
+
+TEGRA_BOOTCONTROL_OVERLAYS += "tegra264-ftpm.dtbo"
```
madisongh: That overlay looks like it's for the kernel, not a boot control overlay. It should be added to the TEGRA_PLUGIN_MANAGER_OVERLAYS setting.

madisongh (changes requested):
See in-line for one needed tweak.

I'm OK with the intent of the changes, but at some point we should provide settings that allow users to exclude the ftpm from builds, maybe as a MACHINE_FEATURES setting.

conf/machine/include/tegra264.inc
ichergui: Good catch. Thanks @madisongh

conf/machine/include/tegra264.inc
```diff
@@ -76,3 +76,5 @@ PREFERRED_VERSION_hafnium ?= "2.11-l4t%"
 
 TEGRA_EDK2_PLATFORM = "t26x"
 MACHINE_EXTRA_RDEPENDS += "nvidia-kernel-oot-compute nv-kernel-module-nvpmodel-clk-cap"
+
+TEGRA_PLUGIN_MANAGER_OVERLAYS += "tegra264-ftpm.dtbo"
```
madisongh: Please just add this to the `?=` setting in the `agx-thor-XXXX.inc` files, instead of using `+=` here.

madisongh (changes requested):

madisongh (approved):

PR #2290, opened 2026-07-07
cuda-target-environment: make host compiler extraction ccache-aware

### LINKED ISSUE #2288: recipes-devtools/cuda/cuda-target-environment_1.0.bb not ccache aware
**What**
recipes-devtools/cuda/cuda-target-environment_1.0.bb is extracting COMPILER_CMD from CXX_FOR_CUDA without checking for ccache usage like done cuda.bbclass hence the CUDAHOSTCXX environment variable in SDK becomes "ccache".

**How**
Building SDK with INHERIT += "ccache" in site.conf

**Suggest** 
Inherit from cuda.bbclass instead of cude-gcc.bbclass and use cuda_extract_compiler() function from cuda.bbclass.

paletteguy:
The recipe extracted COMPILER_CMD directly from the first element of CXX_FOR_CUDA, so when ccache is enabled (e.g. INHERIT += "ccache") the CUDAHOSTCXX variable in the generated SDK environment ended up being "ccache" instead of the actual host compiler.

Inherit cuda.bbclass instead of cuda-gcc.bbclass and use its cuda_extract_compiler() helper, which already skips the ccache wrapper, so the correct compiler is used consistently.

Fixes: OE4T/meta-tegra#2288

### DIFF
--- a/recipes-devtools/cuda/cuda-target-environment_1.0.bb
+++ b/recipes-devtools/cuda/cuda-target-environment_1.0.bb
@@ -6,11 +6,11 @@ SRC_URI = "file://cuda_target.sh.in"
 
 COMPATIBLE_MACHINE = "(cuda)"
 
-inherit cuda-gcc
+inherit cuda
 
 S = "${UNPACKDIR}"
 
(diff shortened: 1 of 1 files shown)


madisongh:
Thanks @paletteguy !

PR #2291, opened 2026-07-07
tegra-helper-scripts: backport T264 partial flash fix to r38.4

qerased:
Backport the T264 partial-flash fix from:

fb04dfd1a96d63b0edeb52c8f800f37a5764d201

The r38.4 branch still invokes `create_l4t_bsp_images.py` twice when
`-k <partition>` is used:

1. full workspace translation
2. second translation with `-k`

The first translation moves staged partition images, so the second call
can fail with `FileNotFoundError`. The wrapper then proceeds with an
inconsistent workspace and bootburn reports `s_ERROR_PARSE_CONFIG`.

Per-partition selection is already passed to `flash_bsp_images.py` as
`-u <partition>`, so the second `create_l4t_bsp_images.py -k` call is
redundant.

Observed on T264 / L4T r38.4 with:

    initrd-flash -k APP_b

Failure sequence:

    FileNotFoundError:
    ./tools/kernel_flash/images/external/my-image.ext4.simg

followed by:

    s_ERROR_PARSE_CONFIG

This change matches the fix already present on master.


madisongh:
I've manually cherry-picked fb04dfd1a96d63b0edeb52c8f800f37a5764d201 for this.

PR #2293, opened 2026-07-09
cuda.bbclass: forward DEBUG_PREFIX_MAP flags to nvcc host-compiler pass

paletteguy:
nvcc compiles .cu sources through a host-compiler pass but does not receive the C/C++ DEBUG_PREFIX_MAP flags. As a result, __FILE__ and debug paths from .cu files bake in the absolute build path (TMPDIR) and leak into packaged binaries, breaking reproducibility.

Add a CUDA_NVCC_PREFIX_MAP_FLAGS variable that wraps DEBUG_PREFIX_MAP in -Xcompiler and append it to CUDA_NVCC_FLAGS, so .cu objects are normalized the same way as .cpp objects for every cuda-inheriting recipe.

Drop the now-redundant local cudify_flags helper and CUDAFLAGS override from tensorrt-plugins, which is superseded by the shared handling in cuda.bbclass.


ichergui (approved):
LGTM. Thanks @paletteguy

paletteguy:
@madisongh Any thoughts?

paletteguy:
@madisongh  Is it possible to get this into origin/whinlatter?

PR #2295, opened 2026-07-09
agx-thor: fix r8168/r8169 typo in MACHINE_EXTRA_RRECOMMENDS

bergman-kai:
kernel-module-8168/8169 are missing the 'r' prefix used by the real package names, so RRECOMMENDS silently no-ops and the Realtek driver never installs on agx-thor-t4000/t5000 despite being built.


ichergui (approved):
Thanks @kaibergmanlodestar

kekiefer (approved):

PR #2297, opened 2026-07-11
Several UEFI-related updates

madisongh:
The main feature being the addition of a simple way to enable a trimmed-down UEFI build with L4TLauncher built directly into the UEFI image, to make boots faster and deal with the reliability problem that the ESP partition presents for A/B updates.

There are also some fixes that aren't directly related to the minimal-boot implementation, plus a setting added to make it easier to get verbose logs out of UEFI.


recipes-bsp/uefi/files/0001-feat-capsule-updates-with-single-boot-configuration.patch
```diff
@@ -0,0 +1,161 @@
...
+From: Matt Madison <matt@madison.systems>
+Date: Thu, 15 Jan 2026 00:40:20 -0800
+Subject: [PATCH] feat: capsule updates with single-boot configuration
+
+Let the EFI boot manager find ESP partitions, even for a single-boot
+setup, so that capsule updates can be applied. Have the PlatformBootOrder
+library prepend the single-boot application to the boot order, so the ESP
+does not need to contain a boot launcher.
+
+Signed-off-by: Matt Madison <matt@madison.systems>
```
joekale-pp, line 18: Does this also require the Upstream-Status? And is this something that should be offered upstream eventually or should it be deemed Inappropriate?

madisongh: I've marked it `Pending` for now. Assuming it works OK, I was planning on opening a PR upstream.

madisongh:
Tested this with Orin and Thor devices, using the setup [here](https://github.com/OE4T/tegra-demo-distro/pull/399).

joekale-pp (approved):
My comments are all resolved.

ovaag:
Great additions! 

Just wanted to let you know about one boot failure I ran into when experimenting with running essentially this patch set (ported to r36.5.0/scarthgap).

It seems to be a problem with the boot-order sort added in `0001-feat-capsule-updates-with-single-boot-configuration.patch`. It never runs on a varstore inherited from an earlier, non-single-boot firmware** — e.g. after a JetPack 5 → JetPack 6 capsule-update transition.

**Symptom** (Orin NX, r35.6.2 to r36.5.0 via swupdate + capsule): the capsule applies fine and the new firmware boots, but the embedded "Boot Application" never runs. The inherited BootOrder still has the disk/ESP entry first, so BDS loads the *old* r35 `BOOTAA64.EFI` from the ESP, which can't boot under the new firmware, and the unit warm-resets in a loop until the rootfs A/B logic drains its retries and fails back to the old boot chain. A verbose-logging build shows it directly:

```
PrintBootOrder: BootOrder[0] = 0x0001 = UEFI KINGSTON ...          <- inherited r35 entry
PrintBootOrder: BootOrder[4] = 0x0008 = Boot Application           <- never reached
add-symbol-file .../35.6.2-r0/.../L4TLauncher/DEBUG/L4TLauncher.dll  <- old ESP launcher runs
Shutdown state requested 1
```

**Root cause**: stock `SetBootOrder()` in `PlatformBootOrderLib.c` starts with

```c
Status = gRT->GetVariable (L"PlatformBootOrderSet", &gNVIDIATokenSpaceGuid, ...);
if (!EFI_ERROR (Status) && (VariableData == TRUE)) {
    return;   // sort already done once, skip
}
```

`PlatformBootOrderSet` is NV and was set to TRUE by the previous firmware generation long ago, so the `SetBootOrder()` call added in `PlatformBootManagerBeforeConsole` is a no-op on any upgraded unit. A freshly flashed varstore doesn't have the variable, the sort runs, and everything works, which is why fresh-flash testing doesn't catch it. Since the variable is BS-only (no RUNTIME_ACCESS), it also can't be cleared from the OS side as a workaround; it has to be handled in firmware.

Tested with this patch, and that did make the r35.6.2->r36.5.0 OTA upgrade succesfull. 

```patch
--- a/Silicon/NVIDIA/Library/PlatformBootManagerLib/PlatformBm.c
+++ b/Silicon/NVIDIA/Library/PlatformBootManagerLib/PlatformBm.c
@@ -2196,6 +2196,27 @@ PlatformBootManagerBeforeConsole (
       mRscHandler->Register (SingleBootStatusCodeCallback, TPL_CALLBACK);
     }
 
+    //
+    // A varstore inherited from an earlier, non-single-boot firmware (e.g.
+    // after a JetPack 5 to 6 capsule update) may already contain
+    // PlatformBootOrderSet=TRUE, which makes SetBootOrder() return without
+    // sorting and leaves a stale BootOrder that boots the old ESP launcher
+    // instead of the embedded boot application. Clear the flag so the sort
+    // runs on every boot.
+    //
+    {
+      BOOLEAN  BootOrderSet;
+
+      BootOrderSet = FALSE;
+      gRT->SetVariable (
+             L"PlatformBootOrderSet",
+             &gNVIDIATokenSpaceGuid,
+             EFI_VARIABLE_BOOTSERVICE_ACCESS | EFI_VARIABLE_NON_VOLATILE,
+             sizeof (BootOrderSet),
+             &BootOrderSet
+             );
+    }
+
     EfiBootManagerRefreshAllBootOption ();
     SetBootOrder ();
 
```

classes-recipe/image_types_tegra.bbclass
```diff
@@ -93,7 +93,7 @@ LNXFILE ?= "boot.img"
 LNXSIZE ?= "83886080"
 TEGRA_RECOVERY_KERNEL_PART_SIZE ??= "83886080"
 RECROOTFSSIZE ?= "314572800"
-TEGRA_EXTERNAL_DEVICE_SECTORS ??= "119537664"
+TEGRA_EXTERNAL_DEVICE_SECTORS ??= "490733568"
```
ichergui, line 96: I think this is aligned with UFS size not the NVMe.  I think we should keep the original value

ichergui:
Thanks @madisongh for this PR.

classes-recipe/image_types_tegra.bbclass
madisongh: Admittedly I based this new default on the EXT_STORAGE_SIZE setting in the L4T t264.conf file. The p3701 and 3767 common configs do still have this dialed down to 57GiB. Doesn't appear to be tied to UFS, though.

Any particular reason why the original value should be kept

ichergui: The original value was calculated based on rootfs size 55GiB IIRC.

ichergui: I will try those changes today and share my feedback.

madisongh: > based on rootfs size 55GiB IIRC

Not rootfs size, but total size of the storage device.

madisongh: I'll take silence as agreement.

madisongh:
@ovaag Thanks for pointing out the boot order problem. I've taken a slightly different approach to resolving that problem, but the effect should be the same.

madisongh:
The latest commits add some variable settings and documentation around the FMP system image type GUID. I think this will be the last of the changes, for now, unless I've missed something.

madisongh:
I've made some additional edits to the documentation and rebased on master latest. After I run through another round of tests, this should be ready to go.

PR #2306, opened 2026-07-16
l4t-graphics-demos: remove unneeded weston dependency

madisongh:
The weston-dmabuf-formats makefile has, since time immemorial, linked directly against core weston libraries. It could be, that at some point in history, the code actually made direct calls to the weston core, but it certainly does not any longer.

Add a patch to remove the unnecessary linkage, and remove weston from the build dependencies.


madisongh:
@leon-anavi This should remove the problematic runtime dependency on weston you're running into.

trini:
I can confirm this fixes the problem @madisongh, thanks!

leon-anavi:
> @leon-anavi This should remove the problematic runtime dependency on weston you're running into.

Thank you! I also confirm that the patch in this pull request fixes the specific problem I was having and all tests from `yocto-check-layer ../meta-tegra/ --machines jetson-orin-nano-devkit-nvme` pass with it.

PR #2308, opened 2026-07-20
Introduce linux-yocto kernel metadata and support tegra264

kekiefer:
This series primarily adds the `tegra-kernel-cache`, to be the source of patch sets and any relevant configuration for linux-yocto. Now we are able to more naturally manage the -rt kernel configs as well, so support for this in linux-yocto is also introduced.

The work also revealed some deficiencies in the existing shared configurations for wifi and ethernet and their associated firmware, so there's some cleanup here.

I do apologize in advance for the change that disables git maintenance during patching, but without this, the giant patch set brought my workstation to its knees, and I'm sure this would hit others as well.


conf/machine/include/devkit-ethernet.inc
```diff
@@ -0,0 +1,10 @@
+# 10G Ethernet (Orin AGX only) built as module in kernel cache
```
madisongh: I forget, are these differences due to different Ethernet MACs being on the carrier board, or are they on the modules themselves?

If they're carrier board differences, I'm thinking maybe we should change the `devkit-<function>.inc` files into carrier-specific files (for p3737, p4071, p3768). That would eliminate the chip-specific overrides, which would be better to avoid, if possible, since it makes the settings harder for users to customize.

madisongh (commented):
LGTM overall, but I had one question in-line

conf/machine/include/devkit-ethernet.inc
kekiefer: These are on the carriers themselves. You’re right that logically tying them to the processor variant is the wrong abstraction in this way. 

The other option would be to add another machine override for the carrier itself and leaving these where they are.

madisongh: Let's start with refactoring the .inc files for now. Hopefully we won't need an override for the carrier, as I can see that adding complexity that would be nice to avoid.

kekiefer:
Being the only devkit- style include left, I also went ahead and factored out the devkit-audio.inc

PR #2309, opened 2026-07-22
Convert licenses to SPDX expressions

krant:
Recent changes in openembedded-core has made meta-tegra (and few other layers too) to produce a lot of license-related warnings. This PR chases the upstream and fixes them.


madisongh:
Thanks!

PR #2311, opened 2026-07-24
New recipes + Fixes 

ichergui:
* tegra-helper-scripts: forward RAMCODE to RCM-boot BCT generation on T234
* python3-jetson-io: set MAINSUM for the L4T debian package
* vulkan-sc: add tegra-libraries-vulkan-sc and tegra-vulkan-sc-samples
* vulkan-sc: add tegra-libraries-vulkan-sc-core
* jetson-sipl-api: add tegra-nvsci-headers dep, exclude nvsci headers from install
* tegra-libraries-nvsci: add build dependency on tegra-nvsci-headers
* tegra-nvsci-headers: add recipe for NvSci build-time headers
* nvgstapps: add nvgstipctestapp
* tegra-libraries-camera: package nvargus_nvraw binary

PR #2314, opened 2026-07-27
New recipes + Fixes [v2]

kekiefer:
These are updates to #2311 addressing review comments. I can't push to the original PR branch, and Ilies is on vacation for a couple weeks, so I'm resubmitting the PR here so I can manage it.


kekiefer:
A couple new commits versus the last one, sipl_coe_sample is new, and I've made a new commit for the license fixups so that backporting these is easier

shr-project:
[This](https://github.com/OE4T/meta-tegra/pull/2314/commits/6e112cf0b43f4228f5a02096975d274a41b6ea74) seems to cause:
```
ERROR: QA Issue: /usr/sbin/nvargus_nvraw contained in package tegra-libraries-camera-nvraw requires libX11.so.6()(64bit), but no providers found in RDEPENDS:tegra-libraries-camera-nvraw? [file-rdeps]
```

I guess it should be installed only when x11 PACKAGECONFIG is enabled.

PR #2315, opened 2026-07-29
tegra-libraries-camera: install nvargus_nvraw only when x11 PACKAGECO…

shr-project:
…NFIG is enabled

Fixes https://github.com/OE4T/meta-tegra/pull/2314

### DIFF
--- a/recipes-bsp/tegra-binaries/tegra-libraries-camera_39.2.0.bb
+++ b/recipes-bsp/tegra-binaries/tegra-libraries-camera_39.2.0.bb
@@ -66,12 +66,12 @@ do_install() {
     cp -R ${B}/var/nvidia ${D}${localstatedir}/
     if ${@bb.utils.contains('PACKAGECONFIG', 'x11', 'true', 'false', d)}; then
         install -d ${D}${libdir}/libv4l/plugins/
-	install -m 0644 ${S}/usr/lib/aarch64-linux-gnu/nvidia/libv4l2_nvargus.so ${D}${libdir}/libv4l/plugins/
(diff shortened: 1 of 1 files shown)


recipes-bsp/tegra-binaries/tegra-libraries-camera_39.2.0.bb
```diff
@@ -67,11 +67,11 @@ do_install() {
     if ${@bb.utils.contains('PACKAGECONFIG', 'x11', 'true', 'false', d)}; then
         install -d ${D}${libdir}/libv4l/plugins/
 	install -m 0644 ${S}/usr/lib/aarch64-linux-gnu/nvidia/libv4l2_nvargus.so ${D}${libdir}/libv4l/plugins/
+        install -m755 ${B}/usr/sbin/nvargus_nvraw ${D}${sbindir}/
```
madisongh: Need to create ${D}${sbindir} here, or this is apt to fail.

madisongh (changes requested):
Thanks, but it needs a tweak

recipes-bsp/tegra-binaries/tegra-libraries-camera_39.2.0.bb
shr-project: I'm sorry I've overlooked it, I should have tried the build with x11 enabled as well, updating it now.

madisongh: Drat... I pulled the trigger too soon, the update wasn't quite it, either

PR #2318, opened 2026-08-05
initrd-flash: fix unmounting of automounted partitions

markyang92:
mount_partition() looks up the mountpoints of the target device with a prefix match on /proc/mounts, so for /dev/sdX1 it also picks up /dev/sdX10 and above when the desktop has automounted them. The unmount loop then trips over the extra entries: it reassigns mnt while indexing through the list, so the second iteration retries the mountpoint it just unmounted and bails out:

  ERR: unmount /media/user/4bafadc4-... on device /dev/sdf1 failed

Anchor the match on the whole device name and iterate over the list.

Seen on an Orin Nano with an 18-partition layout: after one flash the two-digit partitions carry filesystems, the desktop automounts them, and the next initrd-flash run fails to unmount /dev/sdf1.


madisongh:
Thanks!

PR #2319, opened 2026-08-10
Minor Orin Nano storage fixes

TheYoctoJester:
Two minor fixes related to uSD/eMMC handling on an Orin Nano.

d218b2a700d41e3ba2efa7425fa7bcb35cee02a0 saves the uSD card content in the corner case situation that the card is present while flashing the NVME through `tegraflash`
a043d29f12e6ee646ca95ee4197f9a083d1eb06f is just an inert default variable assignment cleanup.


madisongh:
Some of this dates back to the days when the SDcard device *was* `/dev/mmcblk1` instead of `/dev/mmcblk0`, but that was long ago.

TheYoctoJester:
@madisongh thanks! Should this be cherry-picked to `master` too? Or which branches might be sensible, I just happened to find it on `wrynose`.

madisongh:
I've taken care of the cherry-pick to master.

PR #2320, opened 2026-08-12
CONTRIBUTING.md: update and include AI policy

madisongh:
* The old OE wiki link we were referencing has been replaced by a section in the YP documentation.

* Add some text about AI-assisted contributions


ichergui (approved):
LGTM. Thanks @madisongh

CONTRIBUTING.md
```diff
@@ -46,13 +46,61 @@ involved.
...
+
+* not understanding the problem, or a proposed solution, yourself
+* using an AI tool to interact directly with project community members, since
+the signal-to-noise ratio of text generated by AI tools can be very low
+* blindly trusting the output of the tool to be accurate, correct, and complete
+
+For patches, understand the changes you are making and verify that they work, and
+please write the commit messages yourself. Please also write PR descriptions
+yourself. Make use of whatever tools you need to help with grammar and spelling, but
+keep commit messages and PR descriptions concise.
```
dwalkes: I'm wondering about mentioning translation here as something else AI can be used with regarding the commit message or PR description.  I wonder if it might be easier to write commits in a first language and have AI translate to English if I were a non english speaker.  I'd have no problems with that myself. 

I think I'd consider allowing use of AI for commit messages and PR as long as you don't blindly paste commit and PR descriptions from AI tools.   I know that's a little different than what I said earlier, just wondering if this is too restrictive as I read it now.

CONTRIBUTING.md
```diff
@@ -46,13 +46,61 @@ involved.
...
+guidelines when using AI tools to help you with contributions to the project.
+
+## Be up-front about using AI tools
+
+If you have used one or more AI tools to help you with reporting issues,
+creating patches, writing documentation, or making other contributions to
+the project, please say so. For patches/PRs, include an `AI-Generated:`
+line in the commit message to name the AI tool(s) used. You should not
+attribute co-authorship to tools, and the `Signed-off-by:` in your
+commit message must be from you, the developer.
```
dwalkes, line 73: Could include an example here like:
```
Signed-off-by:  me@developer.com
AI-Generated: Claude Code
```
FWIW I see Co-Authored-By: much more common so we should possibly emphasize the *not* in You should not attribute co-authorship if we specifically don't want to use that.  I missed it in my first read.

dwalkes (approved):
Thanks @madisongh two minor thoughts here but please go with your wishes on these.

kekiefer (approved):
Looks good to me, i think this captures all the concerns we’ve discussed.

cbrake:
Nice work on the AI policy!

PR #2321, opened 2026-08-13
Drop deprecated gptfdisk dependency

maximdeclercq:
`oe-core` removed the gptfdisk recipe ([commit](https://git.openembedded.org/openembedded-core/commit/?id=188d84bcc136e87d03736606fb8e49f35df47fae)), so this layer no longer builds on `master`.

- `image_types_tegra` declared `gptfdisk-native`, but nothing in the class
  runs sgdisk, except `make_sdcard`, which runs on the flashing host.
- `tegra-flash-init` used sgdisk, replaced by sfdisk now.

make-sdcard.sh still needs sgdisk on the flashing host, but I left this unchanged since this is a flashing host dependency.


madisongh:
Thanks for taking care of this... I saw the OE-Core patches go by the other day, I just hadn't had a chance to deal with the changes yet.

PR #2323, opened 2026-08-13
Jetpack 7.2.1 update integration

kekiefer:
Bumps the layer to R39.2.1/JetPack 7.2.1, a bugfix point release on top of R39.2.0/JetPack 7.2. Refer to [Jetson Linux R39.2.1 release notes](https://docs.nvidia.com/jetson/archives/r39.2.1/ReleaseNotes/Jetson_Linux_Release_Notes_r39.2.1.pdf) for what's included in the updates. 

A few OE4T-specific fixes included here as well:

- tegra-uefi-prebuilt skips when `TEGRA_MINIMAL_BOOT` is set
- Vendor-kernel check factored out into an include file. The prior check didn't account for the -rt kernel variant
- Fixed a bug in nvgstplayer for codecs like avdec_amrnb/avdec_amrwb that need an audioconvert in the chain to work
- Split previously skipped pylibjetsonpower into a tegra-nvpower-python subpackage. Added as `RRECOMMENDS` to install by default.
- Added kernel-module-spidev to tegra-common's `RRECOMMENDS` so /dev/spidev* nodes are available.
- systemd and boot status cleanup for usb device mode


kekiefer (commented):
@madisongh reflecting on the addition of a variable here and I would like to get your take. It would be a little more clever to inline a function:

```
def using_vendor_kernel(d):
    return (d.getVar('PREFERRED_PROVIDER_virtual/kernel') or '').startswith('linux-noble-nvidia-tegra')
```

Then the check becomes:

```
DEVKIT_ETHERNET_RRECOMMENDS ?= "${@'${NVIDIA_ETHERNET}' if using_vendor_kernel(d) else '${LINUX_ETHERNET}'}"
```

The upside here is then users won't think they can just coerce the value of this variable and get the correct behavior. The downside being that it might be a little harder to discover.

Let me know if you've got a strong opinion here one way or the other.

madisongh:
> @madisongh reflecting on the addition of a variable here and I would like to get your take.

Yeah, I saw the new variable. Other than the too-generic name (it probably should have TEGRA in it), I thought it was an OK approach. If you use a function, since most of the checks are in conf files, rather than recipes, you're better off putting it under `lib/oe4t` - a function definition in a conf file might work, but it's not really the done thing.

Some layers use an override to distinguish between upstream/downstream kernels, but I think that's often because the difference can have more wide-ranging impact on other recipes. I think that's probably overkill for us, though. Besides, overrides can be difficult for users to deal with when they need their own customizations.

Anyway, either variable or function would be OK, I think.

docs/release-notes/JetPack-7.2.1-L4T-R39.2.1-Notes.md
```diff
@@ -0,0 +1,25 @@
+As of 12 Aug 2026, the `master` and `wrynose` branches support JetPack 7.2.1/L4T R39.2.1.
```
madisongh: The text in this line dates back to when we were using wiki pages. Now that the documentation is in the repository itself, folks can look at the commit history to find this out.

docs/Which-branch-should-I-use.md
```diff
@@ -12,8 +12,8 @@ For Jetson Linux (L4T) releases:
 * branches corresponding to long-term support (LTS) OE-Core release branches are kept up-to-date with the Jetson Linux releases. When there is a significant Jetson Linux upgrade, an additional LTS branch is created for the older release series.
 
 Active branches:
-* **master** - never stable, [L4T R39.2.0/JetPack 7.2](release-notes/JetPack-7.2-L4T-R39.2.0-Notes.md) for AGX Orin/Orin NX/Orin Nano/AGX Thor
-* **wrynose** - [L4T R39.2.0/JetPack 7.2](release-notes/JetPack-7.2-L4T-R39.2.0-Notes.md) for AGX Orin/Orin NX/Orin Nano/AGX Thor
+* **master** - never stable, [L4T R39.2.1/JetPack 7.2.1](release-notes/JetPack-7.2.1-L4T-R39.2.1-Notes.md) for AGX Orin/Orin NX/Orin Nano/AGX Thor
+* **wrynose** - [L4T R39.2.1/JetPack 7.2.1](release-notes/JetPack-7.2.1-L4T-R39.2.1-Notes.md) for AGX Orin/Orin NX/Orin Nano/AGX Thor
```
madisongh: Let's hold off on changing the wrynose line until we complete the back-port.

docs/Which-branch-should-I-use.md
```diff
@@ -1,6 +1,6 @@
 # Currently maintained branches #
 
-Last update: 28 Jun 2026
+Last update: 13 Aug 2026
```
madisongh: Again, since the docs are in the repo now, we can omit these history markers.

conf/machine/include/tegra-common.inc
```diff
@@ -29,6 +29,7 @@ TNSPEC_BOOTDEV_DEFAULT ??= "mmcblk0p1"
 TNSPEC_BOOTDEV ??= "${TNSPEC_BOOTDEV_DEFAULT}"
 
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-noble-nvidia-tegra"
+BUILD_IS_USING_VENDOR_KERNEL = "${@'1' if (d.getVar('PREFERRED_PROVIDER_virtual/kernel') or '').startswith('linux-noble-nvidia-tegra') else ''}"
```
madisongh: As I mentioned in our discussion the other day, using a variable like this should be OK, but I'd prefer a less-generic variable name, to avoid potential conflicts with other layers. So something like `TEGRA_USE_VENDOR_KERNEL` would be better.

I know we haven't always been so good at doing that in the past, but I'd rather  not keep introducing new cases of this.

madisongh (commented):
So far, looks good overall, with some good fixes, too, thanks!

conf/machine/include/tegra-common.inc
kekiefer: I went ahead with `TEGRA_USING_VENDOR_KERNEL` to keep the spirit of your ask, but make some feeble attempt to indicate it's a readonly thing.

kekiefer:
FYI rebases here since the updates have just been in keeping this up-to-date after the last round, post comments. No functional changes.

recipes-bsp/uefi/edk2-firmware-core-tegra-39.2.1.inc
madisongh, line 1: Can we get this and `edk2-firmware-tegra-39.2.1.inc` updated with the new branch and SRCREVs? I know the only changes are around the L4TPreIsoInstaller stuff, which aren't relevant to us, but it would be good to align now, just in case future patches come along.

madisongh (changes requested):
See inline. Overall looks good, though, thanks!

recipes-bsp/uefi/edk2-firmware-core-tegra-39.2.1.inc
kekiefer: Done

kekiefer:
Rebased and added a couple more details about the Jetpack SDK component version bumps in the release.

madisongh:
Merged this manually, since GitHub was mistakenly convinced that there were conflicts that needed manual resolution.

PR #2325, opened 2026-08-18
tegra-configs: add video codec container CSV entries

qerased:
Add missing video codec library and symlink entries to tegra-configs drivers.csv so NVIDIA container runtime exposes them inside containers.


madisongh:
Thanks!
BTW, the DCO check failed because it's expecting the name in your `Signed-off-by:` to match the name you set in your GitHub account profile, I think.

qerased:
oh thanks
can you cherry-pick this commit into https://github.com/OE4T/meta-tegra/tree/master-l4t-r38.4.x branch?

ichergui:
@qerased Any reason of using `master-l4t-r38.4.x` ? I recommend to use `master` branch instead.

qerased:
> @qerased Any reason of using `master-l4t-r38.4.x` ? I recommend to use `master` branch instead.

@ichergui We currently need to stay on master-l4t-r38.4.x because our platform is tied to the corresponding BSP baseline. Moving to master would require additional validation on our side. Maybe in nearest future we'll move to it :)

PR #2326, opened 2026-08-18
opencv: Adjust bbappend for 4.14 upgrade from meta-oe

kekiefer:
Version moved from 4.13 to 4.14 and can now drop two CUDA-related patches that were incorporated into the new version.


kekiefer:
Previously, this patch was sitting in the queue for 7.2.1 in order to get checklayer passing, but there's no reason to hold up this update. FYI @ichergui

madisongh:
Thanks @kekiefer, I was wondering why that got mixed in with the 7.2.1 update PR when it was largely orthogonal

PR #2331, opened 2026-08-24
docs: Creating-a-custom-MACHINE.md: generic bbappend for tegra-bootfiles

michaelopdenacker:
Refer to "tegra-bootfiles_%.bbappend" instead of using a fixed version that is more costly to maintain.

This allows to get rid of an outdated reference to version 35.4.1.


michaelopdenacker:
Thanks a lot Matt!

PR #2338, opened 2026-08-28
cuda-culibos: add recipe

ichergui:
Package the cuda-culibos component from the CUDA 13.2.86 binary release. It provides libculibos.a, the static support library that the CUDA math libraries link against, so it needs to be available on target as well as for native and nativesdk builds.


ichergui:
@madisongh This PR should be merged after #2323 
Thanks

madisongh:
Merged manually

PR #2341, opened 2026-09-03
optee: use Type=notify for tee-supplicant.service

embetrix:
tee-supplicant sends READY=1 once it has opened the supplicant device and can serve the secure world (optee_client [a5b1ffcd](https://github.com/OP-TEE/optee_client/commit/a5b1ffcd26e328af0bbf18ab448a38ecd558e05c), v4.3.0), unlike Type=exec which only covers execve().


madisongh:
Thanks!

Issue #2330, opened 2026-08-21
Use of bmap-tools for initrd-flash is incompatible with DM-Verity


Shamino0:
On the system I'm developing (based on the `scarthgap` branch of `meta-tegra`), we're using DM-Verity to sign our root file system (via the `dm-verity-img.bbclass` in the `meta-security` layer).  We got this working and it's all good.

But we found that if `bmap-tools` is installed on the host computer, then `initrd-flash` (or more specifically, the `make-sdcard` script it calls) uses it (as is documented), and doing so breaks the Verity hashes.

What's happening is that if there was a previous build installed on the module, then `bmaptool` does not write any blocks that are not mapped in the source image (which is what it's meant for).  Which means those blocks are not overwritten on the target device.  This isn't normally a problem, because those blocks should not be used by the `ext4` file system, but the presence of non-zero data in those unused blocks breaks the DM-Verity hash, causing the kernel to panic (the behavior we want if the root file system becomes corrupted).

I suspect this might also be a security problem for other reasons - to have data from an old installation present after installing an upgrade might leak something sensitive.

As a quick workaround, I can pass the `--erase-nvme` option to `initrd-flash`.  That solves the problem, but it also means that other partitions not normally overwritten (e.g., UDA) get wiped out.  Making it inappropriate for a system upgrade procedure.

As a somewhat better workaround, I modified `make-sdcard.sh` to not check for `bmaptool`.  This works, causing `make-sdcard` to fall back to using `dd` to write the image.  But of course, it greatly slows down installations.

I think the best solution is to make changes to the install procedure so the device side (running the `tegra-initrd-flash-iniramfs` image) could be told which partitions will be overwritten by image data, so it can call `blkdiscard` on those partitions before the upload.  This way, unused blocks are zeroed out, and we can still use `bmaptool` to speed up the installation.

This change is too much work for me to attempt and would (I think) require a pretty substantial change to the installation procedure, but perhaps someone here is expert enough to make the change.  I think the installation flow would become something like:

In Step 4 (Writing partitions on external storage device):
* The host creates the partitions.
* The host disconnects the USB gadget, and the device destroys the gadget.
* The host tells the device (maybe via the flashpkg instructions) which partitions should be erased
* The device runs `blkdiscard` to erase those partitions
* The device re-creates the gadget and waits for another connection
* The host now proceeds as before, writing images to partitions, knowing that those partitions have been erased

Please consider making this (or a similar) change to the installation process.

[...]

PR #2343, opened 2026-09-07
tegraflash: support encrypted initrd-flash partitions

ericst:
Allow Orin flashing workflows to prepare encrypted partitions directly on their target block device, so the LUKS container, passphrase derivation, and payload write align with the boot-time unlock path.

Bundle the required host-side NVIDIA tools and document the required per-device inputs and security boundary; production key management remains outside the layer.

### DIFF
--- a/classes-recipe/image_types_tegra.bbclass
+++ b/classes-recipe/image_types_tegra.bbclass
@@ -395,6 +395,9 @@ tegraflash_populate_package() {
         cp "${STAGING_DATADIR}/tegraflash/$f" .
     done
     sed -e "\$a\BOOTCONTROL_OVERLAYS=\"$bcoverlays\"" ${STAGING_DATADIR}/tegraflash/flashvars > ./flashvars
+    install -d ./tools/disk-encryption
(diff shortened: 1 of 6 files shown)


madisongh:
As I mentioned [here](https://github.com/OE4T/meta-tegra/issues/2330#issuecomment-5386281158) a couple of weeks ago, I'm not really willing to make major changes to the flashing process right now, until we have a better handle on the Orin/Thor dichotomy.

I'd also be a bit concerned using the approach implemented here. Besides the key management, I'd be worried that mismatches between host and target environments could cause the LUKS formatting by the host to be unreadable on the target.

That's why I was looking (as mentioned in that comment) into switching to per-partition USB exports, keeping all the LUKS stuff self-contained on the target side.

ericst:
Fair enough, we need to have it so I will just take this into our layer. For me the advantage of doing it from the flashing host was that it 1. the eks.img is anyway generated there and 2. it would exercise a recovery path on flashing as well. But of course it all depends on your requirements, and if you need to ensure the keys never leave the target maybe it is better.

I might give a stab at the mapper export approach though.

PR #2345, opened 2026-09-08
l4t-usb-device-mode: give ttyGS0 its own getty template

kekiefer:
agetty invoked with "-" (the serial-getty@ default) takes an exclusive flock() on /dev/console to arbitrate between redundant console= entries. Being an auxiliary virtual console that may not even be plugged in at startup, ttyGS0 shouldn't be competing for /dev/console. If it happens to win the race at startup and remain unenumerated, it will starve the real console getty forever, with no error or log output.

Added a usb-gadget-getty@.service template that invokes agetty with %I instead of "-", and pointed the udev rule at it instead of serial-getty@.

### DIFF
--- a/recipes-bsp/l4t-usb-device-mode/l4t-usb-device-mode.bb
+++ b/recipes-bsp/l4t-usb-device-mode/l4t-usb-device-mode.bb
@@ -7,6 +7,7 @@ SRC_URI = "\
     file://70-l4tbr0.network \
     file://l4tbr0.netdev \
     file://98-usb-gadget-tty.rules \
+    file://serial-getty@ttyGS0.service.d/no-tty-reset.conf \
 "
 
 COMPATIBLE_MACHINE = "(tegra)"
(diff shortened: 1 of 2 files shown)


ichergui (approved):

madisongh:
Can you point me at the code in `agetty` that's doing the `flock` on `/dev/console`? I'm not finding it. I also don't see how `agetty` treats the `-` argument any differently than a device name, except during parsing of its command line.

kekiefer:
I jumped to this conclusion with an strace while this was going on, but I’ll see if I can dig up the source evidence. 

Edit to add, that my conclusion was based on the convention of ‘-‘ in a Unix program as the use of a stream for the IO, and that was /dev/console here and not the device itself.

madisongh:
> Edit to add, that my conclusion was based on the convention of ‘-‘ in a Unix program as the use of a stream for the IO, and that was /dev/console here and not the device itself.

I guess that might be the case...`agetty` is getting the device name for the `stdin` fd, but the standard unit file for serial-getty sets `TTYPath` to the tty device, and sets `StandardInput` to `tty`, so that should be resolving to the tty device, unless you have a drop-in config file somewhere that's overriding the normal serial-getty unit settings?

kekiefer:
Thanks for the push on this. Turns out that the lock was coming from systemd when `TTYReset=yes`, which makes [`prepare_terminal()`](https://github.com/systemd/systemd/blob/v261.2/src/core/exec-invoke.c#L5011) run [`lock_dev_console()`](https://github.com/systemd/systemd/blob/v261.2/src/basic/terminal-util.c#L1032) and write a terminal reset to the tty. I ran into this problem when the USB cable was unplugged, so like that, ttyGS0 goes unread indefinitely, and the lock being on /dev/console blocks all the other console services doing the same thing. That lock path only happens when StandardOutput= is a tty, and I didn't have that in my custom service. So what I did worked, but really only by accident.

Anyhow, that's the better fix then, just make a drop-in setting `TTYReset=no` on serial-getty@ttyGS0, so we never find ourselves in this locking situation.

madisongh:
Yep, this is better, thanks. Why on earth systemd authors decided that every single serial terminal should be treated as a system console, by default and without explicitly being mentioned in the kernel command line, is mystifying to me.

PR #2348, opened 2026-09-09
Update TensorRT recipes

ichergui:
## Validation: TensorRT on Tegra264 (Thor) and Tegra234 (Orin)

Tested the TensorRT C++ sample suite and the TensorRT Python bindings on two
Jetson devkits. **22 of 22 tests pass, no failures and no skips on either
board.**

### Platforms

Both boards ran the same image lineage, so results are directly comparable:

| | Tegra264 | Tegra234 |
|---|---|---|
| Board | NVIDIA Jetson AGX Thor Developer Kit | NVIDIA Jetson AGX Orin Developer Kit (p3737-0000+p3701-0005) |
| L4T | R39, REVISION 2.1, GCID 46758480 | R39, REVISION 2.1, GCID 46758480 |
| Kernel | 6.18.48-yocto-standard | 6.18.48-yocto-standard |
| TensorRT | 10.16.2 | 10.16.2 |
| CUDA runtime | 13.2 (`/usr/local/cuda-13.2/lib/libcudart.so.13`) | 13.2 (`/usr/local/cuda-13.2/lib/libcudart.so.13`) |
| Python / numpy | 3.14.7 / 2.5.2 | 3.14.7 / 2.5.2 |
| Display server | X11 (not required by either module) | X11 (not required by either module) |

Both runs were performed in **MAXN** power mode with `jetson_clocks` pinned.

### TensorRT C++ samples (`run-tensorrt-tests`) — 8/8 on both boards

| Test | Tegra264 | Tegra234 |
|---|---|---|
| dynamic_reshape | PASS | PASS |
| int8_api | PASS | PASS |
| onnx_mnist | PASS | PASS |
| io_formats | PASS | PASS |
| progress_monitor | PASS | PASS |
| editable_timing_cache | PASS | PASS |
| non_zero_plugin | PASS | PASS |
| trtexec | PASS | PASS |

`trtexec` is a real PASS on both, not a skip: `/usr/src/tensorrt/bin/trtexec`
is present and reports `TensorRT v101602`.

### TensorRT Python bindings — 3/3 on both boards

| Test | Tegra264 | Tegra234 |
|---|---|---|
| import/module | PASS | PASS |
| build/identity_engine | PASS (plan 12228 bytes) | PASS (plan 2580 bytes) |
| execute/identity_engine | PASS | PASS |

What these cover:

- **import** — `tensorrt` and `numpy` import and report versions, i.e. the
  native extension loads for this interpreter.
- **build** — `trt.Builder` / `create_network` / `add_identity` /
  `build_serialized_network`, then `deserialize_cuda_engine` and
  `create_execution_context`.
- **execute** — device buffers allocated via `libcudart` through `ctypes`
  (no pycuda/cupy/torch on these images), pointers bound with
  `set_tensor_address()`, inference launched with `execute_async_v3()`, result
  copied back and stream synchronized. The verdict is a value check
  (`assert_allclose(..., rtol=0, atol=0)`), so output must equal input exactly
  — this is what would catch a binding built for the wrong SM.

The differing serialized-plan size (12228 vs 2580 bytes) is the expected result
of building for different GPU architectures, not a discrepancy.

### DIFF
--- a/external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_10.16.bb
+++ b/external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_10.16.bb
@@ -7,7 +7,7 @@ DEPENDS = "python3-pybind11 tensorrt-core tensorrt-plugins"
 
 COMPATIBLE_MACHINE = "(tegra)"
 
-inherit setuptools3 cmake cuda
+inherit cmake cuda python3-dir python3targetconfig
 
 SRC_REPO = "github.com/NVIDIA/TensorRT.git;protocol=https"
 SRCBRANCH = "release/${PV}"
@@ -16,10 +16,9 @@ SRC_URI = "gitsm://${SRC_REPO};branch=${SRCBRANCH} \
     "
 
 # v${PV} tag
-SRCREV = "94e2b9ef6d2cce74c76cdad499cca36cc4949197"
+SRCREV = "52399f555c2f80cb690a4a558b604e1a5f227e7c"
 
 OECMAKE_SOURCEPATH = "${S}/python"
-SETUPTOOLS_SETUP_PATH = "${B}"
 
 EXTRA_OECMAKE = "-DTENSORRT_ROOT=${S} -DTENSORRT_LIBPATH=${STAGING_LIBDIR} -DTENSORRT_MODULE=tensorrt \
                  -DCUDA_INCLUDE_DIRS=${CUDA_PATH}/include \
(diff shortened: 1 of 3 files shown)


external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins/0001-CMakeLists.txt-fix-cross-compilation-issues.patch
```diff
@@ -3,30 +3,51 @@ From: Ilies CHERGUI <ilies.chergui@gmail.com>
...
- option(BUILD_PLUGINS "Build TensorRT plugin" ON)
- option(BUILD_PARSERS "Build TensorRT parsers" ON)
+@@ -74,6 +74,10 @@
+ if (DEFINED GPU_ARCHS)
+   message(STATUS "GPU_ARCHS defined as ${GPU_ARCHS}. Generating CUDA code for SM ${GPU_ARCHS}")
+   separate_arguments(GPU_ARCHS)
++  # A cross-compilation toolchain file may already have seeded
++  # CMAKE_CUDA_ARCHITECTURES; GPU_ARCHS is meant to replace that list, not
++  # extend it.
++  set(CMAKE_CUDA_ARCHITECTURES)
```
madisongh: It would much be cleaner and simpler to drop this part of the patch and just *not set* GPU_ARCHS in CMAKE options passed in the recipe.

external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins_10.16.bb
madisongh, line 1: The commit message mentions the renaming based on the contents of the `VERSION` file in the repo, but doesn't explain **why** it was necessary, or even desirable, to make that change.

From what I can see in the Github repo, the version numbering we were using correspond with how the sources are released. Nor do I see the contents of that VERSION file getting used in the build.

If there really is a good reason to base `PV` on that file, please provide it. Otherwise, please go back to the simpler numbering scheme we had, so users can easily track down and verify based on the tags in the repo.

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_10.16.bb
```diff
@@ -7,19 +7,18 @@ DEPENDS = "python3-pybind11 tensorrt-core tensorrt-plugins"
 
 COMPATIBLE_MACHINE = "(tegra)"
 
-inherit setuptools3 cmake cuda
+inherit cmake cuda python3-dir python3targetconfig
```
madisongh, line 10: The commit message says
```
    10.16 dropped the bindings_wheel setup.py and setup.cfg templates that
    do_configure staged into ${B}, leaving setuptools3 with nothing to
    build.
```
but what really happened was that packaging was reworked to use a customized back end. At the very least, the rationale given for changing the installation step here is incomplete/incorrect.

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_10.16.0.72.bb
```diff
@@ -35,32 +34,30 @@ do_configure() {
...
-    setuptools3_do_compile
+    cp -r ${S}/python/packaging/bindings_wheel/tensorrt/plugin ${B}/tensorrt/
 }
 
 do_install() {
-    setuptools3_do_install
+    install -d ${D}${PYTHON_SITEPACKAGES_DIR}/tensorrt
+    install -m 0644 ${B}/tensorrt/__init__.py ${D}${PYTHON_SITEPACKAGES_DIR}/tensorrt/
+    install -m 0755 ${B}/tensorrt/tensorrt.so ${D}${PYTHON_SITEPACKAGES_DIR}/tensorrt/
+    cp -r ${B}/tensorrt/plugin ${D}${PYTHON_SITEPACKAGES_DIR}/tensorrt/
```
madisongh: If you can't use the actual packaging back-end that is provided in the sources, and have verified that the end result is essentially the same as doing a recursive file copy, please mention that here in a comment.

Usually recursive copies like this include some additional flags (grep around the other recipes for examples). Are any needed here?

madisongh (changes requested):
The commit messages are verbose in exactly the wrong way, and some of the changes, explained in vivid detail, are incorrect, and that makes me suspect that some AI assisted with this. Please follow our [contribution guidelines about  AI use](https://github.com/OE4T/meta-tegra/blob/master/CONTRIBUTING.md#ai-usage-policy), if that's the case.

Otherwise, I'd appreciate it if the commit messages could be made more concise and on-point. They talk way too much about the *what*, not enough about the *why*. If you need help with that, let me know.

external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins/0001-CMakeLists.txt-fix-cross-compilation-issues.patch
ichergui: I will fix that.

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_10.16.0.72.bb
ichergui: I fixed it.

external/meta-python/recipes-devtools/python3-tensorrt/python3-tensorrt_10.16.bb
ichergui: Thanks for the feedback. I addressed this comment.

external/openembedded-layer/recipes-devtools/gie/tensorrt-plugins_10.16.bb
ichergui: I was aiming to use  the version file with content version is X.Y.Z.P to align with the other TRT recipe (`tensorrt-core`, `tensorrt-plugins-prebuilt`, ...) . I noticed that the cmake uses `TRT_VERSION`  with `TRT_MAJOR` + `TRT_MINOR` + `TRT_PATCH` also the scripts and Dockerfile uses similar.

I reverted the changes.

PR #2351, opened 2026-09-15
extlinux: add UBOOT_EXTLINUX_FDT to vardeps

ilyasishbaev:
FDT is read via the auxiliary variable UBOOT_EXTLINUX_FDT, so BitBake cannot automatically determine the dependency. Added to do_create_extlinux_config vardeps to regenerate extlinux.conf when the selected DTB is changed.


madisongh:
Thanks for this, @ilyasishbaev ! I've merged it manually, including a couple of other spots where the variable dependencies weren't quite right.

PR #2354, opened 2026-09-16
linux-yocto-tegra: update kernel metadata

loblik:
I was not able to use linux-yocto with current wrynose, as openembedded linux-yocto moved to 6.18.48 and patches does not apply cleanly.

Fixed by using newer version of kernel metadata also on wrynose branch.


madisongh:
I've cherry-picked https://github.com/OE4T/meta-tegra/commit/08adb7cb8b8efe57ca6a9dc27d5a5a226758bd13 to https://github.com/OE4T/meta-tegra/commit/5656c311b908af9a10bd4b82a7aa705c672905ec on wrynose to address this.

PR #2357, opened 2026-09-17
Draft: Split TensorRT Packages into individual recipes

joekale-pp:
Dispatch and Lean libraries are independent of the "full" tensorrt library so split them to allow applications to depend on the library they need.

Also adds virtual/tensorrt-runtime to support recipes like DeepStream.

Renamed tensorrt-core to tensorrt as the library is a superset of the other libraries. Adds tensorrt-core as a PROVIDES to facilitate migration.


joekale-pp:
This is out here for comment, but if everyone likes the direction I will mark as ready.

madisongh:
For the packaging, it definitely looks like the right general direction, @joekale-pp.

I'm not as sure about the `virtual/tensorrt-runtime` part, though. My impression is that one must build one's application explicitly to use normal tensorrt, or the dispatch one, or the lean one, but maybe I'm misunderstanding?

joekale-pp:
> I'm not as sure about the `virtual/tensorrt-runtime` part, though. My impression is that one must build one's application explicitly to use normal tensorrt, or the dispatch one, or the lean one, but maybe I'm misunderstanding?

The goal here was to mitigate the dependency Deepstream has on tensorrt. We use some features of the Deepstream SDK but we still don't need the full tensorrt installed. One of the plugins will run an engine but I'm unsure if that plugin requires the full library. There is a way in the NvInfer headers to select a different runtime by passing a string, but linking against the runtime just forces the default. 

I'm definitely open to a better way of handling it though. maybe @ichergui has got word from the TensorRT team on if my assessment is correct.

